Issue #19, opened 2022-01-24
Barebones setup


IAlibay:
Please add as desired / tick off as sorted out

We need:

- [x] basic setup.py / setup.cfg / pyproject.toml
- [x] basic directory structure
- [x] license (thanks @richardjgowers !)
- [x] README.md
- [x] environment.yaml?
- [x] dummy workflow in .github folder? (just an action that tests nothing should be fine?)
- [x] pep8speaks?
- [x] coverage / codecov
- [ ] linters?
- [ ] CONTRIBUTING

dwhswenson:
What are we thinking for linters other than pep8speaks? Do we need anything else?

IMO, that's all we need in terms of something directly integrated into GitHub. Personally, I like to use CodeClimate to check complexity/repetition (goes beyond linting!), but I don't think it needs to be a hoop for contributors to jump through. I'll enable it on my fork to keep an eye on potential technical debt. I also like to run `pylint` occasionally, but I don't think we need to enforce style on all the `pylint` warnings.

mikemhenry:
Is there a way to run pep8speaks locally? If we are not using a linter that also takes care of formatting automatically, then it's nice to have something that I can run local before I push.

dwhswenson:
By default I think it's just running [pycodestyle](https://github.com/PyCQA/pycodestyle). pep8speaks has a custom YAML config for pycodestyle; default [here](https://github.com/OrkoHunter/pep8speaks/blob/master/data/default_pep8speaks.yml). You can get similar information from `pylint` or `flake8`.

I use `pylint` locally (when I remember). I think pep8speaks only errors on official errors (codes that start with E).

mikemhenry:
Okay cool, we should probably then include the config that pep8speaks uses so that we get the same errors and warning locally as we do with the bot.

mikemhenry:
We should also add a CONTRIBUTING file, is there something that OMSF is using for this?

dwhswenson:
> Okay cool, we should probably then include the config that pep8speaks uses so that we get the same errors and warning locally as we do with the bot.

Sorry, my writing wasn't clear: pep8speaks has a YAML config for itself; the link was to the default values. pycodestyle uses a [section in `setup.cfg`](https://pycodestyle.pycqa.org/en/latest/intro.html#configuration). But I think the pep8speaks default is basically what you get with no config for pycodestyle. 

> We should also add a CONTRIBUTING file, is there something that OMSF is using for this?

[...]

PR #21, opened 2022-01-24
Initial skeleton

dwhswenson:
This will handle a few parts of #19.

Adds:

* `.gitignore` from https://github.com/github/gitignore/blob/main/Python.gitignore
* `setup.cfg` configured for namespace packages in `openfe` namespace
* trivial `setup.py` (NOTE: the existence of a `setup.py` is required if you want to install in editable/develop mode, which I prefer to do)
* empty packages `openfe.setup`, `openfe.orchestration`, `openfe.simulation`, and `openfe.analysis`
* mostly empty `README.md`
* basic CI

Things still to fill in (beyond the scope of this PR):

**`setup.cfg`**:
* `author` / `author_email` (@richardjgowers, are we putting you on the hook for this? 😆 Actually, we should probably go with something like "The OpenFE Developers" with email as openfe@omsf, since I think that exists. Will wait for feedback on that.)
* `description`: currently empty
* I'm not sure that we'll stay universal, but I started us out that way
* Feel free to add/change [classifiers](https://pypi.org/classifiers/)

**CI**:

Current process (particularly installation) is awful and will need to be improved. I prefer to keep requirements in a single location, rather than have a separate CI conda env file. With OPS we did that with a [simple Python script](https://github.com/openpathsampling/openpathsampling/blob/master/devtools/setup_cfg_reqs.py) to parse requirements from the setup.cfg, and a [shell script](https://github.com/openpathsampling/openpathsampling/blob/master/devtools/conda_install_reqs.sh) that used those requirements, plus any other `requirements.txt`-style files, to create a single conda install command. I can port a version of that stuff over here, if that seems like a good approach.

**`README.md`**:

Needs more details, although the current contents are all anyone really needs to know!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,47 @@
+name: "CI"
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: "0 7 * * *"
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+  tests:
+    runs-on: ${{ matrix.OS }}-latest
+    name: "tests"
+    strategy:
+      matrix:
+        PYTHON:
+          - 3.8
+          - 3.9
+        OS:
+          - ubuntu
+          - windows
+          - macos
+    steps:
+      - uses: actions/checkout@v2
+      - uses: conda-incubator/setup-miniconda@v2
+        auto-update-conda: true
+        python-version: ${{ matrix.PYTHON }}
+        miniforge-variant: Mambaforge
+      - name: "Temporary installation (reqs and code)"
+        run: python -m pip install .[test]
+        # the above step should be changed to only install reqs
+      - name: "Versions"
+        run: conda list
+      - name: "Install"
+        run: python -m pip install --no-deps .
+      - name: "Check imports"
+        run: |
+          python -c "import openfe.setup"
+          python -c "import openfe.orchestration"
+          python -c "import openfe.simulation"
+          python -c "import openfe.analysis"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,38 @@
+[metadata]
+name = openfe
+version = 0.0.1.dev0
+description = 
+long_description = file: README.md
+long_description_type = text/markdown
+#author =
+#author_email =
+license = MIT
+license-file = LICENSE
+url = https://github.com/OpenFreeEnergy/openfe
+classifiers =
+    Development Status :: 1 - Planning
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: MIT License
+    Operating System :: POSIX
+    Operating System :: Microsoft :: Windows
+    Programming Language :: Python :: 3
+    Topic :: Scientific/Engineering :: Bio-Informatics
+    Topic :: Scientific/Engineering :: Chemistry
+
+[options]
+python_requires = >= 3.8
+install_requires =
+    numpy
+    networkx
+    rdkit-pypi
+
+include_package_data = True
+namespace_packages = openfe
+
+[options.extras_require]
+test =
+    pytest
+    pytest-cov
+
+[bdist_wheel]
+universal = 1
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,152 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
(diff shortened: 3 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
+name: "CI"
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: "0 7 * * *"
```
IAlibay, line 10: I'm not super convinced into the need for a daily cron job, mostly because upstream never really changes that quickly. How would we feel about a twice weekly cron instead?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+
```
IAlibay, line 11: Can we put in a concurrency call here? see: https://github.com/OpenFreeEnergy/Lomap/blob/b31e6cde3f8e4117e75fd592ac925c17b77cae71/.github/workflows/CI.yaml#L14

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+    name: "tests"
+    strategy:
+      matrix:
+        PYTHON:
+          - 3.8
+          - 3.9
+        OS:
+          - ubuntu
+          - windows
+          - macos
```
IAlibay, line 28: I'm personally partial to matrix elements looking like this instead (just seems clearer that it's a list):

```suggestion
        PYTHON: [3.8, 3.9]
        OS: [ubuntu, windows, macos]
```

I will say I'm more partial to lower case variables (similar to: https://github.com/OpenFreeEnergy/Lomap/blob/b31e6cde3f8e4117e75fd592ac925c17b77cae71/.github/workflows/CI.yaml#L31-L34) but I can live with things either way.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: conda-incubator/setup-miniconda@v2
```
IAlibay, line 31: More of a personal flair thing, but I'd add some empty lines between steps for readability.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name: "Temporary installation (reqs and code)"
+        run: python -m pip install .[test]
+        # the above step should be changed to only install reqs
+      - name: "Versions"
+        run: conda list
+      - name: "Install"
+        run: python -m pip install --no-deps .
+      - name: "Check imports"
+        run: |
+          python -c "import openfe.setup"
```
IAlibay, line 44: Should take this opportunity to switch this to a test like this so we can have our first pytest call (& get coverage, etc.. started)?

```python
from openfe import setup, orchestration, simulation, analysis
import sys

@pytest.mark.parametrize('module',
    ['setup', 'orchestration', 'simulation', 'analysis])
def test_imported(module):
    """Sample test, will always pass so long as import statement worked"""
    assert module in sys.modules
```

setup.cfg
```diff
@@ -0,0 +1,38 @@
+[metadata]
+name = openfe
+version = 0.0.1.dev0
+description = 
+long_description = file: README.md
+long_description_type = text/markdown
+#author =
```
IAlibay, line 7: I'd also advocate for "OpenFE developers" + "openfe@omsf.io" here.

setup.cfg
```diff
@@ -0,0 +1,38 @@
...
+#author_email =
+license = MIT
+license-file = LICENSE
+url = https://github.com/OpenFreeEnergy/openfe
+classifiers =
+    Development Status :: 1 - Planning
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: MIT License
+    Operating System :: POSIX
+    Operating System :: Microsoft :: Windows
```
IAlibay, line 17: Add MacOS here?

setup.cfg
```diff
@@ -0,0 +1,38 @@
...
+    Programming Language :: Python :: 3
```
IAlibay, line 18: I generally tend to throw in separate supported minor versions here, but I guess it doesn't matter.

setup.cfg
```diff
@@ -0,0 +1,38 @@
...
+    Topic :: Scientific/Engineering :: Bio-Informatics
+    Topic :: Scientific/Engineering :: Chemistry
+
+[options]
+python_requires = >= 3.8
+install_requires =
+    numpy
+    networkx
+    rdkit-pypi
```
IAlibay, line 27: It's not clear to me that `rdkit-pypi` is actually a product of the rdkit developers (indeed the fact that it claims to support py3.10 before there being one on conda-forge seems a bit suspicious).

IAlibay (commented):
regarding CI:

My view here is that whilst `environment.yaml` files aren't the cleanest things in the world (especially since there is always a risk that they'll fall out of sync with setup), they are much more user friendly (especially if we have the one file we have hanging around), than going with custom scripts to pick up from setup.cfg, etc....

I'd prefer go for simplicity & future contributor friendliness than cleanliness & complexity

Plus if anything we'll need a yaml file if we want to set up anything related to `binder`.

setup.cfg
richardjgowers: Yeah conda-forge is the official source afaik. We're going to have to have conda deps at some point I think

setup.cfg
richardjgowers: I'll need to do some tinkering to make openfe@omsf.io redirect properly to all current devs

richardjgowers (approved):
Gonna merge this so we can get moving, but some of the comments probably need to get addressed still

PR #26, opened 2022-01-26
Add base Scorer

### LINKED ISSUE #2: Implement Scorer Base Implementation
```python

s = Scorer(stuff)

m = Mapper.suggest_mapping(m1, m2)

f : float = s.score(m)
```

dwhswenson:
Resolves #2.

## Overview

This provides a base class `Scorer`. The `Scorer.__call__` returns a `NamedTuple` with attributes `score` and `annotation`. Subclass implementers **must** implement the method `score`, and **may** override the method `annotation`.

## Self-criticism

* The module-level docstring is pretty thin. I just wasn't sure what else to say at this point.
* There's perhaps a little too much detail on implementation in the `Scorer` class docstring. I'm concerned about this getting out-of-sync with code, but it will be important to document the idea. Any better approach welcome.
* This is heavier in precise use of typing than I usually care about, but for abstract classes, I think detalied typing might be worth it.
* There's an overly-trivial placeholder fixture for a `mock_atommapping`. Since we don't have a completed `AtomMapping` class, we can't create a good fixture for it yet, but I added the stub.

### DIFF
--- a/openfe/setup/scorer.py
+++ b/openfe/setup/scorer.py
@@ -1,4 +1,71 @@
-class Scorer:
-    def __call__(self, atommapping):
-        return 0.0
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
 
+"""
+Scorers
+=======
+
+A scorer is used to determine a score (and optionally, annotations) for a
+given AtomMapping.
+"""
+
+from typing import NamedTuple, Dict, Any, Union
+
+
+class ScoreAnnotation(NamedTuple):
+    """Container for a score from a mapping and any associated annotations.
+
+    Parameters
+    ----------
+    score : float
+        The score, or ``None`` if there is no score associated
+    annotation : Dict[str, Any]
+        Mapping of annotation label to annotation value for any annotations
+    """
+    score: Union[float, None]
+    annotation: Dict[str, Any]
+
+
+class Scorer:
+    """Abstract base class for Scorers.
+
+    To implement a subclass, you must implement the ``score`` method. If
+    your ``Scorer`` only returns annotations, then return None from the
+    ``score`` method. You may optionally implement the ``annotation``
+    method. The default ``annotation`` returns an empty dictionary.
+
+    Use a ``Scorer`` by calling it as a function.
+    """
+    def score(self, atommapping) -> Union[float, None]:
+        """Calculate the score for an AtomMapping.
+
+        Parameters
+        ----------
+        atommapping : AtomMapping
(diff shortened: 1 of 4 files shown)


richardjgowers (approved):
Looks good.  I did have somewhere in the back of my mind that only requiring `__call__` would allow functions to also be used, obviously methods break that but maybe that's not such a drama.

dwhswenson:
> Looks good. I did have somewhere in the back of my mind that only requiring __call__ would allow functions to also be used, obviously methods break that but maybe that's not such a drama.

Implementers could still write a function, as long as it returns a (duck type of a) `ScoreAnnotation`. Maybe it would be best to make `score` and `annotation` private methods if that case is likely to be used, so they aren't used independently of `__call__`?

openfe/setup/scorer.py
```diff
@@ -1,4 +1,71 @@
...
+        ----------
+        atommapping : AtomMapping
+            AtomMapping to score
+
+        Returns
+        -------
+        Union[float, None] :
+            The score, or ``None`` if no score is calculated
+        """
+        raise NotImplementedError()
```
IAlibay, line 52: I know this _shouldn't_ be ever used on it's own, but for the sake of someone doing something silly + coverage can we a) get a test that triggers this, b) a simple message to tell folks to not call the base class directly?

dwhswenson: I usually use the rule that `raise NotImplementedError` is part of the coverage exclude lines (along with `__repr__`); do we want to make a practice of always testing `NotImplementedError`s? (I'm fine either way.) Will definitely add a clearer message, though.

IAlibay: I'm not super fussed either way, I do find trying to trigger errors cathartic so I'm happy just adding them randomly on a bored weekend evening.

dwhswenson: I'll went ahead and added the test -- it isn't hard and doesn't hurt anything. We can always ignore in coverage later (having a test that covers lines that don't count toward coverage isn't a problem).

Speaking of coverage, CodeCov is being flaky. I'm trying to dig up some old CodeCov problems I've had (and their solutions) to see if it is something I've seen before.

dwhswenson:
Shouldn't pep8speaks have had something to say about this PR? Is it not enabled for PRs coming from forks?

IAlibay:
> Shouldn't pep8speaks have had something to say about this PR? Is it not enabled for PRs coming from forks?

Yeah I'm not sure why it's not being triggered... default options _should_ work on forks.

IAlibay (approved):

Issue #8, opened 2022-01-24
AtomMapping correspondence lookups


richardjgowers:
depends on #20

richardjgowers:
I'm going to close this as this is technically achieved with `AtomMapping.mol1_to_mol2`, but we might revisit and make this easier/better once we start using it in anger

richardjgowers:
I think something chemical might be something like:
```python
class AtomMapping:
    def __getitem__(self, atom: Atom) -> Union[Atom, None]:
        # ???
        return other_atom
```
But this is a lot harder to make work than it initially seems.  I think sticking to indices to a canonically ordered molecule where possible probably saves a lot of headaches.  As before, I'm also waiting on seeing how awkward future tasks are before loading up AtomMapping with too much functionality.

dwhswenson:
Yep, I was thinking exactly along the lines @richardjgowers proposes. I also get the issues from a practical standpoint. However, isn't all of this implicitly required for @mikemhenry's visualization?

Obviously we haven't yet defined where the `Atom` type comes from, which would be a requirement for this.

dwhswenson:
(This is the main reason I'm pausing before moving this to "done" -- if we get that extra functionality for free with Mike's work-in-progress, then maybe it's worth marking that work as solving either this issue or a separate issue that is specifically defined for that.)

PR #28, opened 2022-01-27
atommapping implementation

### LINKED ISSUE #6: Implement AtomMapping container
Should be a read only simple data class holding enough information for mapping two sets of ints (atom indices) to each other...

```python
class AtomMapping:
    AtoB : dict[int, int]
    BtoA : dict[int, int]
```

richardjgowers:
Can probably leave it this simple, makes it easy for others to only use the AtomMappers.

Resolves #6.

### DIFF
--- a/openfe/setup/atommapping.py
+++ b/openfe/setup/atommapping.py
@@ -22,7 +22,7 @@ class AtomMapping:
     """
     mol1: RDKitMol
     mol2: RDKitMol
-    mol1_to_mol2: Dict[int, Union[int, None]]
+    mol1_to_mol2: Dict[int, int]
 
     def __hash__(self):
         # hash of SMILES of each molecule, and a hash of both mappings
--- a/openfe/setup/atommapping.py
+++ b/openfe/setup/atommapping.py
@@ -11,30 +11,30 @@ class AtomMapping:
 
     Attributes
     ----------
-    molA, molB : rdkit.Mol
+    mol1, mol2 : rdkit.Mol
       the two Molecules in the mapping
-    AtoB, BtoA : dict
+    mol1_to_mol2, mol2_to_mol1 : dict
       maps the index of an atom in either molecule **A** or **B** to the other.
       If this atom has no corresponding atom, None is returned.
 
 
-    The size of molecule A/B is given by the length of the AtoB/BtoA dictionary
+    The size of each molecule is given by the length of the x_to_y dictionary
     """
-    molA: RDKitMol
-    molB: RDKitMol
-    AtoB: Dict[int, Union[int, None]]
-    BtoA: Dict[int, Union[int, None]]
+    mol1: RDKitMol
+    mol2: RDKitMol
+    mol1_to_mol2: Dict[int, Union[int, None]]
+    mol2_to_mol1: Dict[int, Union[int, None]]
 
-    def __init__(self, molA: RDKitMol, molB: RDKitMol):
-        self.molA = molA
-        self.molB = molB
-        self.AtoB = dict()
-        self.BtoA = dict()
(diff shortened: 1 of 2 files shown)


IAlibay:
Just having a quick glance - is there a reason for having a Perses specific method here? Ideally we should try to build this in a way that we don't have to change this class every time someone comes up with a new mapper right?

openfe/setup/atommapping.py
```diff
@@ -1,8 +1,41 @@
...
-        return None
 
+    The size of molecule A/B is given by the length of the AtoB/BtoA dictionary
+    """
+    molA: RDKitMol
+    molB: RDKitMol
+    AtoB: Dict[int, Union[int, None]]
+    BtoA: Dict[int, Union[int, None]]
+
+    def __init__(self, molA: RDKitMol, molB: RDKitMol):
```
dwhswenson, line 28: Any reason not to include the dicts in the init sig? I think it's easier for a user to write:

```python
return AtomMapping(molA, molB, myAtoB, myBtoA)
```

than

```python
mapping = AtomMapping(molA, molB)
mapping.AtoB = myAtoB
mapping.BtoA = myBtoA
return mapping
```

If you're worried about users mixing up order, make them keyword required. If not, maybe the whole thing is `@dataclass`?

dwhswenson (changes requested):
A couple small changes. (And you may get two versions of one of my comments, since you updated the PR, fixing one of the issues, when I was reviewing it!)

And I see you just updated the PR again as I was typing this. 😂 Hope my review is still relevant!

openfe/setup/atommapping.py
```diff
@@ -1,8 +1,27 @@
...
+      maps the index of an atom in either molecule **A** or **B** to the other.
+      If this atom has no corresponding atom, None is returned.
 
+
+    The size of molecule A/B is given by the length of the AtoB/BtoA dictionary
+    """
+    AtoB: Dict[int, Union[int, None]]
+    BtoA: Dict[int, Union[int, None]]
+
+    def __init__(self):
```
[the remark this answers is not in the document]
richardjgowers: Yeah I realised point 2 a little later than you

openfe/setup/atommapping.py
```diff
@@ -1,8 +1,34 @@
...
+    mol1, mol2 : rdkit.Mol
+      the two Molecules in the mapping
+    mol1_to_mol2, mol2_to_mol1 : dict
+      maps the index of an atom in either molecule **A** or **B** to the other.
+      If this atom has no corresponding atom, None is returned.
 
+    """
+    mol1: RDKitMol
+    mol2: RDKitMol
+    mol1_to_mol2: Dict[int, Union[int, None]]
```
dwhswenson, line 25: Are values here still `Union[int, None]` or are they now always `int`?

IAlibay: they _should_ just be `int`

dwhswenson, line 25:
```suggestion
    mol1_to_mol2: Dict[int, int]
```

Suggesting and accepting suggestion at suggestion of @richardjgowers

IAlibay (approved):

dwhswenson (approved):
LGTM; will merge. Just commenting on a couple of things that might change later, but don't need to change for this PR:

1. I'd love to see better names for fixtures `simple_mapping` and `other_mapping`, and then to move them to `conftest.py` and allow them to be reused by other tests.

2. The name `mol1_to_mol2` might change because we might want that name for the lookups in [#8](https://github.com/OpenFreeEnergy/openfe/issues/8) (along with `mol2_to_mol1`).

PR #29, opened 2022-01-28
Minor updates to Scorer

dwhswenson:
This is a small PR where I'm cherry-picking some of the changes that have come up as I've been working on AtomMappers. Main things:

* Added `setup/errors.py`, which currently just contains a string template for errors when using abstract methods.
* Make `score` and `annotation` functions in `Scorer` private. This will allow arbitrary functions to be used as scorers (useful if an AtomMapper wants to provide a scoring function as an instance method).
* Move the fixtures for atommappings into conftest.py; use one of them instead of the nonsense placeholder for the scoring tests.

### DIFF
--- a/openfe/setup/errors.py
+++ b/openfe/setup/errors.py
@@ -1,0 +1,3 @@
+ABSTRACT_ERROR_STRING = ("'{cls}' is an abstract class and should not be "
+                         "used directly. Please use a specific subclass of "
+                         "'{cls}' that implements {func}.")
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,6 +63,8 @@ jobs:
           pytest -n 2 -v --cov=openfe --cov-report=xml openfe/tests/
 
       - name: codecov
+        if: ${{ github.repository == 'OpenFreeEnergy/openfe'
+                && github.event != 'schedule' }}
         uses: codecov/codecov-action@v2
         with:
           file: coverage.xml
--- a/openfe/setup/scorer.py
+++ b/openfe/setup/scorer.py
@@ -11,6 +11,8 @@ given AtomMapping.
 
 from typing import NamedTuple, Dict, Any, Union
 
+from openfe.setup.errors import ABSTRACT_ERROR_STRING
+
 
 class ScoreAnnotation(NamedTuple):
     """Container for a score from a mapping and any associated annotations.
@@ -36,7 +38,7 @@ class Scorer:
 
     Use a ``Scorer`` by calling it as a function.
     """
-    def score(self, atommapping) -> Union[float, None]:
+    def _score(self, atommapping) -> Union[float, None]:
         """Calculate the score for an AtomMapping.
 
         Parameters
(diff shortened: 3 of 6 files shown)


richardjgowers (approved):
LGTM

openfe/setup/errors.py
```diff
@@ -0,0 +1,3 @@
+ABSTRACT_ERROR_STRING = ("'{cls}' is an abstract class and should not be "
```
IAlibay, line 1: Do we envision having abstract classes outside of setup? If so maybe it needs to be in the top level directory?

IAlibay (approved):
lgtm just see the one comment re: location of errors.py.

openfe/setup/errors.py
dwhswenson: I had the samne thought. So far, I'm operating under the assumption that we are likely to split into multiple repos (something we should discuss on Monday, probably).

If multiple repos, obviously it must be here. In single repo, then I agree that it belongs top level.

IAlibay: How about a temporary TODO so we get back to it later? (not sure how folks feel about this, there are fun gh bots that will scrape all the TODO comments and turn them into issues for you)

(feel free to say no and merge tbh, it's not like we're likely to forget either)

```suggestion
# TODO: move if we don't split to multiple repos
ABSTRACT_ERROR_STRING = ("'{cls}' is an abstract class and should not be "
```

dwhswenson: I agree with the principle of leaving a TODO on this sort of thing, but in this specific case, I'm hoping we'll decide about whether to split repos very soon -- so then it'll be decided soon enough that we don't need the reminder. And then if we need to move the file, it'll be a quick PR right after the meeting! 😊

IAlibay:
huh weird - I merged this, and it seems like the changes went through.. but somehow the PR hasn't closed and now claims conflicts? O.o

dwhswenson:
> huh weird - I merged this, and it seems like the changes went through.. but somehow the PR hasn't closed and now claims conflicts? O.o

That *is* all kinds of weird... I'll try sync'ing the branch locally and see what shakes out

dwhswenson:
So, sync'ing (just `git pull upstream main && git push`, which only pulled in the changes from #27, I think) seems to have automatically moved the status to "closed"... which is kind of annoying, because I would want the status to be "merged"! (purple, not red!)

IAlibay:
yeah :/ that's definitely a new one to me, never seen github bug out like this on a merge.

dwhswenson:
Just curious -- regular merge or squash merge? I'm not sure how squashing handles cherry-picked commits (although I'd expect any problem to show up in #30, not here).

IAlibay:
Pretty sure it was a regular merge.

dwhswenson:
FWIW, running the [GraphQL Explorer](https://docs.github.com/en/graphql/overview/explorer) shows that a search for merged PRs shows this one.

Query:

```graphql
query {
  repository(name: "openfe", owner: "openfreeenergy") {
    pullRequests(states: MERGED, last: 2) {
      edges {
        node {
          merged
          number
          title
        }
      }
    }
  }
}
```

Results:

```json
{
  "data": {
    "repository": {
      "pullRequests": {
        "edges": [
          {
            "node": {
              "merged": true,
              "number": 27,
              "title": "Prevent CodeCov from running on forks, cron jobs"
            }
          },
          {
            "node": {
              "merged": true,
              "number": 28,
              "title": "atommapping implementation"
            }
          }
        ]
      }
    }
  }
}
```

Anyway, no more debugging from me on that one, though I might search around for similar problems (and might make a post on github.community to see if anyone can explain).

dwhswenson:
Posted issue on github.community: https://github.community/t/pr-marked-as-conflicts-after-merge-syncing-incorrectly-marks-it-closed/226240

PR #30, opened 2022-01-28
AtomMapper base classes

### LINKED ISSUE #5: Implement AtomMapper abstract implementation
```python
class AtomMapper:
    @abstractmethod
    def suggest_mapping(m1: Molecule, m2: Molecule) -> Iterable[AtomMapping]:
        # magic!

    @abstractmethod
    def score(m: AtomMapping) -> float:
        # optionally implement a scorer for the maps you create (caching possible?)
        return 0.0
```

dwhswenson:
Resolves #5.

This implements two abstract classes for atom mappers:

* `AtomMapper`: This is a pretty trivial ABC.
* `ScoredAtomMapper`: This ABC combines `AtomMapper` with a scoring function. See discussion in [#6(comment)](https://github.com/OpenFreeEnergy/openfe/issues/6#issuecomment-1023500766) and related for the inspiration.

For both, implementers must implement `_mappings_generator`. For `ScoredAtomMapper`, implementers must also implement `_calculate_score`.

Note that this PR includes #29, so that one needs to be merged first. Once that's merged I'll pull those changes in here so that the diff doesn't include stuff from #29. (Or I think you can just look at 314430723a4cb3ab3eaccc994160b9df6a8ba1ce).


richardjgowers (changes requested):
I'm not sure we need the `ScoredAtomMapper` abstract class, but it does serve as a useful example.  It's very possible that there's a mapper with 2 scoring functions, so usage alone the lines of:


```python
class MapperWithScorers:
   def suggest_mappings(self, mol1, mol2):
        ...

    def criteria_one(self, mapping):
        return something fancy

    def criteria_two(self, mapping):
        return something else

my_mapper = MapperWithScorers()

network = generate_network(mappers=[my_mapper], scorers=[my_mapper.criteria_one, my_mapper.criteria_two], weights=[0.4, 0.6])
```

IAlibay (approved):

dwhswenson:
@richardjgowers : Agreed, updated. I was mainly trying to keep the spirit of having the `score` method in the `AtomMapper` as issue #5 suggested. It didn't make sense to leave that as an optionally-implemented method as the existence of `Scorer`s implied. So I added the ideas from https://github.com/OpenFreeEnergy/openfe/issues/6#issuecomment-1023500766.

Tests for `ScoredAtomMapper` were a lot of fun, but hey -- sometimes ya gotta murder your darlings.

Current implementation still separates `_mappings_generator` as the internally-implemented thing from `suggest_mappings` as the user-facing method. This was done to keep consistency when working with subclasses of `AtomMapper`. Keep it this way or kill it?

Also, as class with nothing but a single public method, is there a need for `AtomMapper` to be class, rather than just any callable? I'm not sure which approach is more straightforward for inexperienced devs.

richardjgowers:
@dwhswenson yeah could just make this any callable too, will discuss.  Making people implement a private method makes the top level documentation consistent (as this is from the base class).

richardjgowers (approved):

PR #33, opened 2022-01-31
Remove Scorers

dwhswenson:
A scorer will just be any callable (at least, until we need to worry about serialization).


richardjgowers:
RIP

PR #34, opened 2022-02-01
started Network implementation

### LINKED ISSUE #14: Implement Network container

### LINKED ISSUE #17: Implement radial graph network planner
```python
ligands = [Chem.MolFromSomething(f) for f in files]
central_ligand = Chem.MolFromSomething()

n: Network = openfe.setup.generate_radial_network(ligands, hub=central_ligand,
                                          mappers=[openfe.setup.mappers.SimpleMapper()])
```

richardjgowers:
closes #14 
closes #17

hashed out radial graph construction a little

### DIFF
--- a/openfe/setup/network.py
+++ b/openfe/setup/network.py
@@ -1,0 +1,8 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
+
+
+class Network:
+    @classmethod
+    def from_lomap(cls, lomap_file):
+        pass
--- a/openfe/setup/network.py
+++ b/openfe/setup/network.py
@@ -1,13 +1,64 @@
 import math
+import networkx as nx
+from rdkit import Chem
+from typing import TypeVar
+
+
+NetworkXGraph = TypeVar("NetworkXGraph")
+
+
+def _hashmol(mol):
+    # canonical by default
+    # TODO: coordinates
+    return Chem.MolToSmiles(mol)
 
 
 class Network:
+    """A set of ligands and mappings between them"""
+    _graph: NetworkXGraph
+
+    def __init__(self):
+        self._graph = nx.Graph()
+
+    def __getitem__(self, key):
+        key = (_hashmol(key[0]), _hashmol(key[1]))
+
+        return self._graph[key]
+
+    def add_edge(self, atommapping):
+        """Manually add an edge onto an existing Network
+
+        Parameters
+        ----------
+        atommapping : AtomMapping
+          the mapping between two ligands to add
+
+        Raises
+        ------
+        ValueError
+          if an edge between two ligands already exists
+        """
+        key = (_hashmol(atommapping.mol1), _hashmol(atommapping.mol2))
+
+        if key in self._graph:
+            raise ValueError("This edge already exists in Network")
+
+        self._graph.add_edge(key, atommapping)
+
+    def get_edges(self):
+        """Iterable of edges in this Network"""
+        yield from self._graph.edges()
+
     @classmethod
-    def from_lomap(self, lomap_file):
-        pass
+    def from_lomap(cls, lomap_file):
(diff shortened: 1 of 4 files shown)


richardjgowers:
sorry @dwhswenson , forgot to check if you'd already took the issue, hopefully you hadn't started yet

dwhswenson:
> sorry @dwhswenson , forgot to check if you'd already took the issue, hopefully you hadn't started yet

😡 well, I'll just review with all the thoughts from what I had sketched out. 😂 There are a couple differences that are worth mentioning.

openfe/setup/network.py
```diff
@@ -1,13 +1,64 @@
 import math
+import networkx as nx
+from rdkit import Chem
+from typing import TypeVar
+
+
+NetworkXGraph = TypeVar("NetworkXGraph")
+
+
+def _hashmol(mol):
```
dwhswenson, line 10: move to public method in utils? this comes back to my sense that we need an actual hashable molecule class, because it will be reused in so many places, and users will want to use molecules as keys in dicts.

openfe/setup/network.py
```diff
@@ -1,13 +1,64 @@
...
+
     @classmethod
-    def from_lomap(self, lomap_file):
-        pass
+    def from_lomap(cls, lomap_file):
+        raise NotImplementedError  # issue #16
+
+    def to_json(self):
+        """Return json string representation of this Network"""
+        raise NotImplementedError  # issue #15
```
dwhswenson, line 58: Suggestion: serialize by converting to json-serializable dict, not directly to json. This makes created mutated versions of the class very easy. We probably need some more discussion on serialization, generally.

Also, I'm not sure that JSON is the best serialization format here. Much as I hate XML, there are multiple standard file formats for serializing graphs (with arbitrary metadata) as XML.

openfe/setup/network.py
```diff
@@ -1,13 +1,64 @@
...
+    def __init__(self):
+        self._graph = nx.Graph()
+
+    def __getitem__(self, key):
+        key = (_hashmol(key[0]), _hashmol(key[1]))
+
+        return self._graph[key]
+
+    def add_edge(self, atommapping):
+        """Manually add an edge onto an existing Network
```
dwhswenson, line 29: Can we make `Network` immutable?

That's basically:

```python
class Network:
    def __init__(self, edges):
        self.edges = edges
        graph = nx.Graph()
        for edge in edges:
            graph.add_edge(edge)
        self._graph = nx.freeze(self._graph)

    def add_edge(self, edge):
        # now returns a new object; maybe "add" isn't the best name?
        return Network(self.edges + [edge])
```

openfe/setup/network.py
```diff
@@ -1,13 +1,64 @@
...
+    def from_lomap(cls, lomap_file):
+        raise NotImplementedError  # issue #16
+
+    def to_json(self):
+        """Return json string representation of this Network"""
+        raise NotImplementedError  # issue #15
 
 
-def generate_radial_graph(ligands, central_ligand, mappers, extra_scorers=None):
+def generate_radial_graph(ligands, central_ligand, mappers, scorers=None):
```
dwhswenson, line 61: beyond the scope of this PR, but shouldn't this (and `minimal_spanning_graph`) be subclasses of `NetworkPlanner`?

dwhswenson (changes requested):
A couple thoughts based on things I liked more in the implementation I did. Main thing is that I think it would be very good if our actual `Network` objects were immutable, so the `add_edges` (or hopefully we find a better name?) creates a new object.

dwhswenson:
@richardjgowers : To minimize conflicts with the network I've been writing, maybe move `network.py` to `planner.py` or `networkplanner.py` or something? Everything in there is actually planners, right? I can make the same move and that should reduce problems with merge conflicts.

richardjgowers:
@dwhswenson done, hopefully I'm out of your way now

PR #39, opened 2022-02-03
Molecule class

dwhswenson:
Adds `Molecule` class. In principle, this could be extended in the future to automate conversion between properties `rdkit`, `oechem`, `openff`, etc. to allow access to alternate molecular representations.

Includes the hashing the @richardjgowers previously wrote, with addition to support the `_Name` attribute as discussed in Slack.

One thing that might improve this: is there a way to make an rdkit molecule immutable?

### DIFF
--- a/openfe/setup/molecule.py
+++ b/openfe/setup/molecule.py
@@ -1,6 +1,9 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/openfe
 
+from typing import TypeVal
+RDKitMol = TypeVar('RDKitMol')
+
 from openfe.utils.molhashing import hashmol
 
 
@@ -12,18 +15,17 @@ class Molecule:
     rdkit : rdkit.Mol
         rdkit representation of the molecule
     """
-    def __init__(self, rdkit):
+    def __init__(self, rdkit: RDKitMol):
         self._rdkit = rdkit
-        self._hash = None
+        self._hash = hashmol(self._rdkit)
 
     # property for immutability; also may allow in-class type conversion
     @property
-    def rdkit(self):
+    def rdkit(self) -> RDKitMol:
+        """RDKit representation of this molecule"""
         return self._rdkit
 
     def __hash__(self):
-        if self._hash is None:
-            self._hash = hashmol(self.rdkit)
         return hash(self._hash)
 
     def __eq__(self, other):
--- a/openfe/setup/molecule.py
+++ b/openfe/setup/molecule.py
@@ -3,6 +3,7 @@
 
 from openfe.utils.molhashing import hashmol
 
+
 class Molecule:
     """Molecule wrapper to provide proper hashing and equality.
 
--- a/openfe/setup/molecule.py
+++ b/openfe/setup/molecule.py
@@ -1,7 +1,7 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/openfe
 
-from typing import TypeVal
+from typing import TypeVar
 RDKitMol = TypeVar('RDKitMol')
 
 from openfe.utils.molhashing import hashmol
--- a/openfe/setup/molecule.py
+++ b/openfe/setup/molecule.py
@@ -1,0 +1,29 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
+
+from openfe.utils.molhashing import hashmol
+
+class Molecule:
+    """Molecule wrapper to provide proper hashing and equality.
+
+    Parameters
+    ----------
+    rdkit : rdkit.Mol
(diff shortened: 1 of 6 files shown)


openfe/setup/molecule.py
```diff
@@ -0,0 +1,30 @@
...
+    """Molecule wrapper to provide proper hashing and equality.
+
+    Parameters
+    ----------
+    rdkit : rdkit.Mol
+        rdkit representation of the molecule
+    """
+    def __init__(self, rdkit):
+        self._rdkit = rdkit
+        self._hash = None
```
IAlibay, line 17: Is there a case where we don't want this to be calculated on init?

openfe/setup/molecule.py
```diff
@@ -0,0 +1,30 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
+
+from openfe.utils.molhashing import hashmol
+
+
+class Molecule:
+    """Molecule wrapper to provide proper hashing and equality.
+
+    Parameters
```
IAlibay, line 10: It's a bit cyclic but should we mention that there's an rdkit Mol as an attribute in the class?

IAlibay (approved):
lgtm, couple of small probably a bit silly comments.

openfe/setup/molecule.py
dwhswenson: I was 50/50 on which way to go with this. The delayed calculation idiom is more a matter of autopilot than any opinion; happy to switch it if that's preferred.

I suppose that delayed calculation is slower by the `if` statement, so if we always calculate it anyway, might as well do it in `__init__`. But I really doubt it'll be a bottleneck either way.

openfe/setup/molecule.py
dwhswenson: Yeah, I should probably docstring the property and also add typing info to this -- wanted to get a version out fast. I'll make those changes after lunch.

openfe/setup/molecule.py
IAlibay: Yeah I guess it depends if we always expect our use cases to use the hash? If we have cases where it's not needed then this approach makes more sense.

I don't have a strong opinion either way to be honest, the `if` is going to be super inconsequential compared to the cost of `MolToSmiles` (although even then that's pretty quick).

openfe/setup/molecule.py
```diff
@@ -0,0 +1,30 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
+
+from openfe.utils.molhashing import hashmol
+
+
+class Molecule:
```
richardjgowers, line 7: I know I asked for a Molecule class, but what if we called this `NetworkNode`, since that's likely all we use it for.  Then `node.molecule` giving the rdkit Molecule is less awful than `molecule.molecule`

dwhswenson: My thought was that, once we have `Molecule` class, we can use properties to "export" to different codes: `mol.rdkit`, `mol.oechem`, etc. Might also allow "import" as well.

If the user doesn't create the `Molecule` instance for `ligand`, then we don't have `(ligand in network.nodes) is (ligand in list(network.nodes))`, which is the consistency check I'm concerned about.

openfe/tests/setup/test_molecule.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class TestMolecule:
+    def test_rdkit_behavior(self, ethane, alt_ethane):
+        # Check that fixture setup is correct (we aren't accidentally
+        # testing tautologies) and that rdkit continues to behave in a way
+        # that require our custom equality.
+        assert ethane is not alt_ethane
+        assert ethane.rdkit is not alt_ethane.rdkit
+        assert ethane.rdkit != alt_ethane.rdkit  # rdkit might change this
```
richardjgowers, line 26: we don't have to test this, it's not our behaviour

openfe/setup/molecule.py
richardjgowers: put this ahead of time, it reduces the boilerplate and we'll likely always use it

openfe/tests/setup/test_molecule.py
dwhswenson: removed the last check, which is rdkit dependent. Other two are to ensure that the tests don't give false passes if something changes in fixtures.

openfe/setup/molecule.py
```diff
@@ -0,0 +1,32 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
+
+from typing import TypeVal
```
IAlibay, line 4:
```suggestion
from typing import TypeVar
```

dwhswenson: oops! thanks

richardjgowers (approved):

PR #41, opened 2022-02-04
network planning functions started

richardjgowers:
- udpates AtomMapper to use Molecule
- Lomap AtomMapper handles this change too
- radial graph done and tested
- minimal spanning tree stubbed

### DIFF
--- a/openfe/setup/network_planning.py
+++ b/openfe/setup/network_planning.py
@@ -47,7 +47,7 @@ def generate_radial_graph(ligands, central_ligand, mappers, scorer=None):
                 if not scorer:
                     best_mapping = mapping
                     break
-                score = mapper.score_mapping(mapping)
+                score = scorer(mapping)
 
                 if score < best_score:
                     best_mapping = mapping
--- a/openfe/setup/network_planning.py
+++ b/openfe/setup/network_planning.py
@@ -1,3 +1,5 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
 import math
 from typing import Iterable
 
--- a/openfe/setup/network_planning.py
+++ b/openfe/setup/network_planning.py
@@ -6,9 +6,9 @@ from typing import Iterable
 from openfe.setup import Network, Molecule, AtomMapper
 
 
-def generate_radial_graph(ligands: Iterable[Molecule],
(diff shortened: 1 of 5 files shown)


richardjgowers:
@IAlibay @dwhswenson any chance this can get reviewed?

openfe/setup/network_planning.py
```diff
@@ -1,52 +1,82 @@
...
+        best_score = math.inf
         best_mapping = None
 
         for mapper in mappers:
-            for mapping in mapper.suggest_mappings(central_ligand, l):
-                score = mapper.score_mapping(mapping)
+            for mapping in mapper.suggest_mappings(central_ligand, ligand):
+                if not scorer:
+                    best_mapping = mapping
+                    break
```
IAlibay, line 54: The behaviour is kinda odd here and it kinda clashes a bit with how it's described in the docstring since you'll essentially default to taking the first suggested mapping of the last mapper.

openfe/setup/network_planning.py
```diff
@@ -1,52 +1,82 @@
...
+                score = scorer(mapping)
 
                 if score < best_score:
                     best_mapping = mapping
                     best_score = score
 
-            if best_mapping is None:
-                raise ValueError("No mapping found!")
-            n.add_edge(best_mapping)
+        if best_mapping is None:
```
IAlibay, line 61: So thinking about potential user stories for this method. There might be a case for someone to just load a big SDF and go "I want to create a network of all the ligands that will match to this given ligand".

I'm not 100% sure we want to cover this use case, but if we wanted to cover it, do we want to maybe add a "force" argument that just drops ligands that aren't compatible?

openfe/setup/network_planning.py
```diff
@@ -1,52 +1,82 @@
...
-class Network:
-    @classmethod
-    def from_lomap(self, lomap_file):
-        pass
 
-
-def generate_radial_graph(ligands, central_ligand, mappers, extra_scorers=None):
+def generate_radial_network(ligands: Iterable[Molecule],
+                            central_ligand: Molecule,
+                            mappers: Iterable[AtomMapper], scorer=None):
```
IAlibay, line 11: returns Network type annotation?

richardjgowers:
There’s definitely a more generalised thing we’re dancing around, but I’d
rather get this convenience function done and move on for now.

PR #40, opened 2022-02-04
[WIP] Network container class

dwhswenson:
Initial version up for comment while I write tests. Note that the diffs are nonsense (tries to match with renamed file). Easier to look at what's going on just by looking at b3de73fdc20ac1605e25999e704e2d669df36756.

Closes #14.

PR #43, opened 2022-02-05
Tests for Network object

dwhswenson:
Test suite for code introduced in #40.


dwhswenson:
Huh... I wrote this by continuing the branch I already had opened, which closed in #40. Looks like the PR has included commit comments from that PR. (Was #40 squash-merged?) However, diff is accurate.

IAlibay:
Is this a new Github feature? so weird

IAlibay:
Re-opened in other to try to "power cycle" github into fixing itself - looks like at least CI is working now.

dwhswenson:
> Is this a new Github feature? so weird

No, this happens with squash merges (one of the reasons I don't particularly like them). The changes from my old commits were merged in. GitHub rewrote the history of those commits into one commit on `main`. When I opened a PR that contained some of the old commits, GH knows the conversation history of those old commits, and those commits aren't in `main` (the changes are; the commits aren't) so they show up here. (You can comment on commits I've pushed before they're in a PR, too.) Squash merge pollutes PRs if developers continue branches.

Good idea on the old close-reopen for triggering CI, though. Not sure if that problem was related or not; I have seen CI fail to trigger a few times before.

openfe/tests/setup/test_network.py
```diff
@@ -0,0 +1,239 @@
...
+    """Fixture to allow parameterization of the network test"""
+    network_dct = {
+        'simple': simple_network,
+        'doubled_edge': doubled_edge_network,
+        'singleton_node': singleton_node_network,
+    }
+
+    # assertion fails if there is a mismatch between _NETWORK_NAMES and the
+    # keys in the dict by this fixture (error in test suite implementation)
+    assert set(network_dct) == set(_NETWORK_NAMES)
```
IAlibay, line 98: I guess there's no easy way around this (at least none that is obvious to me at this time of the day). It would be good to label this assert's error so it's easy to work out that the error is implementation based not a test failure.

```suggestion
    errmsg = ("test suite implementation error - `_NETWORK_NAMES` does not "
              "match `all_networks` dictionary")
    assert set(network_dct) == set(_NETWORK_NAMES), errmsg
```

IAlibay:
> No, this happens with squash merges (one of the reasons I don't particularly like them). 
Weird, I've never seen that happen on the core MDA repo (where everything defaults to squash-merge).

openfe/tests/setup/test_network.py
```diff
@@ -0,0 +1,239 @@
...
+class TestNetwork:
+    @pytest.mark.parametrize('network_name', _NETWORK_NAMES)
+    def test_graph(self, all_networks, network_name):
+        # The NetworkX graph that comes from the ``.graph`` property should
+        # have nodes and edges that match the Network container object.
+        container = all_networks[network_name]
+        graph = container.network.graph
+        assert len(graph.nodes) == container.n_nodes
+        assert set(graph.nodes) == set(container.nodes)
+        assert len(graph.edges) == container.n_edges
```
IAlibay, line 112: So unless I'm missing something, none of the tests here seem to test that nothing untowards happened on `Network` creation? Ideally we should have a test that specifically checks that we have the edges defined in `std_edges`, the extra node in `singleton_node_network` etc....

It's a bit pedantic, but ideally I'd want to know in the future that some `init` change didn't break my inputs somehow?

IAlibay (changes requested):
For the most part lgtm, just the two comments.

openfe/tests/setup/test_network.py
dwhswenson: Is the test against `set(container.edges)` and `set(container.nodes)` not sufficient? Those are assigned to the correct value of `std_edges`, `list(mols) + [extra_mol]` etc. in the fixtures. Maybe it isn't clear that the `container` designed to package the created `network` as well as (and independently) the expected `nodes` and `edges`?

IAlibay: You're 100% right, I completely misread what was going on here.

openfe/tests/setup/test_network.py
```diff
@@ -0,0 +1,239 @@
...
+    nodes: Iterable[Molecule]
+    edges: Iterable[AtomMapping]
+
+    @property
+    def n_nodes(self) -> int:
+        return len(self.nodes)
+
+    @property
+    def n_edges(self) -> int:
+        return len(self.edges)
```
richardjgowers, line 24: This seems a little overengineered, a nametuple would probably work just as well

openfe/tests/setup/test_network.py
```diff
@@ -0,0 +1,239 @@
...
+    simple_network,
+    doubled_edge_network,
+    singleton_node_network,
+):
+    """Fixture to allow parameterization of the network test"""
+    network_dct = {
+        'simple': simple_network,
+        'doubled_edge': doubled_edge_network,
+        'singleton_node': singleton_node_network,
+    }
```
richardjgowers, line 94: This looks like the problem of trying to have a fixture of fixtures, which is an annoyance of pytest.  A perhaps cleaner way of doing this is through `request`:

```python
File Edit Options Buffers Tools Python Help                                                                                                 
import pytest                                                                                                                               
                                                                                                                                            
                                                                                                                                            
@pytest.fixture                                                                                                                             
def a():                                                                                                                                    
    return 'a'                                                                                                                              
                                                                                                                                            
@pytest.fixture                                                                                                                             
def b():                                                                                                                                    
    return 'b'                                                                                                                              
                                                                                                                                            
@pytest.fixture                                                                                                                             
def c():                                                                                                                                    
    return 'c'                                                                                                                              
                                                                                                                                            
@pytest.fixture(params=['a', 'b', 'c'])                                                                                                     
def myfixture(request, a, b, c):                                                                                                            
    if request.param == 'a':                                                                                                                
        yield a * 3                                                                                                                         
    elif request.param == 'b':                                                                                                              
        yield b * 3                                                                                                                         
    elif request.param == 'c':                                                                                                              
        yield c * 3                                                                                                                         
                                                                                                                                            
                                                                                                                                            
def test_this(myfixture):                                                                                                                   
    assert myfixture == 'aaa'                                                                                                                                                                                            
```

The `myfixture` fixture will make any test using it run 3 times, once for each param.  I think you can use this pattern to clean up the tests below and remove the `assert` inside the `assert` too

openfe/tests/setup/test_network.py
dwhswenson: @richardjgowers wrote:

> I wouldn't want an `assert` outside of a test function. We don't have to test the tests

(before I accidentally deleted his comment instead of deleting the draft comment I'd put in a "start a review" -- sorry Richard! Returning it here for posterity)

richardjgowers (approved):

IAlibay (approved):
Sorry didn't realise I was review blocking, lgtm!

PR #44, opened 2022-02-07
Add a simple coveragerc to ignore tests in reporting

IAlibay:
As per the title - I noticed we were including tests in our coverage.

I know some folks advocate for and some against - debate if you wish.

FWIW - I'm in the against purely because I feel like including tests isn't really helping me with anything except inflate the total %.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[report]
+omit =
+    */openfe/tests/*
+
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,6 @@
 [report]
 omit =
     */openfe/tests/*
-
+exclude_lines =
+    pragma: no cover
+    raise NotImplementedError
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,4 +3,6 @@ omit =
     */openfe/tests/*
 exclude_lines =
     pragma: no cover
+    pragma: no-cover
+    -no-cov
     raise NotImplementedError
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,4 @@
 [report]
-omit =
-    */openfe/tests/*
 exclude_lines =
     pragma: no cover
     pragma: no-cover


dwhswenson:
I'm a recent convert to the other side. Ned Batchelder converted me. When the guy behind `coverage.py` makes the case... 

https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html

I can already say that on projects where I've made this switch, I've caught a few bad tests.

On inflating, we should be getting to 100% ASAP and then never dropping below that. Nothing to inflate if we're already at 100%!

I would like an ignore lines though.... I feel like the standard is `pragma: no-cover`, but I've always liked the simpler `-no-cov-` as a safe, readable, and short string.

IAlibay:
> I would like an ignore lines though.... I feel like the standard is pragma: no-cover, but I've always liked the simpler -no-cov- as a safe, readable, and short string.

Done - added `pragma: no cover`, `pragma: no-cover`, `-no-cov-` which hopefully will suit everyone's tastes?

I've also added an ignore on `NotImplementedError` purely out of reaction for #43 where codecov made me look at those two lines in network.py...

IAlibay:
> I'm a recent convert to the other side. Ned Batchelder converted me. When the guy behind `coverage.py` makes the case...
> 
> https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html
> 
> I can already say that on projects where I've made this switch, I've caught a few bad tests.
> 
> On inflating, we should be getting to 100% ASAP and then never dropping below that. Nothing to inflate if we're already at 100%!

Those are all perfectly acceptable points - I don't have a strong opinion here so I'm happy keeping coverage going for tests. I'll let @richardjgowers and @mikemhenry comment (hopefully so we can form a policy for repos going forward).

dwhswenson:
FWIW, here's the PR where I made the switch to include tests in coverage in the OPS CLI. openpathsampling/openpathsampling-cli#69. More details on my own thoughts, and the results we got from making the switch, in there.

mikemhenry:
I also like including tests in coverage, helps to find tests that are not doing what you think they are doing

Issue #23, opened 2022-01-25
User versioneer?


IAlibay:
It would be a lot easier to just let versions be handled automatically, how do we feel about setting up versioneer?

dwhswenson:
Question on versioneer -- I disliked it when I looked into it a few years back, because (IIRC) there was something incorrect in how it handled non-release versions (note that this may have been an issue in how some repo I was looking at *used* versioneer, as opposed to a problem with the tool itself.)

Specifically, the problem I remember was with development versions. After you release `v1.0.0`, the code on `main` should be `v1.0.1.dev0` (which may bump to `v1.1.0.dev0` or even `v2.0.0.dev0` when a sufficient change has been included). You should never have a time where the code on `main` is ahead of a release while the version number is behind the release (which includes the fact that `v1.0.0.dev` is before `v1.0.0`).

I don't remember if this was something that was outside the scope of versioneer, or something that versioneer handled poorly, or just something that was handled poorly in the specific repos I looked at. In any case, I ended up writing [my own toolkit](https://github.com/dwhswenson/autorelease/) to ensure this behavior (and to automate cutting a release on GitHub/deploying to PyPI). I'm not saying that we should necessarily use that toolkit (some parts of the code are a little janky, to be honest), but I do want to be sure we have a solution to the problems that I solved there.

IAlibay:
I _think_ `dev0` is support as part of some of the modified pep440 styles: https://github.com/python-versioneer/python-versioneer/blob/master/details.md It's not super clean though.

But yeah I'm not too fussed whichever way we go here, as long as we have a detailed workflow and can go `openfe.__version__`.

dwhswenson:
So, here's my understanding from the Versioneer docs (and I'm pretty sure this is what I didn't like): Say we've released 1.0, and we've merged in some major feature that will be part of 1.1. What version does `main` represent at that point? To me, that's clearly the `1.1.0.dev0` : it's a development version of 1.1. Versioneer treats it as still part of 1.0.

If we go with Versioneer, I could probably live with the `pep440-post` style. It's at least technically correct that what I think of as `1.1.0.dev0` could also be (by ordering) `1.0.0.post25`, and it includes the commit hash. Some of the other styles don't respect PEP440 ordering, which I would find very annoying.

mikemhenry:
Is there something other than Versioneer? I've been happy with it, we do want something that is automated.

richardjgowers:
[...]

PR #45, opened 2022-02-08
Add Versioneer

mikemhenry:
Fixes issue #23 

We need to decide how we want to tag releases 
https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md
so we can set `tag_prefix` correctly.

```
tag_prefix:

a string, like 'PROJECTNAME-', which appears at the start of all VCS tags. If your tags look like 'myproject-1.2.0', then you should use tag_prefix='myproject-'. If you use unprefixed tags like '1.2.0', this should be an empty string, using either tag_prefix= or tag_prefix=''.
```


mikemhenry:
Do we want to have pep8speaks ignore `openfe/_version.py` and `versioneer.py` or should I fix them (easy with black or something)

setup.cfg
```diff
@@ -35,3 +35,10 @@ test =
 
 [bdist_wheel]
 universal = 1
+
+[versioneer]
+VCS = git
+style = pep440-post
+versionfile_source = openfe/_version.py
+versionfile_build = openfe/_version.py
+tag_prefix = "" # see https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md
```
IAlibay, line 44: This is causing issues
```suggestion
tag_prefix = '' # see https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md
```

IAlibay (commented):
Couple comments

IAlibay:
> Do we want to have pep8speaks ignore `openfe/_version.py` and `versioneer.py` or should I fix them (easy with black or something)

So I believe for the most part pep8speaks shouldn't complain beyond this one PR (unless we change something). But if you feel like autoformatting it shouldn't have any impact (unless we need to patch the file for some reason in the future)?

setup.cfg
IAlibay: Actually I wonder if it's more the comment afterwards that is causing issues (given it's attempted to add quotes around it)

richardjgowers:
not too fussed on version string style, "v1.2.3" is ok, I'm not offended if we include the package name though "openfe-v1.2.3"

setup.cfg
IAlibay: Sorry I 100% missed that this was probably intentional 🤦🏽‍♂️

setup.py
```diff
@@ -1,2 +1,8 @@
 from setuptools import setup
-setup()
+
+import versioneer
+
+setup(
+    version=versioneer.get_version(),
```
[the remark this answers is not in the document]
mikemhenry: I think it is needed per the installation instructions: https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md

```
add a import versioneer to your setup.py, and add the following arguments to the setup() call:

  version=versioneer.get_version(),
  cmdclass=versioneer.get_cmdclass(),
```

dwhswenson:
Re: pep8speaks. I'd vote for adding the ignore. It's a vendored file; I usually try to leave those as-is if possible. Keeps sane diffs for updates, etc. Also, any future updates shouldn't trigger pep8speaks.

setup.py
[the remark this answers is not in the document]
mikemhenry: oops, you said `setup.cfg` yes that needs to go!

setup.cfg
```diff
@@ -35,3 +35,10 @@ test =
 
 [bdist_wheel]
 universal = 1
+
```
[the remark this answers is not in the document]
mikemhenry: Good catch!

setup.cfg
mikemhenry: Haha, I left that comment in there to remind myself when I made the PR to ask about it. I will fix it.

mikemhenry:
If pep8speaks doesn't complain about the files again, then I won't bother adding them to an ignore file. RE tagging, I'm fine with just tagging the version number, I'm not sure what value adding the `openfe` prefix brings to the tag.

IAlibay:
> If pep8speaks doesn't complain about the files again, then I won't bother adding them to an ignore file. RE tagging, I'm fine with just tagging the version number, I'm not sure what value adding the `openfe` prefix brings to the tag.

Yeah pretty sure that by default it only acts on diffs, see: https://github.com/OrkoHunter/pep8speaks/blob/master/data/default_pep8speaks.yml

richardjgowers:
thanks @mikemhenry !

PR #48, opened 2022-02-10
update AtomMapping type annotations

richardjgowers:
autoconvert rdkit molecules to Molecules inside AtomMapping

added check to Network node type

### DIFF
--- a/openfe/setup/atommapping.py
+++ b/openfe/setup/atommapping.py
@@ -1,8 +1,7 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/openfe
 from dataclasses import dataclass
-from rdkit import Chem
-from typing import Dict, Union, TypeVar
+from typing import Dict, TypeVar
 
 from openfe.setup import Molecule
 
@@ -25,16 +24,6 @@ class AtomMapping:
     mol1: Molecule
     mol2: Molecule
     mol1_to_mol2: Dict[int, int]
-
-    def __init__(self, mol1: Union[Molecule, RDKitMol],
-                 mol2: Union[Molecule, RDKitMol], mol1_to_mol2):
-        if isinstance(mol1, Chem.Mol):
-            mol1 = Molecule(mol1)
-        self.mol1 = mol1
-        if isinstance(mol2, Chem.Mol):
-            mol2 = Molecule(mol2)
-        self.mol2 = mol2
-        self.mol1_to_mol2 = mol1_to_mol2
 
     def __hash__(self):
         return hash((hash(self.mol1), hash(self.mol2),
--- a/openfe/setup/atommapping.py
+++ b/openfe/setup/atommapping.py
@@ -4,6 +4,8 @@ from dataclasses import dataclass
 from rdkit import Chem
 from typing import Dict, Union, TypeVar
 
+from openfe.setup import Molecule
+
 RDKitMol = TypeVar('RDKitMol')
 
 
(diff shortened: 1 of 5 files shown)


richardjgowers:
@dwhswenson I think in the end we decided that the `Molecule` was a necessary evil right?

openfe/setup/atommapping.py
```diff
@@ -13,17 +15,27 @@ class AtomMapping:
...
       If this atom has no corresponding atom, None is returned.
 
     """
-    mol1: RDKitMol
-    mol2: RDKitMol
+    mol1: Molecule
+    mol2: Molecule
     mol1_to_mol2: Dict[int, int]
 
+    def __init__(self, mol1: Union[Molecule, RDKitMol],
```
richardjgowers, line 29: So currently AtomMappers work off an RDKit Mol, which seems correct, and they return an `AtomMapping`.  I think it's probably easiest to do the conversion from rdkit to hashable Molecule here, rather than have the boilerplate in every single `AtomMapper`.

richardjgowers:
@dwhswenson I think a warning would be annoying, besides `mol1 == anything` always fails, that's why we have this annoying wart.

dwhswenson:
This is why I think we need to go all-in on a `Molecule` class. It seems like you're expecting that users will be coming to us with a molecule in RDKit format. I don't understand why that would be the case. I guess I don't see the issues with the user-facing setup being something like:

```python
ligands = [Molecule.from_file(fname) for fname in filenames]  # return list of Molecule
my_ligand = Molecule.from_smiles("CC")    # simple wrapper for RDkit
from_rdkit = Molecule(rdkit=rdkit_mol)
```

All of those bring them into our Molecule class in a way that is explicit to the user, and not too tedious, IMO. Since we need the `name` attribute for hashing, that can be an optional parameter in construction, making it much easier for users than `rdkitmol.setProp("_Name")` (where you'd better remember the underscore and the capital N!) For cheminfo needs, the `rdkit` version of the molecule is accessible at `molecule.rdkit`. Furthermore, this has easy extensibility to other toolkits in the future (if we have lossless conversion).

`AtomMapper` is already separated into a user-facing `suggest_molecules` and an abstract `_mappings_generator`. So we create a subclass of `AtomMapper`, called `RDKitAtomMapper`. Change it so that `suggest_mappings` always takes our `Molecule`. For `RDKitAtomMapper`, it pulls out the `.rdkit` attribute, and passes that to the existing `_mappings_generator`. Implementers inherit from `RDKitAtomMapper`; no boilerplate.

I can tell that something about this approach doesn't sit well with you, but I don't know what it is. To me, this seems to solve the problems we're running into, and it is designed for a future where users may want options other than RDkit. Is there an problem with this approach that I'm missing?

richardjgowers:
@dwhswenson ok I'm convinced.  So removed the `_Name` idiom for just a kwarg, added a test to Network to check dtype of the nodes and updated the AtomMapping type annotations.  I'm not convinced on the `from_file` and `from_SMILES` classmethods (I'd rather use `Molecule(Chem.MolFromSmiles('CC'))`) but that's a debate for another time

openfe/setup/molecule.py
```diff
@@ -14,10 +14,14 @@ class Molecule:
     ----------
     rdkit : rdkit.Mol
         rdkit representation of the molecule
+    name : str, optional
+        if multiple Molecules with identical SMILES but differing positions
+        are used, a name must be given to differentiate these.  This name
+        will be used in the hash.
     """
-    def __init__(self, rdkit: RDKitMol):
+    def __init__(self, rdkit: RDKitMol, name: str = None):
```
dwhswenson, line 22: Not sure about typing here -- is `None` allowed implicitly as `str`? Any reason not to default to `""` anyway?

dwhswenson (approved):
Small change recommended, (default `None` to `""`), but looks basically good.

As for `from_SMILES`, etc. -- I figure we add conveniences like that if they end up being useful for users. First we make things possible, then we make things easier.

openfe/setup/molecule.py
richardjgowers: I thought it'd be like that case where a empty list is mutable, but it's not

PR #52, opened 2022-02-11
Allow Network to reference its own type in type annotation of classme…

richardjgowers:
…thod

https://stackoverflow.com/a/53450349/14693893

I reckon this will sort the mypy failures


dwhswenson (approved):
Thanks... my error in the code. Now I know the correct workaround!

LGTM (assuming the CI turns green)

PR #53, opened 2022-02-11
omit _version.py from coverage

IAlibay:
iirc you only need to omit in [run] and it should carry though - let's see what codecov thinks

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,7 @@
+[run]
+omit =
+    */_version.py
+
 [report]
 exclude_lines =
     pragma: no cover


dwhswenson:
I always only omit in `[report]`, but it's possible that either way works. 🤷‍♂️

IAlibay:
> I always only omit in `[report]`, but it's possible that either way works. 🤷‍♂️

So the coveragerc documentation has never been clear enough for me to know what on earth it does in the background, but I've always understood that adding stuff in [run] prevents coverage from running on that file to begin with, so it saves some CI time. I've never tested it and I have a pretty good chance of being wrong 🙃 - in any case it seems to have gotten rid of the file.

I'm happy to make changes if anyone has strong feelings about it all. There is also the question as to whether or not we should just add the contents of coveragerc directly in setup.cfg.

richardjgowers (approved):

dwhswenson:
> There is also the question as to whether or not we should just add the contents of coveragerc directly in setup.cfg.

I tend to prefer keeping these sorts of things separate, otherwise setup.cfg becomes a huge mess and it's hard to find stuff for each specific tool.

dwhswenson (approved):

PR #55, opened 2022-02-14
Lomap scorers

richardjgowers:
Ported lomap scorers into openfe

These are all `1 - score` compared to lomap.  There's also a `default_lomap_score` which recreates their product combination.


openfe/setup/lomap_mapper.py
```diff
@@ -55,3 +56,52 @@ def _mappings_generator(self, mol1, mol2):
...
+    def _get_mcs(self, mapping):
+        # get mcs from cache, else create and place into cache
+        try:
+            mcs = self._mcs_cache[mapping]
+        except KeyError:
+            mcs = lomap_mcs.MCS(mapping.mol1.rdkit, mapping.mol2.rdkit,
+                                self.time, threed=self.threed,
+                                max3d=self.max3d)
+            self._mcs_cache[mapping] = mcs
+        return mcs
```
richardjgowers, line 69: The object we need from MCS is the `mcs_mol`.  There might be a shortcut from using the indices in the mapping to recreate this without rerunning the MCS alg.

openfe/setup/lomap_mapper.py
```diff
@@ -55,3 +82,219 @@ def _mappings_generator(self, molA, molB):
...
+
+    def _get_mcs(self, mapping):
+        # get mcs from cache, else create and place into cache
+        try:
+            mcs = self._mcs_cache[mapping]
+        except KeyError:
+            mcs = lomap_mcs.MCS(mapping.molA.to_rdkit(), mapping.molB.to_rdkit(),
+                                self.time, threed=self.threed,
+                                max3d=self.max3d)
+            self._mcs_cache[mapping] = mcs
```
dwhswenson, line 94: also put this caching into `_mappings_generator`

openfe/tests/setup/test_lomap_scorers.py
```diff
@@ -0,0 +1,140 @@
...
+        molB = smallmols[j]
+
+        mapping = next(mapper.suggest_mappings(molA, molB))
+        score = scorer(mapping)
+
+        scores[i, j] = scores[j, i] = score
+    scores = 1 - scores
+    # fudge diagonal for comparison
+    for i in range(matrix.shape[0]):
+        scores[i, i] = 0
```
dwhswenson, line 138: wasn't this set in `zeros_like`?

openfe/setup/lomap_mapper.py
```diff
@@ -6,12 +6,38 @@
 
 """
 from lomap import mcs as lomap_mcs
+import math
+import sys
+from typing import Dict
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+     from typing_extensions import TypeAlias
```
dwhswenson, line 15:
```suggestion
from openfe.utils.custom_typing import TypeAlias
```

dwhswenson (changes requested):
* please fix some of the pep8 complaints
* please add the caching when the mapping is generated
* why are we reimplementing so much of LOMAP? I expected everything to just be `return 1 - self.get_mcs(mapping).score_name()`

richardjgowers:
@dwhswenson the reimplementation of each score is so that any mapping regardless of source can be scored.  Currently with the mapper / scorer baked together to get lomap scorers to give scores on foreign mappings would require either a) filling the `lomap.MCS` data structures as if it generated the mapping itself or b) just reimplement.  Since the internals of `lomap.MCS` are a little complex, it's quicker/simpler just to reimplement.

dwhswenson:
Makes sense, but one thing I find a little weird is that the API is different for different scorers if you aren't using the LOMAP mapper.

If a user wants to use one of these reimplemented scorers, it is

```Python
scorer=LomapAtomMapper.reimplemented_scorer
```

If they want to use one that actually uses the MCS, it is

```Python
scorer=LomapAtomMapper().mcs_scorer
```

I suppose they could always use it with instantiating, and maybe that's what we should always do in documentation. It's just that I would usually lean toward not creating the instance for a static method.

Should they maybe be implemented as be completely separate functions (that could be a clearer way to indicate that these aren't just relevant to LOMAP?) A convenience in LomapAtomMapper could be included as well.

openfe/tests/setup/test_lomap_scorers.py
richardjgowers: Yes, but we do `1 - scores` above

richardjgowers:
Ok went the long way which @dwhswenson suggested - every scorer is now reimplemented in `lomap_scorers` module.  There's a full back to back test against the basic lomap data files going through their implementation of scorers.

openfe/tests/setup/test_lomap_scorers.py
```diff
@@ -0,0 +1,251 @@
...
```
richardjgowers, line 249: This is the code snippet that recreates lomap scoring matrix with our machinery

dwhswenson: If our network object already has scores associated, can we just use nx for this? https://networkx.org/documentation/stable/reference/generated/networkx.convert_matrix.to_numpy_array.html

dwhswenson (approved):
LGTM. I made a few small changes in [3e17e2c](https://github.com/OpenFreeEnergy/openfe/pull/55/commits/3e17e2c2cfbbaea5962532d0a3b8b5d6cff6bf55) to make refactoring a little easier in the future, based on the plan:

1. Move these into the LOMAP repo as bare functions that take RDKit molecules and the dict mapping.
2. Change the `lomap.MCS` instance methods for these to use those bare functions.
3. Change the code here to be simple wrappers for those LOMAP bare functions (take our `LigandAtomMapping`, unwrap the RDKit mols and the dict mapping, and pass those to the underlying function).

There are a couple lines missing coverage, but I'm not going to block on that now. Maybe when we move it back to LOMAP we can try to add some extra tests.

PR #56, opened 2022-02-15
Proposed API change on AtomMapper

dwhswenson:
This proposes a change to the API for `AtomMapper` so that the input to `AtomMapper.suggest_mappings` is an `openfe.setup.Molecule` instead of an `rdkit.Mol`, and so that the output of the internal `_mapping_generator` is a `dict`, instead of the `AtomMapping`. A few reasons for this:

1. From a user perspective (while writing CLI code) it felt strange that the input to this was rdkit. Example: consider that we might want to visualize an AtomMapping. This could either be an atommapping we generate on the fly using `AtomMapper`, or one that we load from an existing Network (selecting the edge we're interested in). In these two different use cases, we would need to use different molecule classes -- rdkit for the first, openfe for the second. That was confusing, and supporting both use cases would complicate code for the CLI. Making this change simplifies.
2. From the perspective of an implementer, the choice to only return a `dict` means that the `AtomMapper` implementer doesn't have to think about our objects at all. They take rdkit and return the dict; we take care of wrapping it into an `AtomMapping`.

Tossed this together quickly, so I'll need to double-check that docstrings were all updated.

### DIFF
--- a/openfe/setup/atommapper.py
+++ b/openfe/setup/atommapper.py
@@ -38,7 +38,7 @@ class AtomMapper:
         ))
 
     def suggest_mappings(
-        self, mol1: RDKitMol, mol2: RDKitMol
+        self, mol1: Molecule, mol2: Molecule
     ) -> Iterable[AtomMapping]:
         """
         Suggest :class:`.AtomMapping` options for the input molecules.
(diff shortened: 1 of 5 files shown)


richardjgowers:
I got deja vu reading this then realised you'd made similar changes to me in #41. I think your use of the breakpoint we have behind suggest mappings is better though. Can you push up this change to that PR?

dwhswenson:
> Can you push up this change to that PR?

Will do.

richardjgowers (approved):

PR #58, opened 2022-02-16
added stubs for Molecule conversions

richardjgowers:
added Molecule conversion methods


openfe/setup/molecule.py
```diff
@@ -42,3 +55,21 @@ def __hash__(self):
 
     def __eq__(self, other):
         return hash(self) == hash(other)
+
+
+def as_rdkit(func):
+    @wraps(func)
+    def inner(*args, **kwargs):
+        args = (a.rdkit if isinstance(a, Molecule) else a for a in args)
```
dwhswenson, line 63: definitely need to treat `kwargs` as well (in this and other decorator)

dwhswenson (changes requested):
I'm a little unsure about implicitly treating all arguments with this, although I suppose a user who really needs both our `Molecule` and `rdkit.Mol` can just avoid using the decorator (`mol.rdkit` is not a heavy piece of boilerplate).

A couple alternatives that might be worth considering:

1. Require argument names in the decorator, so you decorate as `@as_rdkit('mol1', 'mol2')`. For implementation, [`inspect.Signature.bind`](https://docs.python.org/3/library/inspect.html#inspect.Signature.bind) can do most of the heavy lifting to correctly associate argument name with value passed by user, whether in `args` or `kwargs`.
2. If we only need this on AtomMappers, then it might be more straightforward to subclass as `RDKitAtomMapper`, etc. (But I suspect may be are other places where this will be useful, even if my tired brain can't think of them this morning.)

Not that either of those are necessarily better, but they're worth consideration.

I'm definitely +1 on the `Molecule.oechem`, `Molecule.openff` properties.

Handling kwargs in the decorators is the only thing I see as a required change. We can't know whether the user is passing the molecules as args or kwargs.

richardjgowers:
Maybe a subclass for a particular toolkit is a more intuitive way for
people to write these too…

environment.yml
```diff
@@ -10,6 +10,7 @@ dependencies:
   - pytest-xdist
   - pytest-cov
   - coverage
+  - openff-toolkit
```
IAlibay, line 13: add `jaimergp/label/unsupported-cudatoolkit-shim` to channels above and `channel-priority: strict` to the conda-incubator/setup-miniconda@v2 in the ci.yaml file?

richardjgowers: Nah I think we're just sucking it up and installing cuda

openfe/tests/setup/test_molecule.py
```diff
@@ -1,5 +1,12 @@
+import openff.toolkit.topology
 import pytest
 
+try:
+    from openeye import oechem
+except ImportError:
+    HAS_OECHEM = False
+else:
+    HAS_OECHEM = oechem.OEChemIsLicensed()
```
richardjgowers, line 9: Looking at the coverage, it looks like the import is working but the license check is failing.  Is this something @mikemhenry that you can help with?

mikemhenry: :+1:

mikemhenry: I think this part needs to be moved up

```
diff --git a/.github/workflows/ci.yaml b/.github/workflows/ci.yaml
index 14bc09c..b279758 100644
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,6 +45,13 @@ jobs:
             environment-file: environment.yml
             activate-environment: openfe
 
+      - name: Decrypt OpenEye license
+        env:
+          OE_LICENSE_TEXT: ${{ secrets.OE_LICENSE }}
+        run: |
+          echo "${OE_LICENSE_TEXT}" > ${OE_LICENSE}
+          python -c "import openeye; assert openeye.oechem.OEChemIsLicensed(), 'OpenEye license checks failed!'"
+
       - name: "Install"
         run: python setup.py develop --no-deps
 
@@ -61,13 +68,6 @@ jobs:
           mamba info -a
           mamba list
 
-      - name: Decrypt OpenEye license
-        env:
-          OE_LICENSE_TEXT: ${{ secrets.OE_LICENSE }}
-        run: |
-          echo "${OE_LICENSE_TEXT}" > ${OE_LICENSE}
-          python -c "import openeye; assert openeye.oechem.OEChemIsLicensed(), 'OpenEye license checks failed!'"
-
       - name: "Run tests"
         run: |
           pytest -n 2 -v --cov=openfe --cov-report=xml openfe/tests/
```

mikemhenry: git has gotten a bit weird, I can commit this fix in a bit, but need to fix my branch which has some merge conflicts with upstream

mikemhenry: Cool looks like you got it it with [69dbfd8](https://github.com/OpenFreeEnergy/openfe/pull/58/commits/69dbfd862ae6429d8c8dbe0a611904916a5c2720)

.github/workflows/ci.yaml
```diff
@@ -25,11 +25,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu', 'windows', 'macos']
+        os: ['ubuntu', 'macos']
```
dwhswenson, line 28: Is it absolutely necessary to drop Windows support?

openfe/setup/molecule.py
```diff
@@ -1,10 +1,12 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/openfe
-
+import warnings
+with warnings.catch_warnings():
```
dwhswenson, line 4: add a comment as to what the warning you're catching is? (so we know when we can remove this)

dwhswenson (commented):
Main issue is that I don't think its good to drop Windows support entirely. At least, I will entirely rely on CI to tell me that code still runs on Win, and I thought that was an important target for our users, especially in the setup stage.

Otherwise, please change the title -- these are no longer stubs; it's the full molecule conversion! (At least, in the "to" direction.) 😊

environment.yml
IAlibay: are we treating environment.yml as something users will use?

dwhswenson: I would think perhaps code contributors, but not users -- env for users is defined by the recipe in the feedstock

IAlibay: In that case I'm not sure why we would needlessly slowdown CI by not including the shim?

.github/workflows/ci.yaml
IAlibay: > Main issue is that I don't think its good to drop Windows support entirely.
Yikes didn't see that change on first reivew.

I agree with you here @dwhswenson. Plus I think that Windows is pretty important for industrial users too, but it seems like OFF has just decided not bother: https://open-forcefield-toolkit.readthedocs.io/en/latest/installation.html#os-support

Maybe in the first instance we drop support, and check if it's a deal breaker on Tuesday? Mainly saying this as "we probably want this in before Tuesday, and dealing with whatever breaks in CI before the release is going to be more time than we probably can afford it".

richardjgowers:
Yeah Windows was dropped because Ambertools didn’t have a package, so it
won’t be a fix done before Tuesday.

.github/workflows/ci.yaml
dwhswenson: Fair enough... I was hoping that this was maybe an OpenEye issue, not OpenFF.

As I recall, our plans specifically state that we will have support for using OpenFF. I don't remember anything about Windows support. 🤷‍♂️ Calling this resolved, with the hopes that someday Windows support comes back.

dwhswenson:
> @dwhswenson what do you think about renaming the `.rdkit` etc to `to_rdkit()` to a) match OFF and b) make it clearer that a copy is happening?

Mixed feelings. I see the value in matching OFF. Do we actually intend this to be a copy, though? I thought we were going to cache it. (I did the copy on the input via `from_rdkit`).

That said, if we make it a copy on output, then I think making it a method, not property, is required. Otherwise you get `mol.rdkit is not mol.rdkit`.

richardjgowers:
Yeah so the deliberate copy means we don’t have to track what changes are
happening to the object downstream, so accessing .rdkit then modifying it
shouldn’t modify the originating Molecule. This then means we aren’t
promising which object is our internal representation

dwhswenson:
Yeah, these are good points on why returning a copy is a good idea. In that case, I'm strongly in favor of `to_rdkit()`. I think that if we return a copy, it should be `mol.to_rdkit()`; if we cache the object once created and always return the cached, it should be `mol.rdkit`.

openfe/setup/molecule.py
```diff
@@ -20,14 +28,37 @@ class Molecule:
...
+    def to_openff(self) -> OFFMolecule:
+        """OpenFF Toolkit representation of this molecule"""
+        m = OFFMolecule(self._rdkit, allow_undefined_stereo=True)
+        m.name = self.name
+
+        return m
+
+    @classmethod
+    def from_openff(cls, openff: OFFMolecule, name: str = ""):
+        raise NotImplementedError
```
richardjgowers, line 61: @dwhswenson so what I've done here is everything is a copy.  So on creation, we copy the rdkit Mol, and all to_X are copies.  The Molecule is then therefore read-only (and documented as such).  Handling manipulation of a Molecule is out of scope for us (i.e. if you want to change a Molecule, recreate it).

dwhswenson: 100% agree with the approach. Reviewing details now.

openfe/setup/molecule.py
```diff
@@ -20,14 +28,37 @@ class Molecule:
...
-        """RDKit representation of this molecule"""
-        return self._rdkit
+    def to_rdkit(self) -> RDKitMol:
+        """Return an RDKit copied representation of this molecule"""
+        return Chem.Mol(self._rdkit)
+
+    @classmethod
+    def from_rdkit(cls, rdkit: RDKitMol, name: str = ""):
+        """Create a Molecule copying the input from an rdkit Mol"""
+        return cls(rdkit=Chem.Mol(rdkit), name=name)
```
dwhswenson, line 42:
```suggestion
        return cls(rdkit=rdkit, name=name)
```

since we copy in the `__init__`, no need to double the copying here.

openfe/tests/setup/test_molecule.py
```diff
@@ -1,5 +1,12 @@
+import openff.toolkit.topology
 import pytest
 
+try:
+    from openeye import oechem
+except ImportError:
```
dwhswenson, line 6:
```suggestion
except ImportError:  # -no-cov-
```

looks like this line codecov is complaining about this; maybe we add something for the test matrix to test against a minimal set of reqs? For now, ignore coverage here.

dwhswenson (approved):
2 minor suggestions; otherwise LGTM. Please make those changes unless they don't make sense, but marking this approved; I don't think I need to review after changes.

PR #60, opened 2022-02-18
Add visualization to AtomMapping

### LINKED ISSUE #37: Add method to visualise an atom mapping
Essentially we'd want our own version of what perses does here: https://github.com/choderalab/perses/blob/a8b490af0e4682082dfa442717ef45c1bd61f538/perses/rjmc/atom_mapping.py#L183-L310

See here for perses notebook: https://github.com/choderalab/perses/blob/master/examples/atom-mapping/Atom-mapping.ipynb

Use case: "I'm sitting in a notebook and I want to cycle through my atom map and see what's going on in each edge".

mikemhenry:
Fixes #37 

Here is a little snippet to run in a notebook to see the visualization:

```
from rdkit import Chem
from openfe.setup.lomap_mapper import LomapAtomMapper


mol_1_smi = "c1ccccc1"  # benzene
mol_2_smi = "Oc1ccccc1"  #phenol
mol_1, mol_2 = Chem.MolFromSmiles(mol_1_smi), Chem.MolFromSmiles(mol_2_smi)
mol_1, mol_2 = Chem.AddHs(mol_1), Chem.AddHs(mol_2)
Chem.rdDepictor.Compute2DCoords(mol_1), Chem.rdDepictor.Compute2DCoords(mol_2)

mapper = LomapAtomMapper()
mapping = next(mapper.suggest_mappings(mol_1, mol_2))


mapping.visualize()
````

I haven't added tests yet since I wasn't sure if this is where we wanted to wire this method in.
I also need to wire this into the cli, but we should discuss what we want that to look like as well.

### CODE REVIEW

File: openfe/setup/atommapping.py [hunk 1]
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
+            highlightAtomColors=atom_colours,
+        )
+        d2d.FinishDrawing()
+        return d2d.GetDrawingText()
+
+    def visualize(self):
+        """Visualize atom mapping in a Jupyter Notebook"""
+        from IPython.display import Image
+
+        return Image(self._draw_mapping())
```
IAlibay, line 122: So as far as I remember the 2DCairo backend should just output a raw stream for a PNG, maybe we can make this an optional return where we either save to file or visualise?

IAlibay (commented):
Thanks @mikemhenry ! Couple of questions up for discussion.

For the wider discussion of "where should this go", I quite like having this as part of AtomMapping since I can directly inspect each object, however I could see this being in Network too if folks prefer a less bloated dataclass.

mikemhenry:
RE mypy `openfe/setup/atommapper.py:62: error: Argument 3 to "AtomMapping" has incompatible type "Dict[int, int]"; expected "AtomMap"` Where should we store types we will use in multiple files? Maybe `openfe/utils/types.py` ?

openfe/setup/atommapping.py
```diff
@@ -1,10 +1,15 @@
...
 from dataclasses import dataclass
-from typing import Dict, TypeVar
+from itertools import chain
+from typing import Dict, NewType
+
+from rdkit import Chem
 
 from openfe.setup import Molecule
 
+AtomMap = NewType("AtomMap", Dict[int, int])
```
richardjgowers, line 11: I'm not sure what this alias adds?

openfe/setup/atommapping.py
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
-                     tuple(self.mol1_to_mol2.items())))
+        return hash(
+            (hash(self.mol1), hash(self.mol2), tuple(self.mol1_to_mol2.items()))
+        )
 
     @classmethod
     def from_perses(cls, perses_mapping):
         raise NotImplementedError()
+
+    def _match_elements(self, idx: int):
```
richardjgowers, line 41: single line comment describing what methods do would be nice

richardjgowers (changes requested):
We don't need the explicit hydrogens for atommapping or visualisation (and it might even be making things worse tbh)

IAlibay:
> We don't need the explicit hydrogens for atommapping or visualisation 

I disagree, it can be very useful to have, particularly downstream when dealing with topologies. Hydrogens are part of the mapping, visualising without the hydrogens obfuscates what you have in your dictionary imho (and it makes it real hard to know that the atom mapping has worked as intended in some cases).

I'd much prefer have the option to do both here, optionally go implicit if required, but otherwise show the whole map.

As additional justification - Perses' atom map visulalization also retains all hydrogens.

openfe/setup/atommapping.py
mikemhenry: It saves from having to use `Dict[int, int]` whenever we have an atom map type, I was thinking we could have a `openfe/utils/types.py` to keep track of our types, I am new to using typing so if this isn't a good idea/not best practices then we shouldn't do it.

openfe/setup/atommapping.py
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
+            return True
+        else:
+            return False
+
+    def _get_unique_bonds_and_atoms(self, mapping: AtomMap):
+
+        uniques = {
+            "atoms": [],  # atoms which fully don't exist in mol2
+            "elements": [],  # atoms which exist but change elements in mol2
+            "bonds": [],  # bonds involving either unique atoms or elements
```
dwhswenson, line 54: As commented below, consider using sets here. These should have no duplicates and order doesn't matter, right? If so, then `set` is the appropriate container.

openfe/setup/atommapping.py
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
+        }
+
+        for at in self.mol1.GetAtoms():
+            idx = at.GetIdx()
+            if idx not in self.mol1_to_mol2.keys():
```
dwhswenson, line 59:
```suggestion
            if idx not in self.mol1_to_mol2:
```

openfe/setup/atommapping.py
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
+                uniques["elements"].append(idx)
+
+        for bond in self.mol1.GetBonds():
+            bond_at_idxs = [bond.GetBeginAtomIdx(), bond.GetEndAtomIdx()]
+            for at in chain(uniques["atoms"], uniques["elements"]):
+                if at in bond_at_idxs:
+                    bond_idx = bond.GetIdx()
+                    # avoid double inclusion
+                    if bond_idx not in uniques["bonds"]:
+                        uniques["bonds"].append(bond_idx)
```
dwhswenson, line 71: Would it be better to avoid double inclusion just by using a `set` here?

openfe/setup/atommapping.py
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
+        mol2_uniques = self._get_unique_bonds_and_atoms(mol2_to_mol1_map)
+
+        atoms_list = [
+            mol1_uniques["atoms"] + mol1_uniques["elements"],
+            mol2_uniques["atoms"] + mol2_uniques["elements"],
+        ]
+
+        # highlight core element changes differently from unique atoms
+        red = (1, 0, 0, 0.5)
+        green = (0, 1, 0, 1)
```
dwhswenson, line 90: Have these colors been checked as distinct for red-green colorblindness? (there are programs that can basically do this for you -- I use SimDaltonism on my Mac)

openfe/setup/atommapping.py
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
+
+        at1_colours = {}
```
dwhswenson, line 92: Mike, are these guys forcing you to speak British? 😂

openfe/setup/atommapping.py
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
+
+        at2_colours = {}
+        for at in mol2_uniques["elements"]:
+            at2_colours[at] = green
+
+        atom_colours = [at1_colours, at2_colours]
+
+        bonds_list = [mol1_uniques["bonds"], mol2_uniques["bonds"]]
+
+        d2d = Chem.Draw.rdMolDraw2D.MolDraw2DCairo(600, 300, 300, 300)
```
dwhswenson, line 104: I think that this specific class can be replaced with others, e.g., `MolDraw2DSVG`, `MolDraw2DQt` -- you're just using the `MolDraw2D` API, right?

In that case, maybe allow the user to pass an `d2d` as an instance of a `MolDraw2D` to the method (default `None` creates what you have here). That allows this code to be reused for other duck-typed canvases.

openfe/setup/atommapping.py
```diff
@@ -19,14 +24,99 @@ class AtomMapping:
...
+        d2d.DrawMolecules(
+            [self.mol1, self.mol2],
+            highlightAtoms=atoms_list,
+            highlightBonds=bonds_list,
+            highlightAtomColors=atom_colours,
+        )
+        d2d.FinishDrawing()
+        return d2d.GetDrawingText()
+
+    def visualize(self):
```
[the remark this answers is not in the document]
dwhswenson: I would suggest just renaming this method to `_ipython_display`.

openfe/setup/atommapping.py
dwhswenson: Doesn't `_draw_mapping` already give that stream? I think that it would be better to have a separate method `save_to_file` (returns `None`) instead of having a `visualize` method that gives multiple different output types. `save_to_file` might even parse the filename for the extension, and pick how to draw based on that.

dwhswenson (changes requested):
Minor changes, just based on a read of the code. I haven't tried it out yet.

I'm ambivalent on whether to have this as an instance method or a bare function. Either way works for me. I think it will only be used for this class (or subclasses), so instance method makes sense. On the other hand, I tend to keep drawing/plotting routines separate, but that's partly because I usually try to make those requirements optional.

openfe/setup/atommapping.py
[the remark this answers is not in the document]
IAlibay: As long as it's not too ugly to call `_ipython_display_()` from the CLI, I would support that.

openfe/setup/atommapping.py
IAlibay: > Doesn't _draw_mapping already give that stream?

Yeah that's what I meant.

Separate methods sounds good, I didn't want to pollute the class too much, but if we can move `_draw_mapping` et al (they don't really need `self` so we could just keep `_ipython_display_` and move the rest to `utils`), then adding an additional method here doesn't hurt.

openfe/setup/atommapping.py
IAlibay: Yeah ok that's probably on me for suggesting the easiest two _colours_, how about just red / blue? (pretty sure that's good for more or less everyone)
```suggestion
        red = (1, 0, 0, 0.5)
        blue = (0, 0, 1, 1)
```

openfe/setup/atommapping.py
mikemhenry: Sounds good! I'm pretty new to using type hints so I thought it would be good to save having to type `Dict[int, int]` each time but I see your point! I'll remove it.

openfe/setup/atommapping.py
mikemhenry: Good catch! I used this guide here for red and blue:
https://davidmathlogic.com/colorblind/#%23005AB5-%23DC3220

Once we get some branding/colors, we can see if we could use those

openfe/setup/atommapping.py
mikemhenry: Haha, we should have some sort of informal style guide for this. I think we should be consistent and I'm inclined to go with american english :us:

mikemhenry:
Weird, @dwhswenson I can't reply to your comment here:
image
But I can here:
image

mikemhenry:
Anyway, re:
https://github.com/OpenFreeEnergy/openfe/pull/60#discussion_r812311617

I do need to call `Image` to get the PNG to render in the notebook:

image

openfe/setup/atommapping.py
mikemhenry: Having a  `save_to_file` method sounds good, but I like `visualize` being used to "visualize" inside a notebook, so the CLI can call `save_to_file`

openfe/setup/atommapping.py
```diff
@@ -12,21 +15,133 @@ class AtomMapping:
...
     mol1_to_mol2 : dict
       maps the index of an atom in either molecule **A** or **B** to the other.
       If this atom has no corresponding atom, None is returned.
 
     """
-    mol1: Molecule
-    mol2: Molecule
+
+    mol1: Chem.rdchem.Mol
+    mol2: Chem.rdchem.Mol
```
dwhswenson, line 27: Why is this switching back to RDKit? Current code has `AtomMapper` making an `AtomMapping` using our `Molecule` class. I believe that we NEED this to be the case for analysis to work correctly.

https://github.com/OpenFreeEnergy/openfe/blob/41068e018575a4da5743d1f45e494fbbdc2a9921/openfe/setup/atommapper.py#L40-L62

dwhswenson:
> I can't reply to your comment here:

The one you couldn't reply to should have been part of a thread, replying to something Irfan had written. I think that's why. I probably should have actually made that as a separate comment; not batched that into the review.

> I do need to call `Image` to get the PNG to render in the notebook:

The idea would be to change the name of the original method (`visualize`, with the `Image` call) to `_ipython_display_`. This would mean that the notebook would act as if the repr was the image. So instead of a cell with input `AtomMapping(mol1, mol2, mapping)` giving output string `AtomMapping(mol1=....)` (describing the object), you would actually see the mapping visualization. It's the same trick used by pandas to give nice HTML tables in notebooks. This way you wouldn't need to call a function to get the visualization; it's just how the notebook would represent the object.

This is one of those things that, IMO, gets near the edge of being a little too clever for our own good. But it's probably worth trying out to see how it feels in practice.

openfe/setup/atommapping.py
mikemhenry: I thought it would need to be an RDKit mol so that things like `bond.GetBeginAtomIdx()` would work in `_draw_mapping`

dwhswenson: You can get the RDKit version from our Molecule with `mol.rdkit` (or, after #58, `mol.to_rdkit()`). From there you can use RDKit internally.

mikemhenry:
Got it! I didn't realize that `_ipython_display_` was a special name at first and thought it was only a more descriptive name, which didn't make a ton of sense to me since it looked like it was going to be a private function then, but I got it now! (missed that last `_`)

IAlibay:
This is so weird, it's as if secrets aren't available 🤔

dwhswenson:
@IAlibay : This is coming from a fork. Secrets *aren't* available in forks.

Don't know how we want to handle that when it comes to outside contributors, who will always have to come from a fork.

mikemhenry:
We run into the same problem with perses and other packages that need openeye, which forces us to work on branches on the main repo and not a fork. We could wire up the CI to skip tests+steps that depend on openeye so that we can run everything that doesn't require it.

IAlibay:
I'll create a copy of the branch and open a new PR once tests are finished.

dwhswenson:
> We run into the same problem with perses and other packages that need openeye, which forces us to work on branches on the main repo and not a fork. We could wire up the CI to skip tests+steps that depend on openeye so that we can run everything that doesn't require it.

Tests that require OE should already be skipping; and it's easy enough to only run the license check on things coming from the original repo. However, the issue I was thinking about is that we'll see a coverage drop on PRs from external forks, which is kind of annoying. Failing PRs because of that is not a good onboarding experience for new contributors. Even only requiring patch coverage could be a problem if a contributor makes an oechem-based contribution. Not sure how we handle that, but it doesn't need to be dealt with today.

mikemhenry:
Yes but this part of the CI:

https://github.com/OpenFreeEnergy/openfe/blob/main/.github/workflows/ci.yaml#L48-L53

runs every time, which is the step that is failing, so we need to add a bit of logic that will skip this on a PR, I will add that

mikemhenry:
This PR should fix that issue https://github.com/OpenFreeEnergy/openfe/pull/71

IAlibay:
image
somethings broken in recent commits - I'm trying to hunt down the cause

IAlibay:
image

Fixed in 96f93f9

openfe/tests/setup/test_atommapping.py
```diff
@@ -12,3 +14,18 @@ def test_atommapping_usage(simple_mapping):
 def test_atommapping_hash(simple_mapping, other_mapping):
     # these two mappings map the same molecules, but with a different mapping
     assert simple_mapping is not other_mapping
+
+
+def test_draw_mapping_cairo(tmpdir, simple_mapping):
+    with tmpdir.as_cwd():
+        simple_mapping.draw_to_file('test.png')
+        filed = pathlib.Path('test.png')
+        assert filed.exists()
```
IAlibay, line 23: Asserting the contents of the files seems like it will become problematic quickly if the backend changes in any way, so I thought it might be best to just stick with "does file exist".

richardjgowers (approved):

IAlibay:
In the interest of getting this delivered today I'm going to go ahead and merge. I've raised #73 to get us to come back to this and clean up as required (probably with a py3dmol backend).

PR #63, opened 2022-02-22
Molecule serialization/deserialization

### LINKED ISSUE #12: Serialisation of Molecule representation

dwhswenson:
Putting this up before writing tests so that it can get some feedback more quickly. Key points:

* We add two SDF tags/RDKit Mol properties: `ofe-name` and `ofe-version`.
* Method `to_sdf` returns a string; `from_sdf_string` is the inverse process. Open to other names; maybe it should be `to_sdf_string`?
* The name is preserved on serialization, and if you try to create a `Molecule` with a conflicting name, an error is raised.
* The version is loaded from an old file, and is accessible via `molecule.rdkit.GetProp("ofe-version")`.
* When writing the SDF string, the current version of OpenFE is always used.

This will be followed by a WIP PR to for network serialization before I come back to finish this.

Progress:
- [x] Serialization code
- [x] Tests
- [x] Typing where reasonable
- [x] Docstrings

Resolves #12.


dwhswenson:
As of the first commit ([81c6e02](https://github.com/OpenFreeEnergy/openfe/pull/63/commits/81c6e0240dd1d5522bd910e6cd4fcfef1e2f3a58)):

```python
In [1]: from rdkit import Chem

In [2]: import openfe.setup

In [3]: mol = Chem.MolFromSmiles("CC")

In [4]: m = openfe.setup.Molecule(mol, name="named")

In [5]: sdf = m.to_sdf()

In [6]: m2 = openfe.setup.Molecule.from_sdf_string(sdf)

In [7]: m2 == m
Out[7]: True
```

openfe/setup/molecule.py
```diff
@@ -20,6 +54,8 @@ class Molecule:
         will be used in the hash.
     """
     def __init__(self, rdkit: RDKitMol, name: str = ""):
+        name = _ensure_ofe_name(rdkit, name)
+        _ensure_ofe_version(rdkit)
         self._rdkit = rdkit
         self._hash = hashmol(self._rdkit, name=name)
```
richardjgowers, line 60: I think I'd rather have the ofe_name stuff in the `from_sdf_string` method.

openfe/setup/molecule.py
```diff
@@ -1,10 +1,44 @@
...
+    if name == "":
+        try:
+            rdkit_name = mol.GetProp("ofe-name")
+        except KeyError:
+            pass
+
+    if name != "" and rdkit_name != name:
+        raise RuntimeError(f"Internal inconsistency: 'name' has "
+                           f"value '{rdkit_name}' in the internal RDKit "
+                           f"representation; expected '{name}'")
```
richardjgowers, line 28: I'm not sure this should be an error, if I'd like to rename a Mol, how do I do this?  I think specifying name=something different is intentionally changing the name.

openfe/setup/molecule.py
```diff
@@ -42,3 +79,26 @@ def __hash__(self):
...
+            sdf.append('>  <%s>\n%s\n' % (prop, val))
+        sdf.append('$$$$\n')
+        return "\n".join(sdf)
+
+    @classmethod
+    def from_sdf_string(cls, sdf_str: str):
+        # https://sourceforge.net/p/rdkit/mailman/message/27518272/
+        supp = Chem.SDMolSupplier()
+        supp.SetData(sdf_str)
+        mols = list(supp)
```
richardjgowers, line 102: `list()` isn't wise on things (sdf files) which can sometimes be Gb.  I'd sooner just call `next()` on it then maybe check that the iterator is spent?

openfe/setup/molecule.py
```diff
@@ -42,3 +79,26 @@ def __hash__(self):
...
+        assert len(mols) == 1, f"Expected 1 molecule, found {len(mols)}"
```
richardjgowers, line 103: Pretty sure asserts aren't for production code?  Use an if and raise something

openfe/setup/molecule.py
dwhswenson: My thought is that a `Molecule` should be the same regardless of how it is created, assuming the molecules hash the same. So the following workflows should all produce a molecule that behaves the same:

* User creates RDKit molecule from our SDF output using only RDKit tools, passes that to our `Molecule`
* User loads our SDF output using our helper functions (i.e., `from_sdf_string`)
* User creates RDKit molecule from SMILES string, passes to our `Molecule` with the appropriate `name`

In order to do that, I think we want to ensure that we have the name/version correctly as part of `__init__`, don't we?

openfe/setup/molecule.py
dwhswenson: I see the point on renaming. However, I also see this as an easy source of user error, and it would be nice if we did something to prevent that. In the newest commit, I'm issuing a warning if you rename something -- this should prevent users from accidentally creating a problem that is only evident much further down the line while also allowing explicit renaming.

I don't think that renaming should be a common activity. You can get weird behavior like

```python
m1 = Molecule(rdkitmol, "name1")
m2 = Molecule(rdkitmol, "name2")
m1.rdkit.GetProp("ofe-name")
# 'name2'
```

Actually, given that, maybe renaming shouldn't be an allowed activity?

openfe/setup/molecule.py
richardjgowers: Ok yeah, I see that `ensure_ofe_name` is required for case 1 and that makes sense

openfe/setup/molecule.py
richardjgowers: I think this actually highlights a different problem, we should be copying (i.e. taking ownership) the input to Molecule, so this wouldn't happen

richardjgowers: I don't think this code is unthinkable...

```python

mol = Chem.MolFromSomewhere()
conf = mol.GetConformer(0)

conf = pose1
m1 = Molecule(mol, name='pose1')

conf = pose2
m2 = Molecule(mol, name='pose2')
```

dwhswenson: Yeah, that was sort of my reason for raising an error. Something like that might be tried, and if it can't do what a user would expect, we should do something to prevent the user from doing it.

What we really need is for the RDKit molecules we wrap to be immutable; I couldn't figure out how to do that. Copying still allows issues like this:

```python
m1 = Molecule(mol)
mol.SetProp('property-needed-by-planner')
network = network_planner([m1] + other_mols)  # m1.rdkit doesn't know the property; unexpected behavior
```

Another option is for us to be indifferent to the `ofe-*` properties except on loading and storing. In the same way that I manually write the `ofe-version`, I could manually write `ofe-name`. This would mean that OpenFE should never use the `ofe-name` or `ofe-version` properties. They would be set on saving. Probably should hint that they are private in that case: `_ofe_name`, `_ofe_version`.

dwhswenson: Actually, IMO this argues for the standard way to creating molecules (from user perspective) to be from a `.from_rdkit` classmethod. I have different expectations between:

```python
m1 = Molecule(rdkit=mol)
assert m.rdkit is mol  # I expect this to be true

m2 = Molecule.from_rdkit(mol)
# I don't have the same expectation here; even less so if we call it `m2.as_rdkit()`
```

richardjgowers: Ok add the constructor changes onto #58 and we can get that change done so this PR makes sense

dwhswenson: > Ok add the constructor changes onto https://github.com/OpenFreeEnergy/openfe/pull/58 and we can get that change done so this PR makes sense

Done in [c0829bf](https://github.com/OpenFreeEnergy/openfe/pull/58/commits/c0829bfae678a6992d3fbb69743367ab5eab18e8)

openfe/setup/molecule.py
```diff
@@ -28,8 +59,9 @@ class Molecule:
         will be used in the hash.
     """
     def __init__(self, rdkit: RDKitMol, name: str = ""):
-        # internally we store as RDKit, for now...
-        self._rdkit = Chem.Mol(rdkit)
+        name = _ensure_ofe_name(rdkit, name)
+        _ensure_ofe_version(rdkit)
+        self._rdkit = rdkit
```
richardjgowers, line 64: This has removed the copy on the input, which might be required, I'll let you put it back in if you want later

richardjgowers (approved):

PR #64, opened 2022-02-22
Network Serialization

### LINKED ISSUE #15: Serialisation of Network container

dwhswenson:
Note that this includes #63, and should only be merged after that.

Resolves #15. Posting early to get feedback while it is still very much in-progress.

One thing that is slightly weird here is that we write out the version both as a graph attribute and for each molecule. In principle, that might be useful if someone wanted to manually assemble the XML for the graph (and if, in the future, we have different code paths for loading molecules depending on the version of OpenFE).

<details><summary>Example output (cd2e8455846d4aa3c9f653caa1ef739396dbf958):</summary>

```xml
<graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
  <key id="d2" for="edge" attr.name="mapping" attr.type="string" />
  <key id="d1" for="node" attr.name="sdf" attr.type="string" />
  <key id="d0" for="graph" attr.name="ofe_version" attr.type="string" />
  <graph edgedefault="directed">
    <node id="mol0">
      <data key="d1">
     RDKit          2D

  2  1  0  0  0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.2990    0.7500    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0
M  END

&gt;  &lt;ofe-name&gt;


&gt;  &lt;ofe-version&gt;
0.post73.dev0+g81c6e02

$$$$
</data>
    </node>
    <node id="mol1">
      <data key="d1">
     RDKit          2D

  3  2  0  0  0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.2990    0.7500    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.5981   -0.0000    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0
  2  3  1  0
M  END

&gt;  &lt;ofe-name&gt;


&gt;  &lt;ofe-version&gt;
0.post73.dev0+g81c6e02

$$$$
</data>
    </node>
    <node id="mol2">
      <data key="d1">
     RDKit          2D

  2  1  0  0  0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.2990    0.7500    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0
M  END

&gt;  &lt;ofe-name&gt;


&gt;  &lt;ofe-version&gt;
0.post73.dev0+g81c6e02

$$$$
</data>
    </node>
    <edge source="mol0" target="mol2" id="0">
      <data key="d2">{"0": 0}</data>
    </edge>
    <edge source="mol1" target="mol0" id="0">
      <data key="d2">{"0": 0, "1": 1}</data>
    </edge>
    <edge source="mol1" target="mol2" id="0">
      <data key="d2">{"0": 0, "2": 1}</data>
    </edge>
    <data key="d0">0.post73.dev0+g81c6e02</data>
  </graph>
</graphml>
```

</details>

Tasks:
- [x] Draft serialization
- [x] Draft deserialization
- [x] Docstring
- [x] Tests
- [x] Typing


dwhswenson:
This is basically ready. It includes #63, so it will be easier to review after we get that one merged in. I'll ping everyone with a review request after that happens.

dwhswenson:
One thing that is a little weird in this -- within a network, the serialized molecules can actually be serialized with different versions of OpenFE. Each molecule's serialized version is kept independently. This means wasting a few bytes of storage, but it makes the code simpler (and means that if someone wanted to glue together a Frankenstein-style network GraphML file, it would be easier on them -- not recommended, though.)

openfe/tests/data/serialization/network_template.graphml
```diff
@@ -0,0 +1,76 @@
...
+  2  1  0  0  0  0  0  0  0  0999 V2000
+    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
+    1.2990    0.7500    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0
+M  END
+
+&gt;  &lt;ofe-name&gt;
+
+
+&gt;  &lt;ofe-version&gt;
```
richardjgowers, line 19: it looks like these tags got mangled? or is this correct?

openfe/setup/network.py
```diff
@@ -53,6 +56,90 @@ def nodes(self) -> FrozenSet[Molecule]:
...
+                        for node, sdf in graph.nodes(data='sdf')}
+
+        edges = [
+            AtomMapping(mol1=label_to_mol[node1],
+                        mol2=label_to_mol[node2],
+                        mol1_to_mol2=dict(json.loads(mapping)))
+            for node1, node2, mapping in graph.edges(data='mapping')
+        ]
+
+        return cls(edges=edges, nodes=label_to_mol.values())
```
richardjgowers, line 111: I'd probably just inline these methods until we have a use for the different entrypoint

openfe/tests/data/serialization/network_template.graphml
dwhswenson: This is correct. SDF isn't properly XML (note the lack of close-tags), but it does use XML-like `<tag>` notation. So you can't embed SDF with angle brackets as in the original usual because it isn't XML. It has to be embedded as text, and the correct way to escape angle brackets in XML is through entities.

One of many reasons I do not like XML. It's a shame that the most widely-used graph serialization formats are both XML, but here we are.

openfe/setup/network.py
dwhswenson: I see the YAGNI argument, but to be honest, I'm not 100% clear on the relative importance of various graph serialization formats. I seemed to me like GraphML and GEXF were most important, and I went with GraphML as the more software-independent solution. But I'd like to leave open the possibility of rapidly changing that if needed, because I'm definitely not an expert in this space.

Separated like this, we (or even a brave user) can immediately use any of the [other formats](https://networkx.org/documentation/stable/reference/readwrite/index.html) that NetworkX can read/write -- at least, any of the formats that preserve node/edge attributes. If you feel strongly about inlining, I can do that, but this is why I didn't inline originally.

PR #67, opened 2022-02-27
Add support for CLI `--mol` to load SDF

dwhswenson:
Now the CLI `--mol` parameter will load an SDF if the filename is given. This is the first way it tried to interpret the user input; if that fails, it interprets it as a string.

### DIFF
--- a/openfecli/parameters/mol.py
+++ b/openfecli/parameters/mol.py
@@ -42,6 +42,6 @@ get_molecule = MultiStrategyGetter(
 
 MOL = Option(
     "--mol",
-    help="Molecule. Can be provided as an SDF file or as a string",
+    help="Molecule. Can be provided as an SDF file or as a SMILES string.",
     getter=get_molecule
 )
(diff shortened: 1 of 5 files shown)


IAlibay (approved):
lgtm

openfecli/parameters/mol.py
```diff
@@ -20,8 +20,19 @@ def _load_molecule_from_smiles(user_input, context):
     return Molecule(rdkit=mol)
 
 
+def _load_molecule_from_sdf(user_input, context):
+    from openfe.setup import Molecule
+    try:
+        with open(user_input, mode="r") as sdf:
+            contents = sdf.read()
+        return Molecule.from_sdf_string(contents)
```
richardjgowers, line 28: I'm not really a fan of blindly reading an entire file.  Would it be possible to add `Molecule.from_sdf_file` where inside this creates a `SDMolSupplier` with the path

dwhswenson: Done. Take another look now that tests pass?

PR #68, opened 2022-02-28
Add support to get molecule name from _Name prop

dwhswenson:
@IAlibay asked if I could make it so that we defaulted to getting the name from the `_Name` property in an rdkit molecule, if that is set. This PR makes that happen. Now names for molecules are set with the following precedence:

1. If you specify the `name` argument when creating a molecule, that name is used. This is the way we explicitly set the name. If the name has been implicitly set by any of methods below and the explicitly-set name differs, you get a warning about renaming the molecule.
2. Otherwise, the RDKit molecule's `ofe-name` property is used (if set).
3. Otherwise, the RDKit molecule's `_Name` property is used (if set).
4. Otherwise, the name is the empty string.

### DIFF
--- a/openfe/setup/molecule.py
+++ b/openfe/setup/molecule.py
@@ -54,6 +54,14 @@ class Molecule:
     to edit the molecule do this in an appropriate toolkit **before** creating
     this class.
 
+    A molecule can have a name associated with it. The name can be
+    explicitly set by the ``name`` attribute, or implicitly set based on the
+    tags in the input molecular representation (if supported, as with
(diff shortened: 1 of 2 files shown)


richardjgowers (changes requested):
Should probably document this name precendence in the Molecule somewhere?

dwhswenson:
> Should probably document this name precendence in the Molecule somewhere?

Done. Take another look now that tests pass?

richardjgowers (approved):

PR #71, opened 2022-02-28
Skip running OE license check on PR

mikemhenry:
This should skip the OE license check when CI runs on a PR that came from a fork, but should run on branches from the main repo.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,7 @@ jobs:
             activate-environment: openfe
 
       - name: Decrypt OpenEye license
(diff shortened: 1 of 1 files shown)


IAlibay (approved):
lgtm, cheers

we'll need a way to deal with the codecov madness long term.

dwhswenson (approved):
lgtm... RE: CodeCov -- if we can trust that OE code remains a small fraction of the total (which it probably should, right?) then we can set some reasonable threshold on the codecov/project, while still requiring improved coverage on codecov/patch. This would still be a problem for externally-contributed code that uses oechem, but I suppose that should be relatively rare.

PR #76, opened 2022-03-02
add cuda shim to env yaml

IAlibay:
In the first instance let's see what happens in terms of actions channel priority vs env yaml file.

Otherwise we'll need to add it to the yaml file which might cause some discussions on if that's an issue as our outwards facing yaml file.

Edit: no good we have to define it in the environment.yaml.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,6 +40,7 @@ jobs:
         with:
             auto-update-conda: true
             use-mamba: true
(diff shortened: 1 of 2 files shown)


IAlibay:
This cuts down linux CI by ~ 1 min (from 3 to 2 mins). One out of the two macOS setups was faster by ~ 3 mins, and the other slower by 2 mins.

I'll cycle the PR to re-trigger CI and see if we can get more stats on this.

PR #78, opened 2022-03-04
use versioneer for the version number

mikemhenry:
@dwhswenson thoughts on this? The one thing that is nice about using versioneer here is the output is really useful during development, but I will admit I'm not an expert on the plugin cli architecture, so not having an `import openfe` may be very intentional.

```
$ openfe --version
openfe, version v0.1.post7+gfdc895f
```

P.S. Really not trying to bike shed hear, so if this PR feels like that is what I'm doing, feel free to close it!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 openfe/_version.py export-subst
+openfecli/_version.py export-subst
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,1 @@
-openfe/_version.py export-subst
 openfecli/_version.py export-subst
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,2 @@
 include versioneer.py
-include openfe/_version.py
+include openfecli/_version.py
--- a/openfe/__init__.py
+++ b/openfe/__init__.py
@@ -4,5 +4,5 @@ from . import simulation
 from . import analysis
 from . import utils
 
-from . import _version
+from openfecli import _version
 __version__ = _version.get_versions()['version']
--- a/openfecli/__init__.py
+++ b/openfecli/__init__.py
@@ -3,3 +3,6 @@
 
 from .plugins import OFECommandPlugin
 from . import commands
+
+from . import _version
+__version__ = _version.get_versions()['version']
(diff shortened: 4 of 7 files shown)


dwhswenson:
The issue is with `import openfe`. That has the potential to slow things like autocomplete and help. In principle, you could build a different `--version` tool that imports on the inside, but it's probably better to use the tools provided by `click`. You probably just need to bring the whole `_version.py` into this package, and then just use `openfecli.__version__`. Since the version is tied to the distribution, `openfecli` should have the same version as `openfe` (two packages, one distribution).

richardjgowers:
Could we hack versioneer to deposit the version string in two places?

dwhswenson:
> Could we hack versioneer to deposit the version string in two places?

Do we want to maintain a fork of versioneer? It should be just as easy to put the same `_version.py` in both subpackages.

dwhswenson:
or we could make `openfecli` the canonical source of the version, and then have `openfe` mirror it.

mikemhenry:
> or we could make `openfecli` the canonical source of the version, and then have `openfe` mirror it.

That should work, right? Then we have `_version.py` in one spot (easier to maintain) and just set `openfe.__version__` to be `openfecli.__version__`

dwhswenson:
yeah, it's a little counter-intuitive, since we think of the CLI as secondary to the library. But (1) they will always be installed together; (2) the CLI is worried about the library slowing its import; the library is not worried about the CLI slowing its import.

richardjgowers:
Yep sounds good then

mikemhenry:
Okay @dwhswenson does this look right?
https://github.com/OpenFreeEnergy/openfe/pull/78/files#diff-9cc6924db7b84ed9ffc983047c2f2c201b9be7aba6ff0c3f65e4298c790083fcR32

It should be okay to import openfecli, right? Or should I get the version like this:

```
from . import _version

# stuff

@click.version_option(version=_version.get_versions()['version'])
```

dwhswenson:
@mikemhenry Either version (pun intended) is fine. I think I prefer `__version__` because it mirrors `openfe.__version__` (one might use `openfecli.__version__` elsewhere in the CLI code, too).

Importing `openfecli` shouldn't be a problem; we need to do that for things like loading reusable parameters.

richardjgowers (approved):

dwhswenson (approved):

dwhswenson:
Merging; the drop is coverage is because this comes from a fork and doesn't have access to our OpenEye creds.

Issue #87, opened 2022-03-08
Molecular box serialisation


richardjgowers:

richardjgowers:
https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/chemicalsystem.py#L94

Issue #86, opened 2022-03-08
Molecular box hashing


richardjgowers:

richardjgowers:
Yeah the box is essentially an "item" in the state (see #85) as this can change the energy of the system.  Because there's only 6 values which completely define a box (for a triclinic cell which covers most/all? cases) we can just dump 6 doubles (aka 6x8 bytes) of raw bits.  This should mean we can get perfect reproduction, I think this is option 3 below.

Yeah serialisation and hashing might end up one and the same.

richardjgowers:
This is part of the `gufe.ChemicalSystem.__hash__`, went for `box.to_bytes()` to get its hash

PR #94, opened 2022-03-08
box representation

### LINKED ISSUE #85: Molecular box representation

richardjgowers:
Closes #85 #86 #87


openfe/setup/box_representation.py
```diff
@@ -0,0 +1,104 @@
...
+    yx, ly, yz = box[1]
+    zx, zy, lz = box[2]
+
+    if (xy != 0.0) or (xz != 0.0) or (yz != 0.0):
+        return False
+    # TODO: Max tilt factors, are these common across all MD engines?
+    return True
+
+
+class BoxRepresentation:
```
IAlibay, line 34: Beyond our other conversations re: maybe not having this defined in all cases, do we need to account for a `None` version of this class? (thinking of say a gas phase calculation)

richardjgowers: So we'd previously agreed that this Box information is optional, and only really necessary if the box size is important to the system (e.g. Membrane), but could be invented by the method if required (e.g. adding padding around the "interesting" parts).  Letting the method define the box ensured compatibility between different nodes.

I think `Vacuum` as a Box is good because it's saying the box isn't implicit here (as above) but instead there really is no box.  The alternative would be to make the above case `AutoBox` to signal to the method that it can invent a box, and the lack of a box meant Vacuum, but I think that's a more clunky option.

openfe/setup/box_representation.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    if box.shape != (3, 3):
+        return False
+    lx, xy, xz = box[0]
+    yx, ly, yz = box[1]
+    zx, zy, lz = box[2]
+
+    if (xy != 0.0) or (xz != 0.0) or (yz != 0.0):
+        return False
+    # TODO: Max tilt factors, are these common across all MD engines?
```
IAlibay, line 30: I think that's just a LAMMPS thing right?

openfe/setup/box_representation.py
IAlibay: Yeah `Vacuum` sounds like a good idea.

IAlibay (approved):
I don't think I have anything else to say here. `Vacuum` would be good to implement. I'll approve but would be good to hear what everyone else thinks.

openfe/setup/box_representation.py
dwhswenson: I think OpenMM strictly requires this, too.

openfe/setup/box_representation.py
```diff
@@ -0,0 +1,104 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
+import numpy as np
+
+
+def is_valid_box(box: np.ndarray) -> bool:
```
dwhswenson, line 6: Is it our job to check the validity of the box vectors? Different MD engines will fail for different kinds of boxes. You can create boxes that pass the tests here and aren't valid for OpenMM. You can create boxes that are valid for Gromacs and aren't valid for OpenMM.

openfe/setup/box_representation.py
dwhswenson: In this case, is `Vacuum` an instance, a class, or just the string?

dwhswenson (approved):
Looks mostly fine. Main question is whether we want to take on responsibility for box validity.

openfe/setup/box_representation.py
mikemhenry: That is a good question, I think we shouldn't evaluate if an engine supports something until we hand it off to the engine, that way a user can defer engine choice until they absolutely need to and then we can have each engine wrapper handle validation instead of having validation checks in each of our container classes

mikemhenry (approved):

openfe/setup/box_representation.py
richardjgowers: And gmx - https://gitlab.com/gromacs/nb-lib/-/blob/nblib/src/gromacs/pbcutil/pbc.cpp#L166

openfe/setup/box_representation.py
IAlibay: +1 to both the above

openfe/setup/box_representation.py
```diff
@@ -0,0 +1,115 @@
...
+                                 ).reshape(3, -1))
+
+    def to_bytes(self) -> bytes:
+        """A byte based representation for serialisation
+
+        This is the 3x3 matrix, expressed as float64 (doubles) in C order
+
+        This allows a lossless serialisation mechanism for the box.
+        """
+        return self.to_matrix().tobytes(order='C')
```
richardjgowers, line 92: We probably need a string based representation right?

openfe/setup/box_representation.py
```diff
@@ -0,0 +1,115 @@
...
+
+class VacuumBox:
+    def __str__(self):
+        return 'vacuum'
+
+    def __hash__(self):
+        return hash('vacuum')
+
+    def __eq__(self, other):
+        return str(other).lower() == 'vacuum'
```
richardjgowers, line 115: So this just acts like `'vacuum'`.  The alternative would be a box full of NaN.

richardjgowers:
Now in `gufe.ChemicalSystem`

Issue #92, opened 2022-03-08
Larger scale atommapping visualisation


richardjgowers:

dwhswenson:
Solved by #110.

Issue #73, opened 2022-03-01
Add extra backends and reformat visualizations


IAlibay:
As discussed with @dwhswenson the current visualization for atommappings fits our purpose but could probably be spruced up.

Potential things we should look into doing:
 - [ ] add a py3dmol backend so we can see atom mappings more easily
 - [ ] reformat the unique atoms / elements lists so they can be used by other backends
 - [ ] rename the d2d arguments as necessary

PR #100, opened 2022-03-17
Refactor visualizaton; add more visualization helpers

dwhswenson:
This is a step toward #92. Also related to #73.

This adds convenience functions to draw a single molecule with the mapping changes highlighted (`draw_one_molecule_mapping`), or to use a similar API to simply draw an unhighlighted molecule (`draw_unhighlighted_molecule`). It refactors the core integration with the RDKit `MolDraw2D` into an internal `_draw_molecules` that is used by all of these tools. Finally, it adds more thorough tests.


openfe/utils/visualization.py
```diff
@@ -83,6 +83,28 @@ def _get_unique_bonds_and_atoms(mapping: Dict[int, int],
     return uniques
 
 
+def _draw_molecules(d2d, mols, atoms_list, bonds_list, atom_colors,
```
IAlibay, line 86: do we need / want types here?

openfe/utils/visualization.py
```diff
@@ -140,22 +157,49 @@ def draw_mapping(mol1_to_mol2: Dict[int, int],
     if not d2d:
         d2d = Chem.Draw.rdMolDraw2D.MolDraw2DCairo(600, 300, 300, 300)
 
-    # Need to squash to 2D
-    mol1_copy = copy.deepcopy(mol1)
-    mol2_copy = copy.deepcopy(mol2)
-    AllChem.Compute2DCoords(mol1_copy)
-    AllChem.Compute2DCoords(mol2_copy)
+    return _draw_molecules(
```
IAlibay, line 160: So thought here - since we'll possibly use py3dmol in the future to get 3D mapping visualisations (could be fun), do we need to consider specifying the backend here (or at least opening up the idea that we could switch between backends?)

Probably should be a one molecule mapping thing though (not sure you can do side-by-side comparison in py3dmol)

openfe/utils/visualization.py
```diff
@@ -140,22 +157,49 @@ def draw_mapping(mol1_to_mol2: Dict[int, int],
...
+    uniques = _get_unique_bonds_and_atoms(mol1_to_mol2, mol1, mol2)
+    atoms_list = [uniques["atoms"] | uniques["elements"]]
+    bonds_list = [uniques["bonds"]]
+    red = (220/255, 50/255, 32/255, 1)
+    blue = (0, 90/255, 181/255, 1)
+
+    atom_colors = [{at: blue for at in uniques["elements"]}]
+
+    if d2d is None:
+        # TODO: we should try to be smarter about the default size here
```
IAlibay, line 185: I think it's probably best not to overthink this one, if folks want a specific resolution / size they should just provide their own d2d object?

openfe/utils/visualization.py
```diff
@@ -140,22 +157,49 @@ def draw_mapping(mol1_to_mol2: Dict[int, int],
...
+        d2d = Chem.Draw.rdMolDraw2D.MolDraw2DCairo(300, 300, 300, 300)
+
+    return _draw_molecules(d2d, [mol1], atoms_list, bonds_list, atom_colors,
+                           red)
+
+
+def draw_unhighlighted_molecule(mol, d2d=None):
```
IAlibay, line 192: Do we want to add this as a method accessed by Molecule? (otherwise I'm not super sure what the intended use case is here)

IAlibay (approved):
couple of comments but nothing blocking

openfe/utils/visualization.py
dwhswenson: I was thinking something like that might be a change to the `_draw_molecules`; maybe rename `d2d` as `renderer` and then `_draw_molecules` can select between `_render_drawmol2d` and `_render_py3dmol` depending on what it receives. My main goal was to move most of the drawing stuff into a separate method so we separate the stuff that is fundamentally our responsibility (decided which atoms/bonds to highlight/color) from the stuff that is just using rdkit's drawing tools to render.

I think this argues for moving the default d2d creation into the `_draw_molecules`. I've done that in [ed3fb45](https://github.com/OpenFreeEnergy/openfe/pull/100/commits/ed3fb45fb112bb4ad29d134f482f311b283f1db8)

openfe/utils/visualization.py
dwhswenson: Agreed. I think some of this came from my matplotlib-specific scratch work. I was thinking about the fact that a lot of stuff in matplotlib is given in inches, whereas rdkit is taking things in pixels -- it could be nice to use the existing matplotlib figure to do something smart here.

On reflection, I think that goes beyond the scope of these methods -- they don't need to know the existence of matplotlib. Comment removed (and relevant code moved to `_draw_molecules` as discussed in another comment).

openfe/utils/visualization.py
dwhswenson: That would be an option, though I'm not going to implement it now. We could use that to get rdkit-like image reprs in notebooks.

The actual idea was that I originally wanted to draw the network with unhighlighted molecules as the nodes, and then when you select an edge, the node visualization changed to the highlighted form. In that case, there's obviously a need for this method. I'm not using it in the final version of my network visualizer, but I figured it could be useful in other cases.

openfe/utils/visualization.py
dwhswenson: types here are pretty ugly... which is, I guess, why I should add them. Done now.

openfe/utils/visualization.py
dwhswenson: We could, or we could just wait until tomorrow when I'm asked to show that image either when you click on the node or when you hover over it (I'm not sure which one they'll ask for).

PR #101, opened 2022-03-17
proposed change to how lambda protocols work

richardjgowers:
It's much easier to pickle/serialise classes, rather than a dict of arbitrary functions.  Therefore I'd rather we pass in different classes (e.g. there'll also be `NamdProtocol` etc) rather than one class that has been filled with the appropriate functions.

Yes this isn't a final solution to this, as serialising a class is still tricky if it's defined in `__main__` etc... but I think it's the start of the correct approach.

@IAlibay is this an exhaustive list of things which scale with lambda? And/or is there even an exhaustive list across all engines?  I'm guessing this will be another question with the answer "it's complicated...."

### DIFF
--- a/openfe/setup/_rbfe_utils/lambdaprotocol.py
+++ b/openfe/setup/_rbfe_utils/lambdaprotocol.py
@@ -8,6 +8,73 @@ import copy
 from openmmtools.alchemy import AlchemicalState
 
 
+class BaseLambdaProtocol:
+    """Base class for defining a Lambda protocol
+
+    """
+    @staticmethod
+    def lambda_sterics_core(value: float) -> float:
+        return value
+
+    @staticmethod
+    def lambda_electrostatics_core(value: float) -> float:
+        return value
+
+    @staticmethod
+    def lambda_serics_insert(value: float) -> float:
+        return value
+
+    @staticmethod
+    def lambda_sterics_delete(value: float) -> float:
+        return value
+
+    @staticmethod
+    def lambda_electrostatic_insert(value: float) -> float:
+        return value
+
+    @staticmethod
+    def lambda_electrostatics_delete(value: float) -> float:
+        return value
+
+    @staticmethod
+    def lambda_bonds(value: float) -> float:
+        return value
+
+    @staticmethod
+    def lambda_angles(value: float) -> float:
+        return value
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


dwhswenson:
> And/or is there even an exhaustive list across all engines?

It seems to me that this has to be a "no" -- this requires a specific representation of the physical forces. Other representations are possible. I'd immediately point to various ways of adding polarizability. And who knows how you might interpolate between different ML-based potential energy surfaces.

I don't think this fits well with a superclass/inheritance model, either. You have many different functions which might be combined in the different ways. You need to create a new subclass for each combination. It's a lot better to register external functions (or classes that break it up smaller: You could have each *type* of protocol be a separate class with subclasses -- for example, are sterics insert/delete usually paired? I would guess so).

IAlibay:
Also this is the way that Perses does it, but it's also very different from all the other engines where folks are asked to give a sequential list of floats per lambda control variables of interest.

IAlibay:
> is this an exhaustive list of things which scale with lambda? And/or is there even an exhaustive list across all engines? I'm guessing this will be another question with the answer "it's complicated...."

Just to add here, if it wasn't clear the answer is definitely no.

For example the control parameters in gromacs are: coulombics, vdw, bonded, restraints, mass, temperature but things are handled differently than in OpenMM / Perses. For example gromacs just knows "scale charges from A to B", it doesn't know that some charges should be added and others removed.

What we probably should define is our own API on how the lambda scheduling should work (do we want to define both the addition and removal of things? do we want to only allow functions that define behaviour over 0 -> 1 or do we want to allow folks to pass a list of values too (i.e. gmx style)?, should we have stages or do we want to define the single 0 -> 1 scale and then break it up into bits if we're using engines that don't handle addition / removal specifically at different parts of the schedule?, etc...).

mikemhenry:
Is there really a way to do this in an engine agnostic way? This may be a bit of the codebase where we need to know what engine the user is going to use in order to know what API they can use.

IAlibay:
> Is there really a way to do this in an engine agnostic way? This may be a bit of the codebase where we need to know what engine the user is going to use in order to know what API they can use.

I think we can encode the base functions in a mostly MD engine agnostic way (at least for the basic things like "scale the charges" + "scale the vdw" + "scale the restraints"). Other things will be MD specific.

PR #102, opened 2022-03-17
Rename typing.py -> custom_typing.py

dwhswenson:
This solves the annoyance that various tools in my toolchain go nuts when a file has the same name as a standard library module (in this case, they were importing our `typing` instead of the stdlib).

It turns out that this also requires doing something to register the full name with `sys.modules`. I don't particularly like messing around with `sys.modules`, but I think this is only a problem in edge cases that border on malicious.


IAlibay (approved):

richardjgowers (changes requested):
Just remove the 'as' so we don't have to toy with sys. KISS

dwhswenson:
@richardjgowers : You'd rather that I change the imports everywhere to `from openfe.utils.custom_typing`? I figured it was nice to have the simpler import name, but I can make that change.

IAlibay:
> @richardjgowers : You'd rather that I change the imports everywhere to `from openfe.utils.custom_typing`? I figured it was nice to have the simpler import name, but I can make that change.

+1 on Richard's suggestion, at the end of the day it's likely only going to get used by us, so if we can keep it simple we might as well?

richardjgowers (approved):

PR #103, opened 2022-03-21
first draft of proposed FEMethod API

richardjgowers:


openfe/setup/rbfe_methods.py
```diff
@@ -0,0 +1,95 @@
...
+"""Contains RBFE methods for
+
+
+"""
+from typing import Dict
+
+from openfe.setup import AtomMapping, Molecule
+
+
+class LigandLigandTransform:
```
richardjgowers, line 12: add parent class

openfe/setup/rbfe_methods.py
```diff
@@ -0,0 +1,95 @@
...
+        self._ligand1 = ligand1
+        self._ligand2 = ligand2
+        self._mapping = ligandmapping
+        self._settings = self.__class__.get_default_settings()
+        if settings is not None:
+            self._settings.update(settings)
+        # TODO: Prepare the workload
+
+    @staticmethod
+    def get_default_settings() -> Dict:
```
richardjgowers, line 45: this is a `Settings` object

openfe/setup/rbfe_methods.py
```diff
@@ -0,0 +1,95 @@
...
+
+from openfe.setup import AtomMapping, Molecule
+
+
+class LigandLigandTransform:
+    """Calculates the free energy of an alchemical ligand swap in solvent
+
+    """
+    def __init__(self,
+                 ligand1: Molecule,
```
richardjgowers, line 17: Make it consume Grand Unified view

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,189 @@
...
+
+    forcefield_settings: ForcefieldSettings
+    state_point: StatePoint
+
+    lambda_protocol: LambdaProtocol
+
+    # solvent model?
+    solvent_padding = 1.2 * unit.nanometer
+
+    barostat: Union[OtherBarostat, OpenMM_MonteCarloBarostat]
```
richardjgowers, line 115: So something annoying and screwy about pydantic I found was that it matches union types in an odd way.  If two classes would have the same fields it will use the first in the Union (and ignore the Basemodel you passed in) see: https://stackoverflow.com/questions/69699729/pydantic-is-confused-with-my-types-and-their-union

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,189 @@
...
+    windows = 11
+    sample_endstates = False
+
+
+class OtherBarostat(BaseModel):
+    class Config:
+        extra = 'forbid'
+
+    freq = 10
+    thing = 20
```
richardjgowers, line 55: A dummy barostat to try and illustrate how this could be variable

richardjgowers:
@dwhswenson @IAlibay @mikemhenry so I had a first attempt at putting all the variables from the notebook into pydantic models.  I'd appreciate a second (or third) set of eyes to check that I pulled out anything that probably should be user tweakable, obviously there's a spectrum of tweakability we could allow though.  Also checking that the groupings are fairly logical would be helpful.

I'm not madly in love with how it, but I think it might be the best of many messy solutions (i.e. the models are fiddly and strange but the validation we'll be able to do is very powerful so probably makes it worth it).

dwhswenson:
Why we have our own objects to wrap each object in OpenMM? I'm extremely confused by everything in `rbfe_methods.py`.

openfe/setup/methods/rbfe_methods.py
mikemhenry: That is annoying!
>As such, it is recommended that, when defining Union annotations, the most specific type is included first and followed by less specific types.

Does that mean we should have `OpenMM_MonteCarloBarostat` first? or is `OtherBarostat` the most specific type?

mikemhenry:
+1 to 
> the validation we'll be able to do is very powerful so probably makes it worth it
I think this will also make it easy to read these in from yaml or some other config as well/help us with serialization of options.

environment.yml
```diff
@@ -16,6 +16,8 @@ dependencies:
   - click
   - openeye-toolkits
   - typing-extensions
+  - pydantic
+  - openff-units
   - pip:
     - git+https://github.com/OpenFreeEnergy/Lomap@main
```
mikemhenry, line 22: Not in scope of PR but, lomap is on conda-forge now so we can move this, unless we need to use the @main version, and if we do, we should get that onto conda-forge.

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # TODO: Validate that the units are correct (i.e. a pressure unit for pressure)
+    @validator('pressure', 'temperature')
+    def must_be_positive(cls, v):
+        if v <= 0:
+            raise ValueError("Must be positive")
+        return v
+
+
+# maybe this shouldn't be a pydantic class and instead our lambda protocol object
```
mikemhenry, line 43: I kind of like the idea of using a pydantic class to set everything in the `__init__` of our lambda protocol object, so I think we should keep it, since we can use pydantic to help validate what lamba protocols we allow (but that might become hairy/duplicating our lambda protocol object code)

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,189 @@
...
+
+    freq = 10
+    thing = 20
+
+
+class OpenMM_MonteCarloBarostat(BaseModel):
+    class Config:
+        extra = 'forbid'
+
+    frequency = 50  # TODO: Use timestep unit?
```
mikemhenry, line 62: Having a time step unit will help make it clear that frequency isn't supposed to be a unit of time (like ps or fs)

IAlibay: Not sure I understand what a "timestep unit" means, just calling it "steps"?

environment.yml
IAlibay: So I sort of agree here, but it also wouldn't hurt to use the "develop version" of lomap for now. That way if we pick up issues in lomap we can quickly fix, check it works and release a new lomap version rather than having to wait for conda-forge to update and then realising that it's not fixed etc...

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # mapping of component name to forcefield path
+    forcefield: Dict[str, str]
+    nonbonded_electostatics = 'PME'
+    nonbonded_cutoff = 0.9 * unit.nanometer
+    constraints = 'HBonds'
+    rigid_water = True
+
+
+class StatePoint(BaseModel):
```
IAlibay, line 50: I get the point of this, but maybe keeping it simple for users / developers would be better? Could we not sure have the one "config" class that just takes in everything you'd define for a simulation (the mdp to your FEP if you wish).

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,257 @@
...
+      number of lambda windows to calculate
+    sample_endstate : bool, default False
+      whether to sample the endstates ???
+    """
+    functions = 'default'
+    windows = 11
+    sample_endstates = False
+
+
+class MonteCarloBarostatSettings(BaseModel):
```
IAlibay, line 95: Something to think about - what if we want to use MonteCarloAnisotropicBarostat instead?

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,257 @@
...
+    output_filename = 'rbfe.nc'
+    checkpoint_interval = 10
+
+
+class SimulationLengthSettings(BaseModel):
+    class Config:
+        arbitrary_types_allowed = True
+
+    minimization = 1000
+    equilibration = 5 * unit.picosecond
```
IAlibay, line 161: for now yeah but long term we should have longer defaults 🤣

IAlibay (changes requested):
There's obviously quite a lot to do here, but I don't hate the way it currently looks as a first pass (especially given when we need to next deliver).

I do wonder if we should be reducing some of the complexity here by going with fewer separations for config parameters.

openfe/setup/methods/rbfe_methods.py
richardjgowers: I think I can take away this default to make people think about it tbh

dwhswenson (changes requested):
Back on my "I don't understand why we have these objects" comment:

1. Barostat and integrator: I don't think we need these. We can directly take in the OpenMM objects, and use their built-in serialization (by customizing pydantic serialization). Wrapping them up ourselves means that we need to maintain wrappers for every integrator anyone wants to use, which will be a headache.
2. Defaults: Please use OpenMM defaults for these values. @IAlibay may have good reasons for using, e.g., 0.9 nm as nonbonded cutoff, but our defaults should be the same as OpenMM (1.0 nm).

Overall, I think the picture here is a nouns/verbs thing. OpenMM/Perses has what it considers nouns (barostat, integrator, etc) and what it considers verbs (`createSystem` etc). We need to nounify the verbs. So we really just need objects that take the arguments of the function calls, and that's it.

As @IAlibay commented, this will technically be versioned with OpenMM. (Although I don't think `createSystem` changes that frequently, except maybe to occasionally add an obscure kwarg.)

openfe/setup/methods/rbfe_methods.py
richardjgowers: It's to try and avoid confusing `ps` and `steps`.  The nice thing about units & pydantic validation is it will raise an error if you try and give a `500 * unit.timestep` to something expecting a `ps` unit variable

openfe/setup/methods/rbfe_methods.py
richardjgowers: I think this is the idea, these `XSettings` objects are mostly ready to be unpacked into the `__init__` of the respective objects.  The pydantic layer gives us a validation and serialization mechanism.

openfe/setup/methods/rbfe_methods.py
richardjgowers: @IAlibay see here wrt multiple barostats (or any `Union` like behaviour)

@mikemhenry I'm not sure we can guarantee that all Barostats (or more generally all classes of a particular modular component) will have a distinct signature, so that we can fix this problem with match precedence

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    class Config:
+        arbitrary_types_allowed = True
+
+    # mapping of component name to forcefield path
+    forcefield: Dict[str, str]
+    nonbonded_electostatics = 'PME'
+    nonbonded_cutoff = 0.9 * unit.nanometer
+    constraints = 'HBonds'
+    rigid_water = True
```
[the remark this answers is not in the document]
richardjgowers: Will add hydrogenMass.

One thing these `Settings` classes also allow/force is a degree of separation between the user settings and the toolkit.  This will occasionally be frustrating (e.g. when I want to change `hydrogenMass` and we forgot to expose it) but it also means that the `Protocol` object can also smooth out any API breaks that happen (i.e. put in two paths around an API break).

openfe/setup/methods/rbfe_methods.py
richardjgowers: So you'd flatten the entire collection of settings into one dict?

openfe/setup/methods/rbfe_methods.py
IAlibay: Yeah sorry I meant more that I couldn't fully parse what Mike was proposing - the `unit.timestep` was my interpretation (which I agree with), I just wasn't sure if that was what was being suggested.

openfe/setup/methods/rbfe_methods.py
IAlibay: If folks are meant to interact with it, yeah. Separation is great but it will become overwhelming quickly. Maybe we can cut it into two things - basic thermodynamic settings and alchemical settings? That way you have a clean separation and it's simple for folks to separate things out in their minds?

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,257 @@
...
+class HybridTopologyFactorySettings(BaseModel):
+    use_dispersion_correction = False
+    softcore_alpha = 0.5
+    softcore_LJ_v2 = True
+    softcore_LJ_v2_alpha = 0.85
+    softcore_electrostatics = True
+    softcore_electrostatics_alpha = 0.3
+    softcore_sigma_Q = 1.0
+    interpolate_old_and_new_14s = False
+    flatten_torsions = False
```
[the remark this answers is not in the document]
IAlibay: I would be tempted to say here "leave it for now with an expected refactor soon enough". The likelihood is the way this is all handled is going to change drastically once we come up with a reasonable way to put all this into openmmtools. I know it's be nice to fix now, and if we have the time we should, but whatever we do now is based on a class that is going to die soon no matter what.

openfe/setup/methods/rbfe_methods.py
dwhswenson: I disagree with the complete flattening because it makes it impossible for users to plug in different combinations of methods, especially those that may be developed outside of OpenFE.

openfe/setup/methods/rbfe_methods.py
dwhswenson: Can we make the default protocol a single instance rather than setting defaults on parameters that users should probably think about?

openfe/setup/methods/rbfe_methods.py
IAlibay: Some degree of modularity doesn't hurt I guess, but I really wouldn't offer "statepoint" and things like "PME, cutoff, barostat" as separate settings, that makes no sense scientifically.

Let's think about this way, with the Langevin integrator, the temperature is a control variable of the integrator, which in turn is affected by the rate of collision. It becomes convoluted really quickly.

IAlibay: If we really want to keep things separate (and we ignore some issues like integrator == thermostat), then split it in a more logical manner imho:

Integrator settings + Reporter settings + Thermostat settings + Barostat settings + Alchemical settings

Essentially unless there's a real reason for it I'd just axe StatePoint.

dwhswenson: > Some degree of modularity doesn't hurt I guess, but I really wouldn't offer "statepoint" and things like "PME, cutoff, barostat" as separate settings, that makes no sense scientifically.

Agreed. I think we should mimic the original API as closely as possible. I'd do a set of setting for the `openmm.app.ForceField` object and its `createSystem` method. I'd just use the OpenMM integrator and barostat directly (they can be serialized to XML). Another set of settings for the `HybridTopology`. Another set for the `openmm.app.Simulation` object (need a place to specify platform settings). Another set to describe the solvation process (basically wrapping `addSolvent` and possible some associated things). Finally, some global settings that collects all these individual settings. Maybe I'm missing another touch point specific to the alchemical setup?

dwhswenson: > Essentially unless there's a real reason for it I'd just axe StatePoint.

+1. I think it's an attempt to avoid having to repeat the temperature setting in the barostat. I think we can just use the barostat directly.

openfe/setup/methods/rbfe_methods.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+class LambdaProtocolSettings(BaseModel):
+    """Settings for the lambda protocol
+
+    This describes a fixed number of windows
+
+    Attributes
+    ----------
+    functions : str, default 'default'
```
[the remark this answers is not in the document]
IAlibay: 🤔 maybe a point for future expansion, but maybe we could just offer an optional entry that's just a list with the definition of the windows?

openfe/setup/methods/rbfe_methods.py
IAlibay: So the question really is here, how much should we be buying into the OpenMM ecosystem?

I would suggest keeping things relatively abstracted from the eventual objects that use these settings if we want to use FEMethod for multiple MD engines eventually.

dwhswenson: I think that settings are going to be engine-specific. We can't assume we'll have the same integrators, etc. available everywhere. Because of that, I say stay as similar to the engine as possible.

IMO, all the engine settings for Gromacs should just be taken from an mdp file. Go to the users where they are; don't make a new whole new thing they need to learn to use our stuff.

openfe/setup/methods/rbfe_methods.py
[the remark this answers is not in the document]
dwhswenson: I think `softcore_LJ_v2` is that use v2 setting, no? That works (can remove duplication of `softcore_alpha`), but I think the better long-term solution is a separate object. What happens when someone comes up with v3? Better to have it be modular and replaceable.

openfe/setup/methods/rbfe_methods.py
IAlibay: I guess it depends at what point in the abstraction layer we're looking at tbh. If it is OpenMM only then we should reflect that in the method name.

If it was engine non-sepcific; every engine we use will have the same integrator type (if it's not Langevin it's not going to work as an FEP). Barostats are more or a pain but we could just fail if we don't have access to the right thing?

re: gmx I'm not sure I agree with mdp file as the entry point, we should be able to have similar entry points for every method, and making mdp files from mdp files just seems like a terrible deal for users. But anyways that's not an issue to be fixed now.

openfe/setup/methods/rbfe_methods.py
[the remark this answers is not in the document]
IAlibay: I think long term folks will be passing their own defined functions for these (at least that's what openmmtools does already, so if we go another route then that'll be up to to use to convince them otherwise).

But yeah, a "pseudo-reasonable short term solution that works for now" is the best option here imho.

openfe/setup/methods/rbfe_methods.py
dwhswenson: > I guess it depends at what point in the abstraction layer we're looking at tbh. If it is OpenMM only then we should reflect that in the method name.

I would say put them in a module to reflect that, rather than change the class names. `openfe.setup.openmm_settings`. You're rarely going to use more than one engine at once (it is possible, but not a typical use case.)

> If it was engine non-sepcific; every engine we use will have the same integrator type (if it's not Langevin it's not going to work as an FEP). Barostats are more or a pain but we could just fail if we don't have access to the right thing?

What about solvation? That's something that different engines do differently as well, and the algorithms are going to be different enough that you probably can't count on the same result from two engines. (Granted, you can solvate with Gromacs and run with OpenMM, but only if you have both installed on your system.) Note that we have to specify *which* Langevin as well, and not all are supported by all engines. The problem is that as we add engines, the number things that are "this only exists for this engine" grows. If we want to support a wide variety of feautres, the number of things that are engine-specific are going to be larger than the number of things that are shared between engines. At some point we need to maintain a database of what works with what engine (and what version of the engine). 

Aside: Why the need for Langevin? (Point to reference?) I would at least think that other stochastic integrators would work.

dwhswenson: Another point: what if an engine supports different options in the "force field" term? Example, maybe some method for electrostatics is/is not supported. If the settings are engine-specific, we can validate this on creation. If not, then it fails later.

My assumption is that most people have decided on the engine they will use before they start setting up the simulation. (In fact, they need to, to select an integrator/barostat that is supported.)

IAlibay: > Run (or reporter) settings

In an ideal world things like "Run" settings should be part of the simulation plan (or whatever Richard named it). I guess we can't current affect reporter settings too well, but you could consider a case where someone goes "o yeah I want to run a 5 ns equilibration printing every 100ps, but I was a 10 ns production printing every 2 ps".

dwhswenson: > In an ideal world things like "Run" settings should be part of the simulation plan (or whatever Richard named it).

This is an interesting point. Does the fact that it is the only thing that *isn't* backend-specific kind of suggest that it should be pushed back to a later point? Probably yes, especially if you're thinking about using some adaptive sampling, where the decision made now becomes irrelevant.

environment.yml
mikemhenry: Not a bad idea, I've forgotten to fix this twice now :sweat_smile: so lets just keep

openfe/setup/methods/rbfe_methods.py
mikemhenry: Yes sorry for the confusion!  `unit.timestep` and `unit.seconds` will be distinct so we can use pydantic to help users from confialting them

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,533 @@
...
+          a set of default settings
+        """
+        return RelativeLigandTransformSettings(
+            system_settings=SystemSettings(
+                constraints='HBonds'
+            ),
+            topology_settings=TopologySettings(
+                forcefield={'protein': 'amber99sb.xml',
+                            'ligand': 'openff-2.0.0.offxml',
+                            'solvent': 'tip3p.xml'},
```
IAlibay, line 491: Don't accept this as a change yet, I need to check it works (was getting issues with 14sb originally, hence why I dropped it from the notebook).

```suggestion
                forcefield={'protein': 'amber14-all.xml',
                            'ligand': 'openff-2.0.0.offxml',
                            'solvent': 'amber14/tip3p.xml'},
```

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,533 @@
...
+        """The uncertainty/error in the dG value"""
+        _, error = self._analyzer.get_free_energy()
+        error = (error[0, -1] * self._analyzer.kT).in_units_of(
+            unit.kilocalories_per_mol)
+
+        return error
+
+
+class RelativeLigandTransform(FEMethod):
+    """Calculates the free energy of an alchemical ligand swap in solvent
```
IAlibay, line 422:
```suggestion
    """Calculates the relative free energy of an alchemical ligand transformation.
```

Technically the way this is set it up, it could either be a complex, solvent, or gas phase transformation?

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,533 @@
...
+class RelativeLigandTransformResults:
+    """Dict-like container for the output of a RelativeLigandTransform"""
+    def __init__(self, settings: RelativeLigandTransformSettings):
+        self._parent_settings = settings
+        fn = self._parent_settings.simulation_settings.output_filename
+        self._reporter = multistate.MultiStateReporter(fn)
+        self._analyzer = multistate.MultiStateSamplerAnalyzer(self._reporter)
+
+    def dG(self):
+        """Free energy difference of this transformation
```
IAlibay, line 394: Here and below, is it worth mentioning we're returning in quantities but default kcal/mol?

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,711 @@
...
+    def from_dict(cls, d: dict):
+        """Deserialize from a dict representation"""
+        return cls(
+            stateA=ChemicalState.from_dict(d['stateA']),
+            stateB=ChemicalState.from_dict(d['stateB']),
+            mapping=LigandAtomMapping.from_dict(d['mapping']),
+            settings=dict(**d['settings']),
+        )
+
+    def run(self) -> bool:
```
IAlibay, line 526: For after Tuesday maybe - shouldn't we just be importing things from `openfe.run`?

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,711 @@
...
+            forcefield=self._settings.topology_settings.forcefield['ligand'],
+            molecules=[stateA_openff_ligand],
+        )
+        omm_forcefield_stateA = app.ForceField(
+            *[ff for (comp, ff) in self._settings.topology_settings.forcefield.items()
+              if not comp == 'ligand']
+        )
+        omm_forcefield_stateA.registerTemplateGenerator(smirnoff_stateA.generator)
+
+        stateA_modeller = app.Modeller(
```
IAlibay, line 540: Where's your protein?

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,711 @@
...
+            stateA_openff_ligand.to_topology().to_openmm(),
+            stateA_openff_ligand.conformers[0],
+        )
+
+        # Solvate the complex in a 0.15 mM cubic water box with 1.2 nm from the
```
IAlibay, line 545:
```suggestion
        # Solvate the complex in a `concentration` mM cubic water box with `solvent_padding` from the
```

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,711 @@
...
+            stateA_modeller.topology,
+            nonbondedMethod=nonbonded_method,
+            nonbondedCutoff=self._settings.system_settings.nonbonded_cutoff,
+            constraints=constraints,
+            rigidWater=self._settings.system_settings.rigid_water,
+        )
+
+        stateA_topology = stateA_modeller.getTopology()
+
+        # TODO: Algorithm this magic number
```
IAlibay, line 574: Should be just half the box length - I really need to work out _why_ this is so necessary...

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,711 @@
...
+        hybrid_factory = _rbfe_utils.relative.HybridTopologyFactory(
+            stateA_system, stateA_positions, stateA_topology,
+            stateB_system, stateB_positions, stateB_topology,
+            old_to_new_atom_map=solvent_mappings['old_to_new_atom_map'],
+            old_to_new_core_atom_map=solvent_mappings['old_to_new_core_atom_map'],
+            use_dispersion_correction=alchem_settings.use_dispersion_correction,
+            softcore_alpha=alchem_settings.softcore_alpha,
+            softcore_LJ_v2=alchem_settings.softcore_LJ_v2,
+            # TODO: Is this setting missing?
+            softcore_LJ_v2_alpha=0.85,
```
IAlibay, line 630:
```suggestion
            softcore_LJ_v2_alpha=alchem_settings.softcore_alpha,
```

Makes not sense to have both a normal and a v2 softcore alpha if you're only going to use one. So I just squashed them into one. If folks want to use the old version they'll have to manually change the default.

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,711 @@
...
+            reporter=reporter,
+            platform=platform,
+            lambda_protocol=lambdas,
+            temperature=self._settings.integrator_settings.temperature,
+            endstates=alchem_settings.unsampled_endstates,
+        )
+        sampler.energy_context_cache = energy_context_cache
+        sampler.sampler_context_cache = sampler_context_cache
+
+        return True
```
IAlibay, line 692: True?

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,711 @@
...
+
+    def is_complete(self) -> bool:
+        results_file = self._settings.simulation_settings.output_filename
+
+        # TODO: Can improve upon this by checking expected length of the
+        #       nc archive?
+        return os.path.exists(results_file)
```
IAlibay, line 699: This won't work long term, the file gets created on sampler creation.

IAlibay (changes requested):
Initial round of comments

openfe/setup/methods/openmm/equil_rbfe_methods.py
richardjgowers: Yeah longer term we'll need a lazy import system to prevent us from slowly grinding to a halt.  I'd rather we keep deps up top so they're visible then later apply a good fix.

I think an intermediate solution could be that this module isn't imported on the base `import openfe` but only on `from openfe.setup.methods import openmm`.

openfe/setup/methods/base.py
```diff
@@ -0,0 +1,34 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
+
+import abc
+from openff.toolkit.utils.serialization import Serializable
+
+
+class FEMethod(Serializable):
```
IAlibay, line 8: We were getting an MRO on this, Serializable already inherits from ABC so passing both here shouldn't be necessary?

richardjgowers: It's because some of the methods (e.g. `to_toml` iirc) of `Serializable` aren't defined so you can't go deep on the inheritance.  I think you need to include `abc.ABC` again to make it clear that this is still technically an abstract class

openfe/setup/methods/openmm/equil_rbfe_methods.py
richardjgowers: So this is currently detecting if run has started but not finished.. If we can infer from the settings how full the .nc should get we can check the "tail" of it.  This is minor as we're currently not thinking about this being threadsafe yet (i.e. we've only got one python process for now).

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,823 @@
...
+        #     stateA_modeller.add(
+        #         stateA_openff_ligand.to_topology().to_openmm(),
+        #         stateA_openff_ligand.conformers[0],
+        #     )
+        # else:
+        # --- add below --
+        stateA_modeller = app.Modeller(
+            stateA_openff_ligand.to_topology().to_openmm(),
+            stateA_openff_ligand.conformers[0],
+        )
```
IAlibay, line 597: Once https://github.com/OpenFreeEnergy/gufe/pull/17 is merged.

```suggestion
        if 'protein' in self._stateA.components:
            pdbfile = self._stateA.components['protein'].to_openmm_PDBFile()
            stateA_modeller = app.Modeller(pdbfile.topology,
                                           pdbfile.positions)
            stateA_modeller.add(
                stateA_openff_ligand.to_topology().to_openmm(),
                stateA_openff_ligand.conformers[0],
            )
        else:
            stateA_modeller = app.Modeller(
                stateA_openff_ligand.to_topology().to_openmm(),
                stateA_openff_ligand.conformers[0],
            )
```

openfe/setup/methods/openmm/equil_rbfe_methods.py
IAlibay: Very dirty but done.

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,823 @@
...
+
+        # setup sampler and feed it the context caches
+        sampler.setup(
+            reporter=reporter,
+            platform=platform,
+            lambda_protocol=lambdas,
+            temperature=self._settings.integrator_settings.temperature,
+            endstates=alchem_settings.unsampled_endstates,
+        )
```
IAlibay, line 784:
```suggestion
        ## Create sampler
        sampler_settings = self._settings.sampler_settings
        
        if sampler_settings.sampler_method == "repex":
            sampler = _rbfe_utils.multistate.HybridRepexSampler(
                mcmc_moves=integrator,
                hybrid_factory=hybrid_factory,
                online_analysis_interval=sampler_settings.online_analysis_interval,
                online_analysis_target_error=sampler_settings.online_analysis_target_error,
                online_analysis_minimum_iterations=sampler_settings.online_analysis_minimum_iterations
            )

            # TODO - update to more verbosely pass in n_replicas and n_states, which will avoid
            # duplication in the SAMS code path
            sampler.setup(
                reporter=reporter,
                platform=platform,
                lambda_protocol=lambdas,
                temperature=self._settings.integrator_settings.temperature,
                endstates=alchem_settings.unsampled_endstates,
            )
        elif sampler_settings.sampler_method == "sams":
            # TODO - add SAMS sampler - see PR #125
            raise AttributeError(f"SAMS sampler is not available yet")
        else:
            raise AttributeError(f"Unknown sampler {sampler_settings.sampler_method}")
```

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,823 @@
...
+        # Get the sub selection of the system to print coords for
+        selection_indices = hybrid_factory.hybrid_topology.select(
+                self._settings.simulation_settings.output_indices
+        )
+
+        # Create the multistate reporter
+        reporter = multistate.MultiStateReporter(
+            self._settings.simulation_settings.output_filename,
+            analysis_particle_indices=selection_indices,
+            checkpoint_interval=self._settings.simulation_settings.checkpoint_interval.m,
```
IAlibay, line 738: typo + additional option
```suggestion
            checkpoint_interval=self._settings.simulation_settings.checkpoint_interval,
            checkpoint_storage=self._settings.simulation_settings.checkpoint_storage,
```

openfe/setup/_rbfe_utils/visualization.py
```diff
@@ -6,8 +6,14 @@
 from rdkit.Chem import AllChem
```
IAlibay, line 6: Actually just delete this entire file? We don't need it anyways...

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,823 @@
...
+        def get_center_offset(omm_system):
+            """Helper function to get the centering offset for a cubic box.
+
+            TODO
+            ----
+            * Deal with non-cubic boxes
+            """
+            # Cubic so we can safely assume the boxes are the same length
+            edge_length = omm_system.GetDefaultPeriodicBoxVectors()[0][0]
+            edge_nm = edge_length.value_in_unit(unit.nanometer)
```
IAlibay, line 648:
```suggestion
            edge_nm = edge_length.value_in_unit(unit.nanometer) / 2
```
oops sorry

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,838 @@
...
+            * Deal with non-cubic boxes
+            """
+            # Cubic so we can safely assume the boxes are the same length
+            edge_length = omm_system.getDefaultPeriodicBoxVectors()[0][0]
+            edge_nm = edge_length.value_in_unit(omm_unit.nanometer) / 2
+            return np.array([edge_nm, edge_nm, edge_nm]) * omm_unit.nanometer
+
+        # Center the positions in the middle of the box by shifting by offset
+        center_offset = get_center_offset(stateA_system)
+        #center_offset = np.array([14, 14, 14])
```
IAlibay, line 666:
```suggestion

```

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,838 @@
...
+        integrator_settings = self._settings.integrator_settings
+
+        # create langevin integrator
+        integrator = openmmtools.mcmc.LangevinSplittingDynamicsMove(
+            timestep=integrator_settings.timestep,
+            collision_rate=integrator_settings.collision_rate,
+            n_steps=integrator_settings.n_steps.m,
+            reassign_velocities=integrator_settings.reassign_velocities,
+            n_restart_attempts=integrator_settings.n_restart_attempts,
+            constraint_tolerance=integrator_settings.constraint_tolerance,
```
IAlibay, line 781:
```suggestion
            constraint_tolerance=integrator_settings.constraint_tolerance,
            splitting=integrator_settings.splitting
```

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,838 @@
...
+        )
+
+        # setup sampler and feed it the context caches
+        sampler.setup(
+            reporter=reporter,
+            platform=platform,
+            lambda_protocol=lambdas,
+            temperature=to_openmm(self._settings.integrator_settings.temperature),
+            endstates=alchem_settings.unsampled_endstates,
+        )
```
IAlibay, line 798:
```suggestion
        ## Create sampler
        sampler_settings = self._settings.sampler_settings
        
        if sampler_settings.sampler_method == "repex":
            sampler = _rbfe_utils.multistate.HybridRepexSampler(
                mcmc_moves=integrator,
                hybrid_factory=hybrid_factory,
                online_analysis_interval=sampler_settings.online_analysis_interval,
                online_analysis_target_error=sampler_settings.online_analysis_target_error.m,
                online_analysis_minimum_iterations=sampler_settings.online_analysis_minimum_iterations
            )
            # TODO - update to more verbosely pass in n_replicas and n_states, which will avoid
            # duplication in the SAMS code path
            sampler.setup(
                reporter=reporter,
                platform=platform,
                lambda_protocol=lambdas,
                temperature=to_openmm(self._settings.integrator_settings.temperature),
                endstates=alchem_settings.unsampled_endstates,
            )
        elif sampler_settings.sampler_method == "sams":
            # TODO - add SAMS sampler - see PR #125
            raise AttributeError(f"SAMS sampler is not available yet")
        else:
            raise AttributeError(f"Unknown sampler {sampler_settings.sampler_method}")
```

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,838 @@
...
+        if not dry:
+            # TODO
+            # add some logging for verbosity
+
+            # minimize
+            sampler.minimize(max_iterations=self._settings.simulation_settings.minimization_steps)
+
+            # equilibrate - possibly on a tdqm loop so you can track progress?
+
+            # production - possibly on a tqdm loop so you can track progress?
```
IAlibay, line 812:
```suggestion
            # Get simulation control settings
            sim_settings = self._settings.simulation_settings
            ts = self._settings.integrator_settings.timestep
            mc_steps = self._settings.integrator_settings.n_steps.m

            # minimize
            sampler.minimize(max_iterations=sim_settings.minimization_steps)

            # equilibrate
            equil_time = sim_settings.equilibration_length.to('femtosecond')
            equil_steps = round(equil_time / (ts * mc_steps))
            sampler.equilibrate(equil_steps.m)
            
            # production
            prod_time = sim_settings.production_length.to('femtosecond')
            prod_steps = round(prod_time / (ts * mc_steps))
            sampler.extend(prod_steps.m)
```

to-do there's an expectation here that production_time is a multiple of the integrator's n_steps. This should be checked explicitly & users warned.

IAlibay (changes requested):

IAlibay (approved):
Pretty API, messy internals 🙃

PR #112, opened 2022-03-31
Add gufe

mikemhenry:
This adds gufe from my anaconda channel while we wait for it to be on conda-forge, as a first pass I didn't try and wire it in, just wanted to see first if it would install okay


mikemhenry:
Gah I was worried the git history would be messed up, lets make sure to squash this one!!

IAlibay (approved):

PR #120, opened 2022-04-01
Reworking method API settings

IAlibay:
Not complete, but I'm adding this here so @richardjgowers can keep track of progress on this.


openfe/setup/methods/rbfe_methods.py
```diff
@@ -19,95 +19,195 @@
...
+      User defined box vectors to fill with water. Default None.
+    box_size : Vec3
+      User defined box size to fill with water. Default None.
+    """
+    model = 'tip3p'
+    padding = Union[float, None]
+    ionic_strength = 0 * unit.molar
+    num_added = Union[int, None]
+    positive_ion = 'Na+'
+    negative_ion = 'Cl-'
```
richardjgowers, line 86: Strength and identity of the ions are covered by `SolventComponent`

openfe/setup/methods/rbfe_methods.py
```diff
@@ -19,95 +19,195 @@
...
+    neutralize = True
+    box_vectors = Union[tuple(Vec3), None]
+    box_size = Union[Vec3, None]
```
richardjgowers, line 89: Technically not covered by the `ChemicalSystem` but must be compatible...

richardjgowers (commented):
looking good

openfe/setup/methods/rbfe_methods.py
dwhswenson: I think we still need something more specific to this solvation method's parameters. I was working on this:

```python
class OpenMMSolvation:
    model: str = 'tip3p'
    boxSize: Tuple[float, float, float]
    padding: float = 1.2
    positiveIon: str = "Na+"
    negativeIon: str= "Cl-"
    ionicStrength: float = 0.0
    
    @classmethod
    def from_chemical_system(cls, chemical_system, model, solvent_component='solvent'):
        # should model default to tip3p here?
        gufe_solvent = chemical_system[solvent_component]
        
        
        pass
    
    def add_solvent(self, unsolvated_topology, unsolvated_positions):
        modeller = openmm.app.Modeller(unsolvated_topology, unsolvated_positions)
        modeller.addSolvent(
            model=self.model,
            boxsize=self.boxsize,
            padding=self.padding,
            positiveIon=self.positiveIon,
            negativeIon=self.negativeIon,
            ionicStrength=self.ionicStrength
        )


```

dwhswenson: Basically, we need an adapter between gufe solvent and the OpenMM `Modeller.addSolvent` method

IAlibay: Ok that clarifies things - everything here but the model should be defined in the SolventComponent (even if it currently isn't).

In that case let's move model to force field (technically it's already defined but for now let's keep it different). and get rid of this class completely?

openfe/setup/methods/rbfe_methods.py
```diff
@@ -19,95 +19,195 @@
...
+    compute_platform : str
+      Which compute platform to perform the simulation on. If 'fastest', the
+      fastest compute platform available will be chosen. Default 'fastest'.
+     
+    TODO
+    ----
+    * In the future make precision and deterministic forces user defined too.
     """
+    compute_platform = 'fastest'
+
```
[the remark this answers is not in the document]
IAlibay: This is why this needs to be user defined imho. The best we can do is say we'll pick up the fastest on the current platform or someone / something will need to define the platform based on expected resources?

openfe/setup/methods/rbfe_methods.py
```diff
@@ -19,95 +19,195 @@
...
-    forcefield: Dict[str, Union[List[str, ...], str]]
-    nonbonded_electostatics = 'PME'
-    nonbonded_cutoff = 0.9 * unit.nanometer
-    constraints = 'HBonds'
+    nonbonded_method = 'PME'
+    nonbonded_cutoff = 1.0 * unit.nanometer
+    constraints = Union[str, None] # Usually use HBonds
     rigid_water = True
+    remove_com = True # Probably want False here
+    hydrogen_mass = Union[float, None]
```
dwhswenson, line 49: some more free code I've already written; thinking along similar lines:

```python
    def create_forcefield(self, chemical_system, standard_components=['protein', 'solvent']):
        # standard_components go into directly into the openmm.app.ForceField, everything else
        # goes through SMIRNOFF
        standard_forcefields = sum([self.forcefields[comp] for comp in standard_components], [])
        forcefield = openmm.app.ForceField(*standard_forcefields)
        custom_components = {k: v for k, v in chemical_system.items() if k not in standard_components}
        by_forcefield = defaultdict(list)
        for comp, ff in custom_components.items():
            by_forcefield[ff].append(comp)
        
        for ff, comps in by_forcefield.items():
            generator = SMIRNOFFTemplateGenerator(
                forcefield=ff,
                molecules=[chemical_system[comp].to_openff() for comp in comps]
            )
            forcefield.registerTemplateGenerator(generator)
        return forcefield
```

openfe/setup/methods/rbfe_methods.py
IAlibay: Sorry didn't see @dwhswenson's message. Need to rethink that one, but can we say that whatever we need to bridge things needs to exist "somewhere else"?

richardjgowers: The identity & strength of ions is covered in `Component`.  If you want to tweak that to tighten up the checks, go over to `gufe`

dwhswenson: I think this is the place for it. "The algorithm we use to solvate the system" is component that should be modular. The `SolventComponent` defines some characteristics of the final product, but the algorithm itself could change.

Again, what we're doing here is nounifying the verbs.

dwhswenson: @richardjgowers So instead of my classmethod `from_chemical_system`, you want that to be `__init__` and make the kwargs for `addSolvent` be properties?

dwhswenson: When do we wrap in defaults? This would require that gufe objects have defaults, wouldn't it?

IAlibay: I've moved this out for now, I'll let this discussion run alongside me doing the other settings. Can always bring things back in later.

openfe/setup/methods/rbfe_methods.py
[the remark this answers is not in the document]
IAlibay: On second thought - this needs to exist when we create the compute context anyways. So at that point we should be on the compute node.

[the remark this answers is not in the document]
dwhswenson: > Some care required here, it's probable that we'll create these objects on a different place than the compute node. A sort of cross compilation type problem

Note that OpenMM will not create a simulation with a platform that isn't available on your machine. So we may need to create on the compute nodes.

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,436 @@
...
+    # Sampling State defining things
+    integrator_settings: IntegratorSettings
+    barostat_settings: BarostatSettings
+    sampler_settings: SamplerSettings
+
+    # Simulation run settings
+    simulation_settings: SimulationSettings
+
+    # solvent model?
+    solvent_padding = 1.2 * unit.nanometer
```
IAlibay, line 368: Wasn't really keeping up with what the current plan is re: solvent - I've left this in for now.

IAlibay:
Not really sure what the plan was re: tests, but should be good to go for now @richardjgowers - we can always add our tests slowly over the weekend.

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -0,0 +1,479 @@
...
+    ----------
+    sampler_method : str
+      Sampler method to use, currently supports repex (hamiltonian replica
+      exchange) and sams (self-adjusted mixture sampling). Default repex.
+    online_analysis_interval : int
+      The interval at which to perform online analysis of the free energy.
+      At each interval the free energy is estimate and the simulation is
+      considered complete if the free energy estimate is below
+      ``online_analysis_target_error``. Default `None`.
+    online_analysis_target_error : float * unit.kt
```
IAlibay, line 166: fudge, there's no kt in `unit`.

PR #121, opened 2022-04-01
Results work

richardjgowers:


openfe/setup/methods/rbfe_methods.py
```diff
@@ -241,7 +270,11 @@ def run(self) -> bool:
         return False
 
     def is_complete(self) -> bool:
-        return False
+        results_file = self._settings.SimulationLengthSettings.output_filename
+
+        # TODO: Can improve upon this by checking expected length of the
```
IAlibay, line 279: Unfortunately not as simple because you might terminate early if you're doing online analysis.

_should_ be feasible to look at the settings and assess the outcome though.

openfe/setup/methods/rbfe_methods.py
```diff
@@ -190,7 +192,34 @@ class Config:
...
+    def dG(self):
+        """Free energy difference of this transformation
+
+        Returns
+        -------
+        dG : unit.Quantity
+          The free energy difference between the first and last states. This is
+          a Quantity defined with units.
+        """
+        dG, _ = self._analyzer.get_free_energy()
```
IAlibay, line 214: I _think_ this also works for SAMS but I might have to check later on, adding TODO for now.

IAlibay (approved):
lgtm for now, as mentioned below I _think_ it'll work for SAMS, but I'll need to run through it later.

PR #125, opened 2022-04-04
Add sams and independent samplers to equil RBFE protocol

IAlibay:
This uses the sams sampler as implemented in perses + an independent sampler which is just vanilla MultiStateSampler.

Long term we will want to split the methods out, especially for the independent lambdas sampler which we might want to parallelise better, but for now (and the sake of ACS) this will do the trick.

Status:
  - Solvent calculation using SAMS and REPEX give within error differences in dG
  - Currently testing the independent windows sampler

TODO:
  - Need to test that the creation of states when n_replicas < lambda_windows is done properly.


richardjgowers:
@IAlibay does this want going in?

IAlibay:
skip for now, it's a bug but it doesn't affect the notebook.

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -209,7 +221,8 @@ class Config:
     online_analysis_target_error = 0.2 * unit.boltzmann_constant * unit.kelvin
     online_analysis_minimum_iterations = 50
     flatness_criteria = 'logZ-flatness'
-    gamma0 = 0.0
+    gamma0 = 1.0
+    n_replicas: Optional[int] = None
```
richardjgowers, line 225: so this is replicas not repeats right?

openfe/setup/methods/openmm/equil_rbfe_methods.py
```diff
@@ -868,21 +883,36 @@ def get_center_offset(omm_system):
...
-            # TODO - update to more verbosely pass in n_replicas and n_states, which will avoid
-            # duplication in the SAMS code path
-            sampler.setup(
-                reporter=reporter,
-                platform=platform,
-                lambda_protocol=lambdas,
-                temperature=to_openmm(self._settings.integrator_settings.temperature),
-                endstates=alchem_settings.unsampled_endstates,
-            )
         elif sampler_settings.sampler_method == "sams":
```
richardjgowers, line 886: I'd be tempted to throw a `.lower()` into any string comparison for luck

openfe/setup/methods/openmm/equil_rbfe_methods.py
IAlibay: Yup as per our new semanticsTM

openfe/setup/methods/openmm/equil_rbfe_methods.py
IAlibay: Done

richardjgowers (approved):

PR #70, opened 2022-02-28
Explicit mappings

richardjgowers:

Issue #54, opened 2022-02-14
AtomMapping create/delete hydrogens


richardjgowers:
Current atom mappings are probably heavy atom only, create a method which creates a new AtomMapping with hydrogens included (and reverse route, drop hydrogens)

richardjgowers:
This should be part of the atom mapping step

PR #135, opened 2022-04-13
[WIP] Helper for hydrogen mappings

dwhswenson:
Off topic for the month, but when I was writing the example notebook in openfreeenergy/ExampleNotebooks#13, I found myself writing code that seemed pretty generic. Basically, as @richardjgowers has pointed out, working with explicit hydrogens is no fun in cheminformatics. But we seem to need explicit hydrogens.

So maybe we can write tools that facilitate the transition to adding hydrogens? This is a first stab going from a heavy-atom mapping to a mapping that includes hydrogens. Putting up the PR as a place for discussion and to get some sense of whether it's worth continuing this at some point in the future.

### DIFF
--- a/openfe/setup/utils.py
+++ b/openfe/setup/utils.py
@@ -1,0 +1,40 @@
+import networkx as nx
+from rdkit import Chem
+
+from typing import Dict
+
+def hydrogen_mapping(
+    heavy_mapping: Dict[int, int],
+    molA: Chem.Mol,
+    molB: Chem.Mol
+) -> Dict[int, int]:
+    """Create the hydrogen mappings for a given heavy atom mapping.
+
+    Parameters
+    ----------
+    heavy_mapping: Dict[int, int]
+        mapping for heavy atoms. If a heavy atom is included on either side
+        of the mapping, it should be included here (even if the atom it maps
+        to is a hydrogen).
+    molA, molB : Chem.Mol
+        RDKit representation molecules for this mapping pair
+
+    Return
+    ------
+    Dict[int, int]
+        hydrogen-to-hydrogen mappings
+    """
+    hyd_mapping = {}
+    graphA = nx.Graph([(b.GetBeginAtomIdx(), b.GetEndAtomIdx())
+                       for b in molA.GetBonds()])
+    graphB = nx.Graph([(b.GetBeginAtomIdx(), b.GetEndAtomIdx())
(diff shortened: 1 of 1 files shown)


IAlibay:
Just FYI this logic is already in lomap if I remember correctly. It might make sense to just expose that method.

dwhswenson:
I think this is the relevant code in lomap: https://github.com/OpenFreeEnergy/Lomap/blob/b2d0fd99f9f16aa307d5703be53563cdc1439a99/lomap/mcs.py#L1187-L1238

It's not clear to me how easy it would be to split that off and make it usable by an arbitrary atommapping that we provide. But I do think we should have something that does that (whether my little start here ends up being what we use or not).

richardjgowers:
This is for #54 right?  I had a draft PR (#70) which I think is very similar to this approach.  I think at the time @IAlibay wanted the hydrogens to be assigned to the best match wrt geometry.

richardjgowers (changes requested):
Not a fan of transferring everything into nx, it's heavier than e.g. a dict and we're not using it for anything more than a dict.  If we are needing a graph we should use the graph representation we already have (hello rdkit), rather than copy data structures into a different format.

openfe/setup/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    molA: Chem.Mol,
+    molB: Chem.Mol
+) -> Dict[int, int]:
+    """Create the hydrogen mappings for a given heavy atom mapping.
+
+    Parameters
+    ----------
+    heavy_mapping: Dict[int, int]
+        mapping for heavy atoms. If a heavy atom is included on either side
+        of the mapping, it should be included here (even if the atom it maps
```
IAlibay, line 17: Doesn't this mean that we then have to have a means of going from our "heavy only" mapping to a "heavy with hydrogens"? I'm not sure how you can a priori know that a heavy atom is going to map to a hydrogen otherwise.

IAlibay (commented):
From a quick glance - yeah I think we'll need to account for coordinates for this to be minimally functional. Otherwise when you're mapping CH3 to CH3 how do you know that you've not crossed your hydrogens?

dwhswenson:
This was mainly as a starting point because I figured we should have code that solves this problem in general and I wasn't aware of #54 already tracking this problem.

Don't think this approach will work, as pointed out by @IAlibay. Closing it.

PR #137, opened 2022-04-15
[WIP] Add score annotation to radial network

dwhswenson:
Putting this up immediately to show @richardjgowers how I see this working (it's a pretty small change). Will add a test this afternoon (organizing a few other things right now).

PR #139, opened 2022-04-18
Minimal spanning network

dwhswenson:
Trying to fill in the stub for `minimal_spanning_graph`.

I think this is working now (has a basic test that covers code and checks the things I could think of as reasonable).

This includes #137, and so should only be merged after that.

I'd welcome any additions to the tests.


dwhswenson:
I think this is ready for review. I still feel like it is somewhat under-tested, so I'd appreciate suggestions on that.

Question: What should happen if a given node is unreachable (no edge is created that connects it to the network)? Currently, I think that would pass silently (you get an unconnected node in the output network). Should that error instead? Raise a warning?

openfe/setup/ligand_network_planning.py
```diff
@@ -72,18 +76,37 @@ def generate_radial_network(ligands: Iterable[SmallMoleculeComponent],
...
-      any callable which takes an LigandAtomMapping and returns a float
+      any callable which takes a LigandAtomMapping and returns a float
     """
-    raise NotImplementedError
+    nodes = list(ligands)
+    mappings: List[LigandAtomMapping] = []  # for mypy 🙄
+
+    # First create a network with all the proposed mappings (scored)
+    mappings = sum([list(mapper.suggest_mappings(molA, molB))
+                    for molA, molB in itertools.combinations(nodes, 2)
```
richardjgowers, line 99: I can't remember if we ever decided if mappers are reversible/symmetric, this is obviously assuming A->B === B->A.  @IAlibay are mappings **always** symmetric?

openfe/setup/ligand_network_planning.py
```diff
@@ -72,18 +76,37 @@ def generate_radial_network(ligands: Iterable[SmallMoleculeComponent],
...
       the AtomMappers to use to propose mappings.  At least 1 required,
       but many can be given, in which case all will be tried to find the
       lowest score edges
     scorer : Scoring function
-      any callable which takes an LigandAtomMapping and returns a float
+      any callable which takes a LigandAtomMapping and returns a float
     """
-    raise NotImplementedError
+    nodes = list(ligands)
+    mappings: List[LigandAtomMapping] = []  # for mypy 🙄
```
richardjgowers, line 95: please non non-ascii in the source

openfe/setup/ligand_network_planning.py
```diff
@@ -72,18 +76,37 @@ def generate_radial_network(ligands: Iterable[SmallMoleculeComponent],
...
+
+    # First create a network with all the proposed mappings (scored)
+    mappings = sum([list(mapper.suggest_mappings(molA, molB))
```
richardjgowers, line 98: style quibble, keep this as a generator expression and evaluate it in the next step?

richardjgowers (changes requested):
LGTM, I think maybe once the lomap scorers are in we can have a "real" example to regression test onto too, and it might be worth holding off till then

openfe/setup/ligand_network_planning.py
IAlibay: As far as I'm aware they should be, otherwise there's a gap in the free energy between the forward and reverse transformation.

richardjgowers:
@dwhswenson and re: unreachable node, this should be an error for this function, but not necessarily in the Network object.  I imagine nx gives an error if it can't construct the MST and we should catch & nicely format that

IAlibay (approved):
lgtm!

@richardjgowers is that LGTM meant to be blocking?

dwhswenson:
One remaining possible blocker is to add the test for unreachable nodes (which I can do this afternoon).

openfe/setup/ligand_network_planning.py
```diff
@@ -72,18 +76,43 @@ def generate_radial_network(ligands: Iterable[SmallMoleculeComponent],
...
+        for mapper in mappers
+    )
+    mappings = [mapping.with_annotations({'score': scorer(mapping)})
+                for mapping in mapping_generator]
+    network = Network(mappings, nodes=nodes)
+
+    # Next analyze that network to create minimal spanning network. Because
+    # we carry the original (directed) LigandAtomMapping, we don't lose
+    # direction information when converting to an undirected graph.
+    min_edges = nx.minimum_spanning_edges(nx.MultiGraph(network.graph),
```
richardjgowers, line 109: does nx silently provide a non spanning network if there were no edges for a particular node?

dwhswenson: Yes. Test added in https://github.com/OpenFreeEnergy/openfe/pull/139/commits/a4a30e3c5dc7730033e17066674ef89f256b544f shows that it would happen -- we catch the error I added, not something from nx.

Maybe this is a difference between `nx.minimum_spanning_edges` vs. some of the other related nx functions?

richardjgowers (approved):

PR #141, opened 2022-05-19
Fix the number of production steps being used

IAlibay:
Couple of typos which means that equilibration length == production length

This fixes this


richardjgowers (approved):
D'oh

IAlibay:
Ok everything seems to work otherwise (for now).

PR #143, opened 2022-06-20
change Lomap atom mapper to default to threed=True

richardjgowers:


IAlibay (approved):
sounds like a good idea

PR #145, opened 2022-06-23
Mapping distance

richardjgowers:
adds function to calculate array of distances between atoms in mapping

todo: make this a np array or keep as list?


IAlibay (changes requested):
NumPy array would be great, also if we can compare some actual numbers that'll help us if there's a big breaking change upstream.

dwhswenson (approved):
Meant to mention earlier: I'm +1 on using NumPy. Other than that, full approval.

IAlibay:
(codecov is being silly so merged without that returning)

PR #146, opened 2022-06-23
added check for element changes in openmm rbfe protocol

### LINKED ISSUE #144: RBFE Protocol element change mappings checks
This should happen as a pre-flight check

richardjgowers:
Fixes #144

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,10 +19,8 @@ dependencies:
   - click
   - openeye-toolkits
   - typing-extensions
-  # temporarily disable lomap to use main instead
-  # - lomap2
+  - lomap2
   - openmmtools
   - pip:
     - git+https://github.com/OpenFreeEnergy/gufe@main
-    - git+https://github.com/OpenFreeEnergy/Lomap@main
     - plugcli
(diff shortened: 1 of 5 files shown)


IAlibay:
> Ok this breaks all the tests as we currently often do this.  Once lomap gets patched we can fix our tests here.

Yeah you've pretty much moved the problem upstream from run 🤣 Do you want to remove the element checking code in utils whilst you're here?

richardjgowers:
Done.

So I'm seeing elsewhere you take an average of masses.  If we're excluding element changes then this is not needed right? https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/_rbfe_utils/relative.py#L391

IAlibay:
We should keep that in for now, it's part of the underlying perses stuff so technically I guess they do allow for element changes but don't by default?

IAlibay (approved):
Approving the changes I didn't make - if we could go ahead and merge with temporary switch from conda installing lomap2 to pip installing from main that'd be grand. That way I can move on with the mapping tests I need to get done for this afternoon. I can raise an issue for us to revert the change once the release is made and things are pushed over to conda-forge (which usually take several hours).

Issue #11, opened 2022-01-24
Allow ingestion of Perses AtomMappings


richardjgowers:
Would be nice to be able to ingest an `AtomMapping` object from perses

```python
m = openfe.AtomMapping.from_perses(perses_atommapping)


```

IAlibay:
Unless there's any objections I'll work on this?

I'm already having a deep dive into perses for the hybrid topology stuff, so I might as well carry it on with it?

IAlibay:
Note: potentially blocked by #24 - needs investigation

Issue #10, opened 2022-01-24
Wrap Perses AtomMapper into openfe.setup.atommapping


richardjgowers:
The task of generating a mapping between two ligands is done by an [AtomMapper](https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/ligandatommapper.py#L10) which generates an [AtomMapping](https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/ligandatommapping.py#L13)

There is an abstract base class for defining how an Atom Mapper should function here:

https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/ligandatommapper.py#L10

There is an Atom Mapper in perses here we could assimilate:

https://github.com/choderalab/perses/blob/main/perses/rjmc/atom_mapping.py#L543

RiesBen:
This I think is done :)

PR #147, opened 2022-06-24
WIP: Perses atom mappers

RiesBen:
Here we adress https://github.com/OpenFreeEnergy/tasks/issues/3.
Goal is to embed the Perses atom mapping infrastructure like the lomap code.

Todo:
- [x] implement perses_mapper
- [x] implement perses_scorer 
- [x] write tests perses_mapper
- [x] write tests perses_scorer
- [x] pep8
- [x]  add test for Not implemented error test_perses_scorer
- [x] clean up tests with fixtures.
- [x] making perses import optional? need to decide on centralized func call or decorators
- [ ] show results

this is related to: #3 #11 #10


RiesBen:
@richardjgowers 
I think this part is nearly done! :) (Except the normalization for the scorer with position-use. Is that something we also want? )
Still I have two questions:
- [ ] Can I add perses as setup dependency?
- [ ] The Openeye license might be a problem on the CI-side right?

IAlibay:
same thing here, can you deal with the PEP8 issues before we get to review?

IAlibay:
>  Can I add perses as setup dependency?
Unless we've changed our approach - We have to make perses an optional dependency, otherwise we're tying in our uses to need an OpenEye license which is going to be problematic. We'll need to put try/except guards here.

Re: CI - that's fine, see: https://github.com/OpenFreeEnergy/openfe/blob/d4a882d467da7a670c2d83ee1c287af0b03894c2/.github/workflows/ci.yaml#L48-L54

richardjgowers:
@RiesBen we've already solved the oechem CI thing, so that should just work.  Yes add perses to CI.  I think longer term we'll set up a "low-fat" CI build that doesn't require OEChem/other optionals.

openfe/tests/setup/atom_mapping/test_lomap_scorers.py
```diff
@@ -7,20 +7,22 @@
...
 from numpy.testing import assert_allclose
 import openfe
-from openfe.setup import lomap_scorers
+from openfe.setup.atom_mapping import lomap_scorers
+
 import pytest
 from rdkit import Chem
 from rdkit.Chem.AllChem import Compute2DCoords
 
+atom_mapping_cls = openfe.setup.atom_mapping.LigandAtomMapping
```
IAlibay, line 16: I'm not sure why we can't just import & use LigandAtomMapping here, but maybe I'm missing something obvious?

In any case, if we need to do this, it might be preferrable to do this instead?
```python
from openfe.setup.atom_mapping import LigandAtomMapping as foo
```

openfe/tests/setup/atom_mapping/test_perses_scorers.py
```diff
@@ -0,0 +1,67 @@
...
+from numpy.testing import assert_allclose, assert_
+import openfe
+from openfe.setup.atom_mapping import perses_scorers, LigandAtomMapping
+from perses.rjmc.atom_mapping import AtomMapper, AtomMapping
+
+atom_mapping_cls = openfe.setup.atom_mapping.LigandAtomMapping
+
+
+# full back to back test again lomap
+def test_perses_regression(lomap_basic_test_files_dir,  # in a dir for lomap
```
IAlibay, line 15: I appreciate this is WIP, but this is quite a lot, is there a way to make this a lot cleaner by using parametrize & fixtures here?

openfe/tests/setup/conftest.py
```diff
@@ -9,8 +9,8 @@
 
 import gufe
 import openfe
-from openfe.setup import LigandAtomMapping, SmallMoleculeComponent
-
+from openfe.setup import SmallMoleculeComponent
+from openfe.setup.atom_mapping import LigandAtomMapping
```
IAlibay, line 14:
```suggestion


```

jchodera:
@RiesBen : Hey! Can we chat about what you're trying to accomplish with this and whether we might be able to coordinate around a single implementation (that can live anywhere---we're flexible!) that avoids code duplication and can meet the needs of both OpenFE and perses? We know we need to generalize and refactor this code anyway, so would be great to know if we can just use what you've done and delete it from perses, or vice versa.

I'm surprised to just learn of this now, despite @mikemhenry also being part of both the perses and OpenFE teams.

mikemhenry:
@jchodera this is just a wrapper for the perses atom mapper, I think there was a miscommunication about us vendoring the atom wrapper code, but this just imports the atom mapper and makes the api more like lomap's.

RiesBen:
I can just repeat @mikemhenry . ;)

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+        # self.map_stength = map_strength
+        self.use_positions = use_positions
+        self.coordinate_tolerance = coordinate_tolerance
+
+    def _mappings_generator(self, molA, molB):
+        _atom_mapper = AtomMapper(use_positions=self.use_positions, coordinate_tolerance=self.coordinate_tolerance)
+
+        # Type of mapping
+        try:
+            if (self.mapping_type == PersesMappingType.best):
```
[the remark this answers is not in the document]
RiesBen: Like this approach will be changed

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+The MCS class from Perses shamelessly wrapped and used here to match our API.
+
+"""
+
+from enum import Enum
+
+from openmm import unit
+
+from perses.rjmc.atom_mapping import AtomMapper, InvalidMappingException
```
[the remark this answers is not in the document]
RiesBen: @IAlibay ... error on import or on class construction? my gut feeling says on class construction? (was a bit difficult to get the error, but now it should work)

openfe/tests/setup/atom_mapping/test_lomap_scorers.py
RiesBen: resolved

openfe/tests/setup/test_network_planning.py
```diff
@@ -7,7 +7,7 @@
 import openfe.setup
 
 
-class BadMapper(openfe.setup.LigandAtomMapper):
+class BadMapper(openfe.setup.atom_mapping.LigandAtomMapper):
```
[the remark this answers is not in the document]
RiesBen: set back to old style.

openfe/tests/setup/test_openmm_equil_protocols.py
```diff
@@ -52,7 +52,7 @@ def toluene_complex_system(benzene_modifications, T4_protein_component):
 
 @pytest.fixture
 def benzene_to_toluene_mapping(benzene_modifications):
-    mapper = setup.LomapAtomMapper()
+    mapper = setup.atom_mapping.LomapAtomMapper()
```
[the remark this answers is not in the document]
RiesBen: as above

openfe/setup/atom_mapping/perses_scorers.py
```diff
@@ -0,0 +1,79 @@
...
+    # normalize
+    if (normalize):
+        oeyMolA = mapping.molA.to_openeye()
+        oeyMolB = mapping.molB.to_openeye()
+        if (use_positions):
+            raise Exception("Not Implemented yet")
+
+        else:
+            # Helpfer Function / reducing code amount
+            def _getAllMappableAtomsWith(oeyMolA, oeyMolB, numMaxPossibleMappingAtoms: int, criterium: callable) -> int:
```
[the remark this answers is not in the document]
RiesBen: refactored. ... pep8 Is generally mean... 79 chars is not much ...

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,82 @@
...
+            can be mapped, default: 0.25*unit.angstrom
+
+        """
+
+        self.unmap_partially_mapped_cycles = full_cycles_only
+        self.preserve_chirality = preserve_chirality
+        self.use_positions = use_positions
+        self.coordinate_tolerance = coordinate_tolerance
+
+        if (not toolkits.OPENEYE_AVAILABLE):
```
IAlibay, line 56: Can you even import perses without openeye?

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # Type of mapping
+        try:
+            _atom_mappings = _atom_mapper.get_all_mappings(
+                    old_mol=molA.to_openff(), new_mol=molB.to_openff())
+        except InvalidMappingException:
+            _atom_mappings = []
+
+        # Post processing
+        if (self.unmap_partially_mapped_cycles):
```
IAlibay, line 73: As previously mentioned, can this not be directly passed on class construction instead of doing it in post? My understanding is that this gets called by `get_all_mappings` if set on class construction?

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+                _atom_mappings = _atom_mapper.get_all_mappings(old_mol=molA.to_openff(), new_mol=molB.to_openff())
+            else:
+                raise ValueError(
+                    "Mapping type value error! Please chose one of the provided Enum options. Given: " + str(
+                        self.mapping_type))
+        except InvalidMappingException:
+            _atom_mappings = []
+
+        # Post processing
+        if (self.unmap_partially_mapped_cycles):
```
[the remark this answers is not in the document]
RiesBen: refactored naming. aehm but the mechanism was the way you described it, right? On construction a default is set and this can be changed anytime from there on .

openfe/setup/atom_mapping/perses_mapper.py
RiesBen: yes you can :)

RiesBen: so the if/raise is basically testing on the spot, if there is a license (similar how its not done in perses itself) 
@richardjgowers do you really want to fix this somewhere else? right now it looks to me to be very local in the perses functionality only.

openfe/setup/atom_mapping/perses_scorers.py
```diff
@@ -0,0 +1,79 @@
...
+    ----------
+    mapping: LigandAtomMapping
+        is an openFE Ligand Mapping, that should be mapped
+    use_positions: bool, optional
+        if the positions are used, perses takes the inverse eucledian distance of mapped atoms into account.
+        else the number of mapped atoms is used for the score.
+        default True
+    normalize: bool, optional
+        if true, the scores get normalized, such that different molecule pairs can be compared for one scorer metric, default = True
+        *Warning* does not work for use_positions right now!
```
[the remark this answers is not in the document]
RiesBen: added :)

openfe/setup/atom_mapping/perses_mapper.py
RiesBen: @richardjgowers Do you already have an idea, where I should move the license testing to?

RiesBen:
Things left:
 - [ ] add test for Not implemented error test_perses_scorer
 - [ ] clean up tests with fixtures.

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,87 @@
...
+
+"""
+
+from openmm import unit
+
+# Test if Openeye license present! on class construction.
+from openff.toolkit.utils.exceptions import LicenseError
+from openff.toolkit.utils import toolkits
+
+from perses.rjmc.atom_mapping import AtomMapper, InvalidMappingException
```
richardjgowers, line 15: we'll have to make perses optional here

openfe/setup/atom_mapping/perses_mapper.py
richardjgowers: The license testing should be happening inside of perses.  We shouldn't be policing how dependencies are handling their dependencies.

openfe/setup/atom_mapping/perses_mapper.py
richardjgowers: so a pattern to look up is how openff toolkit has a `requires` decorator that you can label functions/classes with.

RiesBen: suggestion: openfe/utils/integration_tools.py

one could probably also pack this into a decorator. would that work for you?

RiesBen: ok I will try the require dekorator.

richardjgowers:
@jchodera yeah this would be the standard API we're trying to put the perses mapper into: https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/ligandatommapper.py#L35

We should probably relocate that into `gufe` and just keep the individual wrappings we create in `openfe`

openfe/utils/integration_tools.py
```diff
@@ -0,0 +1,139 @@
+"""
```
richardjgowers, line 1: We don't need 100+ lines of code to optionally import something

openfe/utils/integration_tools.py
```diff
@@ -0,0 +1,139 @@
...
+
+"""
+ ===========================================================================
+    SOLUTION 1
+ ===========================================================================
+"""
+# Licenses
+# OpenEye ###################################################
+
+if (not toolkits.OPENEYE_AVAILABLE):
```
richardjgowers, line 27: this is probably going to make the package print a load of OEChem stuff to stderr on import, which is annoying

openfe/utils/integration_tools.py
richardjgowers: Also integration_tools doesn't mean anything, optional_imports maybe?

RiesBen: I think you could see it as tools to integrate other packages, but if you prefer optional_imports - no hard feelings. migth there be a case, where we want to use these tools for non optional stuff? Probably not.

Lines will be reduced after deciding for one solution :)

openfe/utils/integration_tools.py
RiesBen: here openff takes care of it and removes the logging for the testing.

in Solution 2 I would need to add that.

jchodera:
> @jchodera yeah this would be the standard API we're trying to put the perses mapper into: > https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/ligandatommapper.py#L35

Is the idea that this would be a simple mix-in that just provides the needed `suggest_mappings()` public API or the `_mappings_generator()` (which will never be part of the public API)? And that you in no way want to standardize any of the details that the initializer for the atom mapper should accept that would control how atom mappings are produced?

> We should probably relocate that into gufe and just keep the individual wrappings we create in openfe

That sounds like a good idea.

richardjgowers:
@jchodera yes a mapper just implements `_mappings_generator` (a generator producing `dict[int, int]`).  I don't think there's a way to generalise how these might be initialised, so this is left to the implementation.

jchodera:
Perhaps when you generalize this back into `gufe`, you can generalize it to use OpenFF `Molecule` objects instead of RDKit molecules so we're not trapped in one ecosystem?

IAlibay:
> Perhaps when you generalize this back into `gufe`, you can generalize it to use OpenFF `Molecule` objects instead of RDKit molecules so we're not trapped in one ecosystem?

@jchodera 💯 - the intent is that the interfacing object being passed to the atom mapper is the gufe SmallMoleculeComponent. As previously detailed, this would be flexible enough to allow for rdkit/openff/oemol inputs & outputs. The component is then intended to be converted to/from whatever the mapper API we interface with expects (e.g. rdkit for lomap). I realise this is not so clear from the current code on `main`, but I believe part of this PR is to synch back up with that intent.

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # Post processing
+        if (self.preserve_chirality):
+            [x.preserve_chirality() for x in _atom_mappings]
+
+        if(len(_atom_mappings) > 0):
+            mapping_dict = map(lambda x: x.old_to_new_atom_map, _atom_mappings)
+        else:
+            mapping_dict = [{}]
+        return mapping_dict
```
richardjgowers, line 85: this'll need to be `yield from mapping_dict`

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+
+"""
+
+from openmm import unit
+from openfe.utils import requires_package
+
+try:
+    from perses.rjmc.atom_mapping import AtomMapper, InvalidMappingException
+except ImportError:
+    pass    # Don't throw  error, will happen later
```
IAlibay, line 15: We'll need a test that blocks the perses import and checks that the `requires_package` works.

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+        _atom_mapper = AtomMapper(
+            use_positions=self.use_positions,
+            coordinate_tolerance=self.coordinate_tolerance,
+            allow_ring_breaking=self.allow_ring_breaking)
+
+        # Type of mapping
+        try:
+            _atom_mappings = _atom_mapper.get_all_mappings(
+                    old_mol=molA.to_openff(), new_mol=molB.to_openff())
+        except InvalidMappingException:
```
IAlibay, line 74: Is there a way we can trigger this in a test?

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+            _atom_mappings = []
+
+        # Post processing
+        if (self.preserve_chirality):
+            [x.preserve_chirality() for x in _atom_mappings]
+
+        if(len(_atom_mappings) > 0):
+            mapping_dict = map(lambda x: x.old_to_new_atom_map, _atom_mappings)
+        else:
```
IAlibay, line 83: As above

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+        This class uses the perses code to facilitate the mapping of the
+        atoms of two molecules to each other.
+
+        Parameters
+        ----------
+        allow_ring_breaking: bool, optional
+            this option checks if on only full cycles of the molecules shall
+            be mapped, default: False
+        preserve_chirality: bool, optional
+             , default: True
```
IAlibay, line 39: Docs need compling?

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+        mapping_type: PersesMappingType, optional
+            how to calculate the mapping and amount of mappings,
+            default: PersesMappingType.best
+        coordinate_tolerance: float, optional
+            tolerance on how close coordinates need to be, such they
+            can be mapped, default: 0.25*unit.angstrom
+
+        Raises
+        ------
+        LicenseError
```
IAlibay, line 51: Needs changing

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+            If Openeye License can not be found, the perses functionality can
+            not be used.
+
+        """
+        # error_if_no_perses(__name__)  # Todo: Remove depending on team
+        # error_if_no_openeye_license(__name__) # Todo: as above
```
IAlibay, line 57: Remove

openfe/setup/atom_mapping/perses_scorers.py
```diff
@@ -0,0 +1,117 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/openfe
+
+from typing import Callable
+
+from openfe.utils import requires_package
+
+try:
+    from perses.rjmc.atom_mapping import AtomMapper, AtomMapping
+except ImportError:
```
IAlibay, line 10: As per mapper we should block the import and make sure users are suitably errored out.

openfe/setup/atom_mapping/perses_scorers.py
```diff
@@ -0,0 +1,117 @@
...
+        if true, the scores get normalized, such that different molecule pairs
+        can be compared for one scorer metric, default = True
+        *Warning* does not work for use_positions right now!
+
+    Raises
+    ------
+    NotImplementedError
+        Normalization of the score using positions is not implemented right
+        now.
+    LicenseError
```
IAlibay, line 62: Remove

openfe/setup/atom_mapping/perses_scorers.py
```diff
@@ -0,0 +1,117 @@
...
+                          use_positions: bool = False,
+                          normalize: bool = True) -> float:
+    """
+        This function is accessing the default perser scorer function and
+        returns, the score as float.
+
+    Parameters
+    ----------
+    mapping: LigandAtomMapping
+        is an openFE Ligand Mapping, that should be mapped
```
IAlibay, line 46: OpenFE

openfe/setup/atom_mapping/perses_scorers.py
```diff
@@ -0,0 +1,117 @@
...
+    LicenseError
+        If Openeye License can not be found, the perses functionality can
+        not be used.
+
+    Returns
+    -------
+        float
+    """
+    # error_if_no_perses(__name__)  # Todo: Remove depending on team
+    # error_if_no_openeye_license(__name__)  # Todo: as above
```
IAlibay, line 71: Remove

openfe/tests/setup/atom_mapping/test_perses_atommapper.py
```diff
@@ -0,0 +1,58 @@
...
+
+    mapper = PersesAtomMapper()
+
+    mapping_gen = mapper.suggest_mappings(toluene, NigelTheNitrogen)
+    with pytest.raises(StopIteration):
+        i = 0
+        while True:
+            next(mapping_gen)
+            if (i > 1000):
+                raise Exception("this is an inf loop... but should not be!")
```
IAlibay, line 56: This isn't being triggered. Why not just do a for loop here?

openfe/tests/setup/atom_mapping/test_perses_scorers.py
```diff
@@ -0,0 +1,87 @@
...
+from numpy.testing import assert_allclose, assert_
+import pytest
+from lomap import mcs as lomap_mcs
+from openfe.setup import perses_scorers, LigandAtomMapping
+
+pytest.importorskip('perses')
+try:
+    from perses.rjmc.atom_mapping import AtomMapper, AtomMapping
+except ImportError:
+    pass
```
IAlibay, line 15: I can't remember 100% but I _think_ the importorskip might take precedence here? Do double check.

IAlibay (changes requested):

openfe/tests/setup/atom_mapping/test_perses_atommapper.py
richardjgowers: It's hard to get perses to not give a mapping, so I've removed this test for now

richardjgowers (approved):
ok I think I've addressed comments.  there's not a clear way to get perses to not give an atom mapping, so there's a few code paths we can't seem to reach.  I'll raise an issue about this.  @IAlibay can we get this merged?

IAlibay (approved):
Would not just doing an `I` -> `F` style transformation do the trick?

PR #148, opened 2022-06-27
Visualization align mapped mols to eachother

RiesBen:
this small change, adds alginments to the the molecule pair drawing in jnbs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -158,3 +158,6 @@ cython_debug/
 
 # vim
 *.swp
+
+#pre-commit
+.pre-commit-config.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -158,6 +158,3 @@ cython_debug/
 
 # vim
 *.swp
-
-#pre-commit
-.pre-commit-config.yaml
--- a/openfe/utils/visualization.py
+++ b/openfe/utils/visualization.py
@@ -2,7 +2,6 @@
 # For details, see https://github.com/OpenFreeEnergy/openfe
 from typing import Dict, Set, Tuple, Any, Collection, Optional
 from itertools import chain
-import copy
 
 from rdkit import Chem
 from rdkit.Chem import AllChem
@@ -119,10 +118,13 @@ def _draw_molecules(
         RGBA tuple for the default highlight color used in the mapping
         visualization
     atom_mapping: Dict[Tuple[int,int], Dict[int, int]], optional
-        used to align the molecules to each othter for clearer visualization,
+        used to align the molecules to each othter for clearer visualization.
+        The structure contains the indices of the molecules in mols as key
+        Tuple[molA, molB] and maps the atom indices via the value Dict[
+        molA_atom_idx, molB_atom_idx]
         default None
     """
-    #input standardization:
+    # input standardization:
     if atom_mapping is None:
         atom_mapping = {}
 
--- a/openfe/utils/visualization.py
+++ b/openfe/utils/visualization.py
@@ -134,15 +134,10 @@ def _draw_molecules(
 
     if atom_mapping is not None and len(atom_mapping) > 0:
         for (i, j), atomMap in atom_mapping.items():
-            try:
-                AllChem.GenerateDepictionMatching2DStructure(
-                    copies[j], copies[i])
-            except Exception as err:
-                print("doof, err: ", type(err), err, err.args)
-                AllChem.AlignMol(
-                    copies[j], copies[i], atomMap=[
-                        (k, v) for v, k in atomMap.items()]
-                )
+            AllChem.AlignMol(
+                copies[j], copies[i],
+                atomMap=[(k, v) for v, k in atomMap.items()]
+            )
 
     # standard settings for our visualization
     d2d.drawOptions().useBWAtomPalette()
--- a/openfe/utils/visualization.py
+++ b/openfe/utils/visualization.py
@@ -37,7 +37,7 @@ def _match_elements(mol1: RDKitMol, idx1: int, mol2: RDKitMol, idx2: int) -> boo
 
 
 def _get_unique_bonds_and_atoms(
-    mapping: Dict[int, int], mol1: RDKitMol, mol2: RDKitMol
+        mapping: Dict[int, int], mol1: RDKitMol, mol2: RDKitMol
 ) -> Dict:
     """
     Given an input mapping, returns new atoms, element changes, and
@@ -73,6 +73,7 @@ def _get_unique_bonds_and_atoms(
         elif not _match_elements(mol1, idx, mol2, mapping[idx]):
             uniques["elements"].add(idx)
 
+
     for bond in mol1.GetBonds():
         bond_at_idxs = [bond.GetBeginAtomIdx(), bond.GetEndAtomIdx()]
         for at in chain(uniques["atoms"], uniques["elements"]):
@@ -84,13 +85,13 @@ def _get_unique_bonds_and_atoms(
 
 
 def _draw_molecules(
-    d2d,
-    mols: Collection[RDKitMol],
-    atoms_list: Collection[Set[int]],
-    bonds_list: Collection[Set[int]],
-    atom_colors: Collection[Dict[Any, Tuple[float, float, float, float]]],
-    highlight_color: Tuple[float, float, float, float],
-    atom_mapping: Dict[Tuple[int, int], Dict[int, int]] = None,
+        d2d,
+        mols: Collection[RDKitMol],
+        atoms_list: Collection[Set[int]],
+        bonds_list: Collection[Set[int]],
+        atom_colors: Collection[Dict[Any, Tuple[float, float, float, float]]],
+        highlight_color: Tuple[float, float, float, float],
+        atom_mapping: Dict[Tuple[int, int], Dict[int, int]] = None,
 ) -> str:
     """
     Internal method to visualize a molecule, possibly with mapping info
@@ -121,7 +122,7 @@ def _draw_molecules(
     """
     if d2d is None:
         # select default layout based on number of molecules
-        grid_x, grid_y = {1: (1, 1), 2: (2, 1),}[len(mols)]
+        grid_x, grid_y = {1: (1, 1), 2: (2, 1), }[len(mols)]
         d2d = Chem.Draw.rdMolDraw2D.MolDraw2DCairo(grid_x * 300, grid_y * 300, 300, 300)
 
     # squash to 2D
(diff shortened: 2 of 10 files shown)


IAlibay:
@RiesBen can you address all the pep8speaks issues in the first instance?

IAlibay (changes requested):
The diff isn't showing properly unfortunately, @dwhswenson did you say there was a way to fix this?

(just a cursory review for now)

dwhswenson:
> The diff isn't showing properly unfortunately

Can you tell me more about what's missing? I'm seeing a diff -- is there something not included in it?

dwhswenson:
Yeah, try fixing conflicts first. But you might need to merge in main, fix conflicts, then unrevert the revert commit (because merging in main might undo the changes you wanted to keep) -- at which point you'd have to *again* fix conflicts.

Moral of the story: we should have branch protection on main. 🤣

mikemhenry:
> Moral of the story: we should have branch protection on main. rofl

Added!

openfe/utils/visualization.py
```diff
@@ -116,24 +116,29 @@ def _draw_molecules(
...
-            AllChem.GenerateDepictionMatching2DStructure(copies[j], copies[i])
-        except:
-            rms = AllChem.AlignMol(copies[j], copies[i], atomMap=[(k,v) for v,k in atomMap.items()])
+    [AllChem.Compute2DCoords(mol) for mol in copies]
+
+    if (atom_mapping is not None):
+        for (i, j), atomMap in atom_mapping.items():
+            try:
+                AllChem.GenerateDepictionMatching2DStructure(copies[j], copies[i])
+            except:
```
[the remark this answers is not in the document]
IAlibay: Sorry my point here was badly worded, my main thing is "what error actually gets thrown here?" we should be covering that error specifically if we can rather than using base exceptions.

openfe/utils/visualization.py
```diff
@@ -116,24 +116,29 @@ def _draw_molecules(
...
-        try:
-            AllChem.GenerateDepictionMatching2DStructure(copies[j], copies[i])
-        except:
-            rms = AllChem.AlignMol(copies[j], copies[i], atomMap=[(k,v) for v,k in atomMap.items()])
+    [AllChem.Compute2DCoords(mol) for mol in copies]
+
+    if (atom_mapping is not None):
+        for (i, j), atomMap in atom_mapping.items():
+            try:
+                AllChem.GenerateDepictionMatching2DStructure(copies[j], copies[i])
```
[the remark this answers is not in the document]
IAlibay: Do we have a test case where we can trigger this failure?

RiesBen:
> > Moral of the story: we should have branch protection on main. rofl
> 
> Added!

Usually not the way, how I enforce good practices on others....

openfe/utils/visualization.py
```diff
@@ -115,20 +117,27 @@ def _draw_molecules(
...
-            1: (1, 1),
-            2: (2, 1),
-        }[len(mols)]
-        d2d = Chem.Draw.rdMolDraw2D.MolDraw2DCairo(grid_x*300, grid_y*300,
-                                                   300, 300)
+        grid_x, grid_y = {1: (1, 1), 2: (2, 1), }[len(mols)]
+        d2d = Chem.Draw.rdMolDraw2D.MolDraw2DCairo(grid_x * 300, grid_y * 300, 300, 300)
 
     # squash to 2D
     copies = [copy.deepcopy(mol) for mol in mols]
```
richardjgowers, line 129: If this is copying rdkit mols, then `Chem.Mol(mol)` is probably wiser.  I'm not sure that a `copy.deepcopy` is guaranteed to get this correct (because bindings & pointers etc).

openfe/utils/visualization.py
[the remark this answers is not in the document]
RiesBen: Here the exception gets triggered:
image

openfe/utils/visualization.py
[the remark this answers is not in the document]
RiesBen: see above

openfe/utils/visualization.py
RiesBen: changed it

openfe/utils/visualization.py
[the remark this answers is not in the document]
RiesBen: hm... removed the call in the try part (and the try except part as well)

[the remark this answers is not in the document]
RiesBen: should resolve this and the next comment

IAlibay (approved):
lgtm!

IAlibay:
@dwhswenson do you want to give this a glance given you were last to touch that code & you had some views on where it could go forward in terms of being more modular?

openfe/utils/visualization.py
```diff
@@ -115,20 +118,26 @@ def _draw_molecules(
     highlight_color: Tuple[float, float, float, float]
         RGBA tuple for the default highlight color used in the mapping
         visualization
+    atom_mapping: Dict[Tuple[int,int], Dict[int, int]], optional
+        used to align the molecules to each othter for clearer visualization,
+        default None
```
dwhswenson, line 123: A little more detail would be useful in the docstring here -- namely, how, do I interpret this object? We're mapping a tuple to a dict; how do I understand what each int corresponds to?

openfe/utils/visualization.py
```diff
@@ -115,20 +118,26 @@ def _draw_molecules(
...
+        grid_x, grid_y = {1: (1, 1), 2: (2, 1), }[len(mols)]
+        d2d = Chem.Draw.rdMolDraw2D.MolDraw2DCairo(
+            grid_x * 300, grid_y * 300, 300, 300)
 
     # squash to 2D
-    copies = [copy.deepcopy(mol) for mol in mols]
-    for mol in copies:
-        AllChem.Compute2DCoords(mol)
+    copies = [Chem.Mol(mol) for mol in mols]
+    [AllChem.Compute2DCoords(mol) for mol in copies]
```
dwhswenson, line 133: [don't use a list comprehension unless you're assigning it](https://treyhunner.com/2019/03/abusing-and-overusing-list-comprehensions-in-python/#Loops_disguised_as_comprehensions)

openfe/utils/visualization.py
```diff
@@ -115,20 +118,26 @@ def _draw_molecules(
...
+
+    if atom_mapping is not None and len(atom_mapping) > 0:
```
dwhswenson, line 135: `len` check is meaningless here; looping over `items` for an empty dict will just do nothing.

Generally, I think the better approach is to normalize `atom_mapping` to always be a dict by the time you get here with an earlier statement like:

```python
if atom_mapping is None:
    atom_mapping = {}
```

I'd usually put that at right at the start of the function. Then you've normalized/standardized your inputs, and don't have to worry about special cases later. This reduces the cognitive load on a reader looking at later parts of your function.

The result would be that the entire `if` statement here, which handles special cases, instead gets handled by making so that by the time you get here, special cases aren't special. No need for the `if`.

(There's general philosophy that anything that involves deeper nesting increases cognitive load readers of your code, so avoiding nested methods is good. This idea comes from the concept of [cognitive complexity](https://docs.codeclimate.com/docs/cognitive-complexity), which I think of as an easier-to-estimate cousin to the more formally interesting [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity).)

openfe/utils/visualization.py
```diff
@@ -89,7 +91,8 @@ def _draw_molecules(
     atoms_list: Collection[Set[int]],
     bonds_list: Collection[Set[int]],
     atom_colors: Collection[Dict[Any, Tuple[float, float, float, float]]],
-    highlight_color: Tuple[float, float, float, float]
+    highlight_color: Tuple[float, float, float, float],
+    atom_mapping: Dict[Tuple[int, int], Dict[int, int]] = None,
```
dwhswenson, line 95:
```suggestion
    atom_mapping: Optional[Dict[Tuple[int, int], Dict[int, int]]] = None,
```

looks like mypy isn't complaining about this, but let's explicitly allow `None` here.

.gitignore
```diff
@@ -158,3 +158,6 @@ cython_debug/
 
 # vim
 *.swp
+
+#pre-commit
+.pre-commit-config.yaml
```
dwhswenson, line 163: Note that you can also add gitignores to your local global gitignore (not that I haven't been guilty of tossing my own local things into our project gitignores).

If you're using a a pre-commit hook to blacken code (or to use some other code formatter), this can be annoying since we don't actually require a black code style. It means that PRs from you will reformat other people's code, meaning that reviewing your PRs will involve looking over changed lines that are just style changes. If you can configure your reformatter to only reformat lines you've edited, that would make things easier.

(I see a few changes in here that look like black or YAPF decided to change things that didn't match its style preferences.)

dwhswenson (changes requested):
Good start -- a couple minor changes to help with readability/general Python coding suggestions.

At some point we might want to put the work that `AllChem.AlignMol` is doing into a separate function that might be reused elsewhere in openfe (and to separate concerns; aligning the molecules is a separate question from drawing the molecule). I don't see a need to do that now (single function call), but if we find ourselves adding handling for special cases, then we should separate the concerns.

openfe/utils/visualization.py
RiesBen: uh that is a nice way of typing that!

RiesBen:
> Good start -- a couple minor changes to help with readability/general Python coding suggestions.
> 
> At some point we might want to put the work that `AllChem.AlignMol` is doing into a separate function that might be reused elsewhere in openfe (and to separate concerns; aligning the molecules is a separate question from drawing the molecule). I don't see a need to do that now (single function call), but if we find ourselves adding handling for special cases, then we should separate the concerns.

Thanks for the cool review! :) I adapted some stuff.

dwhswenson (approved):
Looks good now!

openfe/utils/visualization.py
```diff
@@ -227,13 +245,18 @@ def draw_one_molecule_mapping(mol1_to_mol2, mol1, mol2, d2d=None):
...
 
-    return _draw_molecules(d2d, [mol1], atoms_list, bonds_list, atom_colors,
-                           red)
+    return _draw_molecules(d2d,
+                           [mol1],
+                           atoms_list,
+                           bonds_list,
+                           atom_colors,
+                           red
+                           )
```
IAlibay, line 259: I'm not going to block over something like this.,.. but this is the kind of autoformatter shenanigans that makes me hate them.

mikemhenry: If you are going to format it like this, at least add the trailing `,` after red so that if we ever add something to the list of args then the diff is smaller

PR #157, opened 2022-07-13
Add lower pin for lomap2 in environment yaml file

IAlibay:
Pretty simple single line change.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
   - click
   - openeye-toolkits
   - typing-extensions
-  - lomap2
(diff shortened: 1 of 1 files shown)


dwhswenson (approved):

mikemhenry:
Should we add the same pin to our conda-forge recipe? @IAlibay  https://github.com/conda-forge/openfe-feedstock/blob/d4183f68a2a8e47c620bdf031bf993e12fa7eb83/recipe/meta.yaml#L25
I'm not sure if we want to/need to do a release, but I can add the pin and just bump the build number.

PR #158, opened 2022-07-15
154 add test for persesmapper not finding a mapping

RiesBen:
Adding an error to the Perses mapper, if it does not find a mapping.

### DIFF
--- a/openfe/setup/atom_mapping/perses_mapper.py
+++ b/openfe/setup/atom_mapping/perses_mapper.py
@@ -20,6 +20,7 @@ from .ligandatommapper import LigandAtomMapper
 class PersesAtomMapper(LigandAtomMapper):
     allow_ring_breaking: bool
     preserve_chirality: bool
+    use_positions: bool
 
     @requires_package("perses")
     def __init__(self, allow_ring_breaking: bool = True,
@@ -42,7 +43,7 @@ class PersesAtomMapper(LigandAtomMapper):
         coordinate_tolerance: float, optional
             tolerance on how close coordinates need to be, such they
             can be mapped, default: 0.25*unit.angstrom
-        
+
         """
         self.allow_ring_breaking = allow_ring_breaking
         self.preserve_chirality = preserve_chirality
@@ -50,18 +51,20 @@ class PersesAtomMapper(LigandAtomMapper):
         self.coordinate_tolerance = coordinate_tolerance
 
     def _mappings_generator(self, molA, molB):
+        # Construct Perses Mapper
         _atom_mapper = AtomMapper(
             use_positions=self.use_positions,
             coordinate_tolerance=self.coordinate_tolerance,
             allow_ring_breaking=self.allow_ring_breaking)
 
-        # Type of mapping
+        # Try generating a mapping
         try:
             _atom_mappings = _atom_mapper.get_all_mappings(
                     old_mol=molA.to_openff(), new_mol=molB.to_openff())
         except InvalidMappingException:
             _atom_mappings = []
-        
+
+        # Catch empty mappings here
         if(_atom_mappings is None):
             _atom_mappings = []
 
@@ -69,9 +72,10 @@ class PersesAtomMapper(LigandAtomMapper):
         if (self.preserve_chirality):
             [x.preserve_chirality() for x in _atom_mappings]
 
+        # Translate mapping objects
         if(len(_atom_mappings) > 0):
             mapping_dict = map(lambda x: x.old_to_new_atom_map, _atom_mappings)
         else:
-            mapping_dict = []
+            mapping_dict = [{}]
+
         yield from mapping_dict
-   
--- a/openfe/setup/atom_mapping/perses_mapper.py
+++ b/openfe/setup/atom_mapping/perses_mapper.py
@@ -76,6 +76,6 @@ class PersesAtomMapper(LigandAtomMapper):
         if(len(_atom_mappings) > 0):
             mapping_dict = map(lambda x: x.old_to_new_atom_map, _atom_mappings)
         else:
-            mapping_dict = [{}]
+            mapping_dict = []
 
         yield from mapping_dict
--- a/openfe/setup/atom_mapping/perses_mapper.py
+++ b/openfe/setup/atom_mapping/perses_mapper.py
@@ -65,7 +65,7 @@ class PersesAtomMapper(LigandAtomMapper):
             _atom_mappings = []
 
         # Catch empty mappings here
-        if(_atom_mappings is None):
+        if (_atom_mappings is None):
             _atom_mappings = []
 
         # Post processing
--- a/openfe/setup/atom_mapping/perses_mapper.py
+++ b/openfe/setup/atom_mapping/perses_mapper.py
@@ -42,11 +42,6 @@ class PersesAtomMapper(LigandAtomMapper):
         coordinate_tolerance: float, optional
             tolerance on how close coordinates need to be, such they
             can be mapped, default: 0.25*unit.angstrom
-
-        Raises
-        ------
-        ValueError
-            This Error is triggered if no Atom mapping was found
         
         """
         self.allow_ring_breaking = allow_ring_breaking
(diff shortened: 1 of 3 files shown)


richardjgowers (changes requested):
Sorry, I thought I'd said to yield an empty list of mappings and an error isn't raised

RiesBen:
ah... ups! switched :)

RiesBen:
@richardjgowers
If a Mapping is empty the perses_scorer throws an : 
image

shall we replace it with an np.inf score?

dwhswenson:
Adding to @richardjgowers: if you want minimal spanning tree, you need to attempt all pairwise mappings. If a pair has no reasonable mapping, that's fine; you can (hopefully) still generate the MST from mappings that do work. (And our MST code has a check to ensure all nodes are connected.)

So our design is to give an empty iterator back to the user, rather than have the user manage error handling.

RiesBen:
> I think we effectively forbid the existence of empty mappings (although I'm not sure that we've stated that officially). It would be formally correct for empty mappings to _always_ have an infinite score, regardless of the scorer used. However, that's a code path that should never be reached, at least not within our workflow -- we never create empty mappings.
> 
> My vote is to just say that an empty mapping is not valid. This would be better handled in the `__init__` of `LigandAtomMapping`, instead of requiring that every single score function have a test for this special case.

hmmm... in theory there could also be a mapping, that does not directly map atoms to each other, but generates distance restraints to keep the ligands together, called linked dual-topology. (this might be something interesting for the future, as it is rather simple to implement :) )

IAlibay:
> > I think we effectively forbid the existence of empty mappings (although I'm not sure that we've stated that officially). It would be formally correct for empty mappings to _always_ have an infinite score, regardless of the scorer used. However, that's a code path that should never be reached, at least not within our workflow -- we never create empty mappings.
> > 
> > My vote is to just say that an empty mapping is not valid. This would be better handled in the `__init__` of `LigandAtomMapping`, instead of requiring that every single score function have a test for this special case.
> 
> hmmm... in theory there could be a mapping, that does not directly map atoms to each other, but generates distance restraints to keep the ligands together, called linked dual-topology. (this might be something interesting for the future, it is simple to implement and somehow an extreme case of the hybrid topology :) )
> 

I think restraints will need to be a different class. It's more of a protocol thing than a "network generation" thing too, so it has to live further downstream.

openfe/tests/setup/atom_mapping/conftest.py
```diff
@@ -41,3 +43,20 @@ def gufe_atom_mapping_matrix(lomap_basic_test_files_dir,
...
 
     return ligand_atom_mappings
+
+
+@pytest.fixture()
+def mol_pair_to_shock_perses_mapper() -> Tuple[SmallMoleculeComponent,
+                                               SmallMoleculeComponent]:
+    """
+        This pair of Molecules leads to an empty Atom mapping in
+        Perses Mapper with certain settings.
```
IAlibay, line 53:
```suggestion
    """
    This pair of Molecules leads to an empty Atom mapping in
    Perses Mapper with certain settings.
```

The indentation doesn't match the visual style.

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -50,24 +51,31 @@ def __init__(self, allow_ring_breaking: bool = True,
...
-        # Type of mapping
+        # Try generating a mapping
         try:
             _atom_mappings = _atom_mapper.get_all_mappings(
                     old_mol=molA.to_openff(), new_mol=molB.to_openff())
         except InvalidMappingException:
             _atom_mappings = []
 
+        # Catch empty mappings here
+        if(_atom_mappings is None):
```
IAlibay, line 68:
```suggestion
        if (_atom_mappings is None):
```

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -50,24 +51,31 @@ def __init__(self, allow_ring_breaking: bool = True,
...
+            _atom_mappings = []
```
IAlibay, line 69: Would it make sense to merge this if block with the ones below? seems kinda redundant to set an empty return list under 79 again.

IAlibay (changes requested):

dwhswenson:
For linked dual topology, wouldn't you still need a mapping to say which atoms between the two topologies have restraints?  If I understand the method, you essentially add a "bond" between the two topologies so that specific pairs of atoms, one from `molA` and one from `molB`, stay near each other. The mapping in `LigandAtomMapping` would define those pairs. It's entirely possible that the best mappings for alchemical transformations would be different from the best mappings for linked dual topology, but it seems like the concept of "mapping" is still present.

Restraints may also need additional information, and would certainly be consumed differently, which is why I would agree with @IAlibay that it would need to be a different class (possibly one that contains a `LigandAtomMapping`).

openfe/setup/atom_mapping/perses_mapper.py
dwhswenson: edited suggestion to remove unnecessary parens

RiesBen:
@dwhswenson good point! you would kind of build a mapping with a different mapping algorithm/metric idea ... (not necessarily considering structure similarity) 

@IAlibay 
> It's more of a protocol thing than a "network generation" thing too, so it has to live further downstream.
different class is probably a good idea, but I think you might want to build your network depending on your restraint mappings :)

openfe/setup/atom_mapping/perses_mapper.py
dwhswenson: For clarification: under what circumstances does perses return `None` instead of raising an `InvalidMappingException`?

dwhswenson: @IAlibay : it needs to be a list before it gets to `preserve_chirality`. But I'd recommend an early exit for the special case of no mappings. Easiest way to do that would be for the `except` clause to set `_atom_mappings = None` (instead of `[]`) and have this be

```python
if _atom_mappings is None:
    return
```

Then you don't need the `if len(_atom_mappings)` check.

Alternate (possibly better): if `_atom_mappings` is normalized to a list, the translate process can be a comprehension, instead of a `map`:

```
mapping_dict = (x.old_to_new_atom_map for x in in _atom_mappings)
```

This will be valid for the case where `_atom_mappings == []`, and removes the need for the if statement.

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -50,24 +51,31 @@ def __init__(self, allow_ring_breaking: bool = True,
...
         except InvalidMappingException:
             _atom_mappings = []
 
+        # Catch empty mappings here
+        if (_atom_mappings is None):
+            _atom_mappings = []
+
         # Post processing
         if (self.preserve_chirality):
             [x.preserve_chirality() for x in _atom_mappings]
```
dwhswenson, line 73:
```suggestion
        if self.preserve_chirality:
            for x in _atom_mappings:
                x.preserve_chirality()
```

1. [Don't use superfluous parens](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/superfluous-parens.html)
2. [Don't use a list comprehension if you're not assigning it](https://treyhunner.com/2019/03/abusing-and-overusing-list-comprehensions-in-python/#Loops_disguised_as_comprehensions)

openfe/setup/atom_mapping/perses_mapper.py
RiesBen: > For clarification: under what circumstances does perses return `None` instead of raising an `InvalidMappingException`?

hm... I looked in that and it is not entirely trivial. the Exception is thrown if  allow_ring_breaking set to False - leading to an unmapping. if this fails somehow, it will throw an InvalidMappingException.

openfe/setup/atom_mapping/perses_mapper.py
```diff
@@ -50,24 +51,29 @@ def __init__(self, allow_ring_breaking: bool = True,
...
+        # Try generating a mapping
         try:
             _atom_mappings = _atom_mapper.get_all_mappings(
                     old_mol=molA.to_openff(), new_mol=molB.to_openff())
         except InvalidMappingException:
-            _atom_mappings = []
+            return
+
+        # Catch empty mappings here
+        if (_atom_mappings is None):
```
IAlibay, line 68:
```suggestion
        if _atom_mappings is None:
```
Odd @dwhswenson's edit didn't filter through.

IAlibay (approved):
Just the one thing but I'll early approve it so I'm not blocking.

dwhswenson (approved):
Also approving, pending @IAlibay's currently-suggested change.

RiesBen:
@richardjgowers  somehow, git thinks there is a change request by you? but I can not see it anymore and think it might be outdated?

mikemhenry:
This is the request @RiesBen 
image

PR #161, opened 2022-07-21
Fix rms tolerance check

IAlibay:
Add tolerance as a setting for alchemical_settings for now (since it declares whether or not things should really be mapped-ish).

We need this so we can go ahead with our benchmark runs.


IAlibay:
Note: this needs merging as a priority - otherwise we can't have other folks do the benchmarking

richardjgowers (approved):

PR #162, opened 2022-08-02
remove py 3.8 from CI

richardjgowers:


mikemhenry:
Should we wack in python 3.10 as well?

PR #163, opened 2022-08-03
Fix tests

richardjgowers:
we stopped allowing conformer-less molecules in gufe, this fixes tests after that change


dwhswenson (approved):
LGTM. Optionally add the line that PEP8 complains about.

PR #164, opened 2022-08-03
XML testing

dwhswenson:
I did have code in place  for the XML test -- I just hadn't created the fixtures yet. I didn't see it because it I hadn't made a commit on the branch yet. The perses data is the output from https://github.com/OpenFreeEnergy/Experiments. There's not much to this, and it would require that things come out in the same order.

What I found was that for hybrid-system.xml, we didn't get the same order, but for solvated-hybrid-system.xml, we did. I think @RiesBen had found a potential reason for the reordering in some cases? For this, the only differences in the XML should be the expected ones (rounding difference in box vectors, renaming of some internal class, etc.)

@IAlibay : feel free to continue this. The basic approach is just that the only lines that should be different are the ones we expect to be different (and we know what we output for those lines)


mikemhenry:
Looks like it is building this PRhttps://readthedocs.org/projects/openfe/builds/18517799/
But I don't see the check come in on the github web ui side of things

PR #165, opened 2022-08-08
Remove universal bdist_wheel in setup.cfg

IAlibay:
As per [wheels docs](https://wheel.readthedocs.io/en/stable/user_guide.html#building-wheels)

"If your project contains no C extensions and is expected to work on both Python 2 and 3, you will want to tell wheel to produce universal wheels by adding this to your setup.cfg file:" (referring to adding `universal = 1` under `[bdist_wheel]`).

It's surprisingly caused a ton of chaos for some MDA python-only components, so we should probably eagerly remove it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,9 +42,6 @@ test =
 console_scripts = 
     openfe = openfecli.cli:main
 
-[bdist_wheel]
-universal = 1
-
 [versioneer]
 VCS = git
 style = pep440-post


dwhswenson (approved):
+1. This was taken from stuff that had taken stuff from OPS, which swore to maintain Py2.7 support until we release a 2.0 (oh, how I need to release an OPS 2.0). I copied it over, forgetting that this wasn't "pure Python" but rather "pure Python and supports Py2 and Py3". This does not belong in openfe (gufe probably has a similar issue.)

PR #142, opened 2022-05-25
Using gufe protocols

richardjgowers:
fixes #176

PR #171, opened 2022-09-22
quickrun CLI command

### LINKED ISSUE #172: OpenFE CLI: Run a dumped transformation

dwhswenson:
Takes a `ProtocolDAG` saved as JSON, runs the DAG, and returns results.

A few things before this is ready to merge:

- [x] Decide whether the input is a serialized `ProtocolDAG` or a serialized `Transformation`. These are essentially equivalent, but there's some discussion to be had on what it means from user experience.
- [x] Decide on how to ensure serializability of results. In particular, we need to decide how to handle `Path` objects, and whether any non-JSON-serializables other than `Path`s are allowed return values.
  - implemented in [OpenFreeEnergy/gufe#68](https://github.com/OpenFreeEnergy/gufe/pull/68)
- [x] Add a test
- [ ] ~Merge openfreeenergy/gufe#62~ This isn't actually needed here, but will be needed in the future.
- [x] Merge PR that adds `GufeTokenizable` support to atom mappings (included in #142, but might split that part off of that PR)

Resolves #172.


richardjgowers (commented):
I think this will require a `Transformation` not just a DAG.  A DAG doesn't quite have enough info to analyse itself, the results it passes out are paths to raw data, the Transformation has a reference to the Protocol, which has the `gather()` method to analyse that data.

richardjgowers:
Re serialising, since `DAGResults` are consumed by the `Protocol.gather()`, we can just pass strings and know they're paths.

I think where this might become less preferable is when thinking about storage, the storage system isn't privy to knowing what strings are actually paths, and so serialising them as `Path` objects is much nicer

openfecli/commands/quickrun.py
```diff
@@ -0,0 +1,36 @@
...
+)
+@click.argument('dagfile', type=click.File(mode='r'))
+@click.option('-o', '--outfile', type=click.File(mode='w'))
+def quickrun(dagfile, outfile):
+    """Run (in serial) the DAG associated with a given edge.
+    """
+    from gufe.protocols.protocoldag import execute, ProtocolDAG
+    dct = json.load(dagfile)
+    dag = ProtocolDAG.from_dict(dct)
+    result = execute(dag)
```
richardjgowers, line 21: create a `gufe.Context` with a temp dir. and pass in here

openfecli/commands/quickrun.py
```diff
@@ -0,0 +1,53 @@
...
+    creating the transformation):
+
+    \b
+        import json
+        with open(filename, 'w') as f:
+            json.dump(transformation.to_dict(), f)
+    """
+    import gufe
+    from gufe.protocols.protocoldag import execute
+    print("Loading file...")
```
richardjgowers, line 28: this is fine for quick and dirty, but we need to figure out how we're doing logging eventually

openfecli/commands/quickrun.py
```diff
@@ -0,0 +1,53 @@
...
+    dct = json.load(transformation)
```
richardjgowers, line 29: also worth thinking about wrapping each step in a try/except and catching any errors?  (speaking of, what are these errors...)

richardjgowers (commented):
lgtm, would be nice to also test failure modes

openfecli/commands/quickrun.py
dwhswenson: If there are errors that we can handle, then we could. If this is just to avoid dumping a stack trace, then let click do it for it (it does it automatically).

dwhswenson:
So, apparently I had already written a test for this, and I'd just forgotten about it. Assuming the MacOS tests pass, this is ready for review.

openfecli/commands/quickrun.py
```diff
@@ -0,0 +1,53 @@
...
+    trans = gufe.Transformation.from_dict(dct)
+    print("Planning the campaign...")
```
IAlibay, line 31: I suspect these are going to be lost in the amazing amounts of noise our protocol generates. This isn't a problem for now, but we should think about how we deal with our stdout / stderr streams.

openfecli/commands/quickrun.py
```diff
@@ -0,0 +1,53 @@
...
+    dag = trans.create()
+    print("Running the campaign...")
+    dagresult = execute(dag)
```
IAlibay, line 34: We should have a way to pass `shared` here? That way the files aren't temporary?

openfecli/tests/commands/test_quickrun.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+import click
+import importlib
+from click.testing import CliRunner
+
+from openfecli.commands.quickrun import quickrun
+
+def test_quickrun():
+    runner = CliRunner()
```
richardjgowers, line 9: I'm guessing this is doing the same as calling the cli via subprocess?

openfecli/tests/dev/write_transformation_json.py
```diff
@@ -0,0 +1,43 @@
...
+tol_dict = {'ligand': toluene}
+solv_dict = {'solvent': solvent}
+
+solv_benz = gufe.ChemicalSystem(dict(**benz_dict, **solv_dict))
+solv_tol = gufe.ChemicalSystem(dict(**tol_dict, **solv_dict))
+
+mapping = None
+# waiting on AtomMapping being gufe tokenizable
+# mapper = openfe.setup.LomapAtomMapper()
+# mapping = list(mapper.suggest_mappings(benzene, toluene))[0]
```
richardjgowers, line 36: this is possible now btw

richardjgowers (commented):
couple of points but looks good

openfecli/commands/quickrun.py
richardjgowers, line 31: nitpick, this isn't a campaign, more like a single battle (edge)

openfecli/commands/quickrun.py
dwhswenson: Added `-d` / `--directory` option. Can rename to `--shared` if that makes more sense.

Still need to test that; it didn't seem used in a manual test with the `DummyProtocol`, but that may be the protocol.

openfecli/tests/commands/test_quickrun.py
dwhswenson: I'm not sure if it internally splits off a subprocess, but it is similar enough -- catches stdout, stderr, and return code -- so it gives us what we need. It also handles tempdir, etc. in the `isolated_filesystem` context manager.

[This is the recommended way to test `click` applications.](https://click.palletsprojects.com/en/8.1.x/testing/)

openfecli/commands/quickrun.py
```diff
@@ -0,0 +1,53 @@
...
+    print("Done! Analyzing the results....")
+
+    prot_result = trans.protocol.gather([dagresult])
+    estimate = prot_result.get_estimate()
+    uncertainty = prot_result.get_uncertainty()
```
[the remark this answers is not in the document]
dwhswenson: Added `-o` / `--output` option to give a file here. It is a filename relative to cwd (just as the input file is relative to cwd).

What should be done if the file exists? Currently we'll overwrite it, but it's easy to change the `click` setup to require a path that does not already exist.

openfecli/tests/dev/write_transformation_json.py
dwhswenson: good catch; updated to use real mappings now

openfecli/commands/quickrun.py
dwhswenson: I've abstracted this into `openfecli.utils.write`, so we can easily change the details later. FWIW, this is only here because I was curious as to what was taking so long with the dummy protocol (`create` was slow... I think this got fixed by some updates Dotson made).

dwhswenson:
@IAlibay @richardjgowers : This should be ready for another look.

I haven't tested the shared directory stuff yet, since `DummyProtocol` doesn't support it. However, I do pass the shared directory to `execute`, so if there are problems, it will be with the underlying protocols and the CLI wrapper. One thing that isn't clear to me is whether protocols expect the `shared` directory to already exist, or if they create them on the fly if needed.

IAlibay (approved):
this looks fine to me

IAlibay:
It creates the directory. I've not had to create them for things to work for the example notebooks.

richardjgowers (approved):

PR #188, opened 2022-10-21
Fix `results_cls` attribute name

IAlibay:
As per: https://github.com/OpenFreeEnergy/gufe/blob/f58fd82108c2954e9b6604cd1585bf66829f6078/gufe/protocols/protocol.py#L255

It should be `results_cls` not `_results_cls`?


IAlibay:
CI is failing because codecov is down. Ideally if we can merge here that'd be great so I can finish off the notebook things.

richardjgowers (approved):

PR #190, opened 2022-10-21
Make unit close reporter at end of simulation

IAlibay:
Alternatively we could just del() the sampler, which does this: https://github.com/choderalab/openmmtools/blob/2e64980d86cf087f69e282927eebf2b41b73a3ea/openmmtools/multistate/multistatesampler.py#L827-L830

Maybe that's a better option?

### DIFF
--- a/openfe/setup/methods/openmm/equil_rbfe_methods.py
+++ b/openfe/setup/methods/openmm/equil_rbfe_methods.py
@@ -1124,6 +1124,9 @@ class RelativeLigandTransformUnit(gufe.ProtocolUnit):
                 logger.info("running production phase")
 
             sampler.extend(int(prod_steps.m / mc_steps))  # type: ignore
+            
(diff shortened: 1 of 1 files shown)


richardjgowers (commented):
Del doesn't force things, it decrements the reference count. I think better is to explicitly close things

IAlibay:
> Del doesn't force things, it decrements the reference count. I think better is to explicitly close things

Ah yeah I forget that __del__() only gets called once the counter goes to zero.

Well the options here are to manually call the reporter `close()` or call `simulation.__del__()`?

richardjgowers:
Probably makes most sense for the Unit to clean up after the simulation in
the run method.

IAlibay:
> Probably makes most sense for the Unit to clean up after the simulation in the run method.

Both of these options should do that?

richardjgowers (approved):

PR #194, opened 2022-10-26
pin dependencies for release

richardjgowers:
in preparation of next release:

pins pint to fix a current issue
pins openff-units to specific version as it seems very mobile right now
pins openff-toolkit to current 0.11.z release to allow patches but hopefully no breaks

the idea going forwards is to manually move these pins as needed

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,8 +13,8 @@ dependencies:
   - pytest-cov
   - pydantic
   - coverage
-  - openff-toolkit
-  - openff-units>=0.1.6
+  - openff-toolkit==0.11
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -15,6 +15,7 @@ dependencies:
   - coverage
   - openff-toolkit
   - openff-units>=0.1.6
+  - pint==0.19.2
```
IAlibay, line 18:
```suggestion
  - pint<0.2.0
```
?

IAlibay (approved):

PR #204, opened 2022-10-28
[DOCS] Fixing RTD

### LINKED ISSUE #203: CLI docs not working
The docs added in #202 aren't building correctly on RTD. My guess is that RTD isn't installing the CLI into the build environment (build works fine for me locally, but the incorrect build that is live now is exactly what I got when incorrectly labelling the command to document; i.e., it is what you'd get if it couldn't find the command in the environment).

This requires RTD access to debug, which I don't have. I'm `dwhs` over there, if @richardjgowers wants me to deal with this.

dwhswenson:
Resolves #203.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -23,6 +23,9 @@ dependencies:
   - openmmtools
   - openmmforcefields>=0.11.2
   - perses
+  # docs
+  - myst-parser
+  - sphinx-click
   - pip:
     - git+https://github.com/OpenFreeEnergy/gufe@main
     - plugcli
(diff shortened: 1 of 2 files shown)


dwhswenson:
This fixed it: https://openfe.readthedocs.io/en/docs/cli/quickrun.html

Side note: Can we keep this branch (`docs`) alive? I often find it really useful to have a long-lived branch that is a staging ground for docs (especially if we're adding new extensions/major functionality to the docs, and want to be sure that RTD builds it without error before we deploy to `latest`.

richardjgowers (approved):

PR #209, opened 2022-10-28
[WIP] quickrun should error if a unit errors

dwhswenson:
Errors, but saves the details error in the output file (and probably dumps it to stdout as well, though it'll be part of a mess).

- [x] raise error at the end if we get any error
- [ ] test case with a erroring unit


richardjgowers (approved):

dwhswenson:
Ack! No! This was still WIP! No test, and not aimed for this release! It will require a new gufe release!

dwhswenson:
Actually, the stuff that requires new gufe was in a  commit that hadn't been pushed. But what should I do here? This PR was not ready to merge. It will kind of work (not error), but be mostly terrible for the user.

We can revert, or I can start a PR to extend this.

dwhswenson:
I'd argue that this is usable, but still not particularly informative. The commit I haven't pushed is, I think, using some unreleased gufe stuff, so we'll need a new release there.

IMO, we have a usable version, it just doesn't do good error handling. Adding error handling can be in the next deliverable.

PR #212, opened 2022-10-28
[Docs] Installation info

dwhswenson:
Leaving room for someone else (@mikemhenry?) to fill in the single file installer.

Developer install here may be a bit overkill. I'm hoping potential contributes don't actually need the instructions on how to clone a git repo.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,0 +1,103 @@
+Installation
+============
+
+The page has information for installing ``openfe``, installing software
+packages that integrate with ``openfe``, and testing that your ``openfe``
+installation is working.
+
+``openfe`` currently only works on POSIX system (macOS and UNIX/Linux). It
+is tested against Python 3.9 and 3.10.
+
+Installing ``openfe``
+---------------------
+
+When you install ``openfe`` through any of the methods described below, you
+will install both the core library and the command line interface.
+
+Installation with ``conda`` (recommended)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We recommend installing ``openfe`` with ``conda``, because it provides easy
+installation of other tools, including molecular dynamics tools such as
+OpenMM and ambertools, which are needed by ``openfe``. 
+Conda can be installed either with the `full Anaconda distribution
+<https://www.anaconda.com/products/individual>`_, or with
+the `smaller-footprint miniconda
+<https://docs.conda.io/en/latest/miniconda.html>`_. We also recommend
+`mamba
+<https://mamba.readthedocs.io/en/latest/installation.html#installation>`_ as
+a more perfomant variant of ``conda``.
+
+You can get ``openfe`` from the ``conda-forge`` channel. To install the most
+recent release, use the command ::
+
+  $ conda install -c conda-forge openfe
+
+With that, you should be ready to use ``openfe``!
+
+Single file installer
+~~~~~~~~~~~~~~~~~~~~~
+
+[COMING SOON!]
+
+.. TODO: maybe Mike can fill this in? just needs (1) how to download the
+   single file installer; (2) how to use the single file installer
+
+Developer install
+~~~~~~~~~~~~~~~~~
+
+If you're going to be developing for ``openfe``, you will want an
+installation where your changes to the code are immediately reflected in the
+functionality. This is called a "developer" or "editable" installation.
+
+Getting a developer installation for ``openfe`` first installing the
+requirements, and then creating the editable installation. We recommend
+doing that with ``conda`` using the following procedure:
+
+First, clone the ``openfe`` repository, and switch into its root directory::
+
+  $ git clone https://github.com/OpenFreeEnergy/openfe.git
+  $ cd openfe
+
+Next create a ``conda`` environment containing the requirements from the
+specification in that directory::
+
+  $ conda env create -f environment.yml
+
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -0,0 +1,103 @@
...
+We recommend installing ``openfe`` with ``conda``, because it provides easy
+installation of other tools, including molecular dynamics tools such as
+OpenMM and ambertools, which are needed by ``openfe``. 
+Conda can be installed either with the `full Anaconda distribution
+<https://www.anaconda.com/products/individual>`_, or with
+the `smaller-footprint miniconda
+<https://docs.conda.io/en/latest/miniconda.html>`_. We also recommend
+`mamba
+<https://mamba.readthedocs.io/en/latest/installation.html#installation>`_ as
+a more perfomant variant of ``conda``.
```
richardjgowers, line 29: if you're having to describe conda, it probably doesn't help to describe the three different variants all at once.

Pick a thing and run with it.  We "recommend" mamba, but then don't actually use this in the example.

I'd just trim this to; use conda, here's miniconda, done

docs/installation.rst
```diff
@@ -0,0 +1,103 @@
...
+installation is working.
+
+``openfe`` currently only works on POSIX system (macOS and UNIX/Linux). It
+is tested against Python 3.9 and 3.10.
+
+Installing ``openfe``
+---------------------
+
+When you install ``openfe`` through any of the methods described below, you
+will install both the core library and the command line interface.
```
richardjgowers, line 15: I'd include that the CLI is the prefix `openfe` (or is it `openfecli`?)

docs/installation.rst
```diff
@@ -0,0 +1,103 @@
...
+`mamba
+<https://mamba.readthedocs.io/en/latest/installation.html#installation>`_ as
+a more perfomant variant of ``conda``.
+
+You can get ``openfe`` from the ``conda-forge`` channel. To install the most
+recent release, use the command ::
+
+  $ conda install -c conda-forge openfe
+
+With that, you should be ready to use ``openfe``!
```
richardjgowers, line 36: I'd sooner make them make an environment (best practices etc), so `conda create -n openfe -c conda-forge openfe` then `conda activate openfe`.  A bit slower, but avoids the hell of a base env with everything in it

docs/installation.rst
```diff
@@ -0,0 +1,103 @@
...
+Optional dependencies
+---------------------
+
+Certain functionalities are only available if you also install other,
+optional packages.
+
+* **perses tools**: To use perses, you need to install perses and OpenEye,
+  and you need a valid OpenEye license. To install both packages, use::
+
+    $ conda install -c conda-forge -c openeye perses openeye
```
richardjgowers, line 87: `openeye-toolkits` no?

docs/installation.rst
```diff
@@ -0,0 +1,103 @@
...
+
+``openfe`` has a thorough test suite, and running the test suite is a good
+start to troubleshooting any installation problems. The test suite requires
+``pytest`` to run. You can install ``pytest`` with::
+
+  $ conda install -c conda-forge  pytest
+
+Then you can run the test suite (from any directory) with the command::
+
+  $ py.test --pyargs openfe openfecli
```
richardjgowers, line 100: isn't this `pytest` now? I thought the whole `py` prefix was being deprecated (you're showing your age :p)

docs/installation.rst
```diff
@@ -0,0 +1,103 @@
...
+
+The test suite contains several hundred individual tests. All should
+complete with status either passed, skipped, or xfailed (expected fail).
```
richardjgowers, line 103: should take a couple minutes right?

docs/installation.rst
richardjgowers: also maybe this would be a good thing to add to the cli... `openfecli validate_installation`

richardjgowers (changes requested):

docs/installation.rst
dwhswenson: Clarify: do you mean: 

* the command, called from your shell, is `openfe`?
* or, the import name for things within the CLI package is `openfecli`?

Both of these statements are true. I tend to think that making a big deal our of how to import the CLI will only lead to confusion (audience here is pretty inexperienced). Importing `openfecli` should almost never be done; exception is external plugins (which we haven't enabled yet and would need to be documented elsewhere if we enabled).

docs/installation.rst
dwhswenson: take a look at the update; I think this'll work

docs/installation.rst
dwhswenson: I've had `alias pytest="py.test"` in my zshrc for so long that it was in my bashrc.

Not sure that I agree on the `validate_installation`, unless it does other things. In particular, not fond of running a subshell, and I don't offhand know how to run pytest from within a python process.

docs/installation.rst
richardjgowers: yeah the shell version, I couldn't remember what it was called

docs/installation.rst
richardjgowers: I think it's just `pytest.main()` https://docs.pytest.org/en/7.1.x/how-to/usage.html#calling-pytest-from-python-code

I'm thinking it would be nice if there was a non-developer way to run the tests, which would pick up things like bad environments, linking problems etc.  Just a smoke test that is ran as part of the install process.

richardjgowers (approved):

PR #217, opened 2022-11-02
Dry run now returns sampler via debug dictionary entry

### LINKED ISSUE #187: Should dry runs returns dicts?
Follow up from #186 

In my opinion dry runs shouldn't have a return type that is similar to the expected return type. Ideally it should really just be `None` or `True`.

IAlibay:
Fixes #187 

The MultiStateSampler should contain everything we need for the openmm rbfe protocols, especially the HTF class that then also includes pointers to the individual component A and B systems / topologies.

### DIFF
--- a/openfe/tests/setup/test_openmm_equil_protocols.py
+++ b/openfe/tests/setup/test_openmm_equil_protocols.py
@@ -7,6 +7,7 @@ from openff.units import unit
 from openff.units.openmm import ensure_quantity
 
 from openmm import app
+from openmmtools.multistate.multistatesampler import MultiStateSampler
 
 from openfe import setup
 from openfe.protocols import openmm_rbfe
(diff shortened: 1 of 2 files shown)


openfe/tests/setup/test_openmm_equil_protocols.py
```diff
@@ -148,15 +149,16 @@ def test_dry_run_default_vacuum(benzene_vacuum_system, toluene_vacuum_system,
...
 
     # create DAG from protocol and take first (and only) work unit from within
-    dag = protocol.create(
-        stateA=benzene_vacuum_system,
-        stateB=toluene_vacuum_system,
-        mapping={'ligand': benzene_to_toluene_mapping},
-    )
-    unit = list(dag.protocol_units)[0]
-
     with tmpdir.as_cwd():
```
IAlibay, line 152: We need to be careful with our tmpdir locations, as far as I understand just calling `protocol.create` will create directories.

richardjgowers: Yeah I'm getting dirs pop up during testing but it won't be in `create()` afaik... I'll track them down

richardjgowers (changes requested):
this is a little overzealous with tmpdir usage, can I take this over @ialibay?

richardjgowers (approved):

IAlibay:
ah nvm, just saw your recent commit

PR #218, opened 2022-11-02
Add --log option to CLI

dwhswenson:
A lot of this was taken from the OpenPathSampling CLI.

See https://github.com/openpathsampling/openpathsampling-cli/pull/15

As with that, note that the `--log` option must come *before* the command, i.e.,:

```
openfe --log logging.conf command arg --opt foo
```

### DIFF
--- a/openfecli/cli.py
+++ b/openfecli/cli.py
@@ -2,6 +2,8 @@
 # For details, see https://github.com/OpenFreeEnergy/openfe
 
 import pathlib
+import logging
+import logging.config
 
 import click
 from plugcli.cli import CLI, CONTEXT_SETTINGS
@@ -34,11 +36,13 @@ the OpenFE Python library.
 @click.command(cls=OpenFECLI, name="openfe", help=_MAIN_HELP,
                context_settings=CONTEXT_SETTINGS)
 @click.version_option(version=openfecli.__version__)
-def main():
-    # currently empty: we can add options at the openfe level (as opposed to
-    # subcommands) by adding click options here. Subcommand runs after this
-    # is the processed.
-    pass
+@click.option('--log', type=click.Path(exists=True, readable=True),
+              help="logging configuration file")
+def main(log):
+    # Subcommand runs after this is processed.
+    # set logging if provided
+    if log:
+        logging.config.fileConfig(log, disable_existing_loggers=False)
 
 
 if __name__ == "__main__":  # -no-cov- (useful in debugging)
--- a/openfecli/tests/test_cli.py
+++ b/openfecli/tests/test_cli.py
@@ -5,11 +5,11 @@ import click.testing
 from openfecli.cli import OpenFECLI, main
 from openfecli.plugins import OFECommandPlugin
 
+@pytest.fixture
+def cli():
+    return OpenFECLI()
 
 class TestCLI:
-    def setup(self):
-        self.cli = OpenFECLI()
-
     def test_invoke(self):
         runner = click.testing.CliRunner()
         with runner.isolated_filesystem():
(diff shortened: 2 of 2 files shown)


dwhswenson:
This is ready for review

richardjgowers:
Why doesn't click allow arbitrary ordering of arguments?

openfecli/cli.py
```diff
@@ -34,11 +36,13 @@ def get_installed_plugins(self):
 @click.command(cls=OpenFECLI, name="openfe", help=_MAIN_HELP,
                context_settings=CONTEXT_SETTINGS)
 @click.version_option(version=openfecli.__version__)
-def main():
-    # currently empty: we can add options at the openfe level (as opposed to
-    # subcommands) by adding click options here. Subcommand runs after this
-    # is the processed.
-    pass
+@click.option('--log', type=click.Path(exists=True, readable=True),
+              help="logging configuration file")
```
richardjgowers, line 40: I'd also like `--log=True` or something rather than having to mess with files.

richardjgowers (changes requested):
I think this PR might make more sense to me if I could see an example of how logging would work inside the library to connect to this.

richardjgowers:
I've just seen the gufe PR that is the other half to this.. so I think maybe my comment isn't valid

dwhswenson:
> Why doesn't click allow arbitrary ordering of arguments?

This is because, in click, `openfe` is a command and, e.g., `quickrun` is a subcommand. Separate arguments are associated at each level. Fundamentally, you're changing the behavior of the whole `openfe` command, regardless of subcommand.

Having subcommands arguments interpreted depending on order is standard in CLI world. Let's look at `conda`:

* `conda env list`: list conda environments
* `conda list env`: list installed version of package `env` in current environment

or `git`:

* `git add -v`: Use the `--verbose` flag to `git add`
* `git -v add`: Use the `--version` flag to `git` (show version and exit)

This is the correct way to handle arguments to avoid conflicts between main command and subcommand: we process from first thing written to last thing written.

Other options are:

* force every command to have a `--log` option
* have `--log` option at the level of both `openfe` and its subcommands

I considered both of these options when I first implemented this. I decided this was the best approach for a number of reasons:

* this is fundamentally something where we're changing the way OpenFE works, so this approach makes sense
* there may be other things we want in the future that do the same (e.g., imagine in the future we allow rcfiles that change `openfe` behavior -- specifying the rcfile at the command line is something we'd want to enable)
* allowing at both levels is confusing because it can be used twice; expected behavior in that case is ambiguous
* requiring every command to have a `--log` option is effectively reserving a keyword that they can't use; e.g., `plot --log` (referring to axes) isn't unreasonable
* requiring every command to have a `--log` option is tedious, and a likely source of error. The obvious approach to automate that will make `--log` the first option shown, whereas it should be the last option shown. I'm not entirely sure if there's a simple way to push it so that `--log` is the last option (I mean, `click` is awesome, so there might be.)
* this is an advanced use case, typically only used when trouble shooting. Most of the time when a user does this, it will be because we said, "here, rerun this as `openfe --log file-we-sent.conf $WHATEVER_YOU_DID`." Users will copy-paste. Little room for confusion.

openfecli/cli.py
dwhswenson: What would `--log=True` mean? What level of logging are we doing?

How about I allow `DEBUG`, `INFO`, `WARNING`, `ERROR`, or `CRITICAL` (actually, I would allow any capitalization of those, and if you really wanted, I could allow their numeric equivalents as well).

If I do that, what do you want it to apply to? Loggers in `openfe`? Loggers in `openfe` and `gufe`? All loggers? (that would give all debug output from `openmmtools`, `openmm`, `numpy`, etc. -- this is the point of the file-based setup; you can select which loggers you want)

richardjgowers (approved):

PR #225, opened 2022-11-11
Bug hunt fixes

IAlibay:
Work done in this PR:

- Up to 0b71e23 : Fixes the constraint demapping call (a typo in the assigned order of the map items 😓 )
  - If we can follow this up with some actual tests that'd be grand. For now this fully replicate's perses.
- 35f37ee : Add virtual bond for imaging which was missed from the vendored HTF code
  - Removed the "center in box" code, which was both problematic and uncessary.
- Added minimize call on creation of thermodynamic states
- Added experimental code for CPU fallback when GPU minimisation fails (possibly remove).

To-do:
- XML diff tests of p38 edge and benzene/phenol edge
- Coordinates diffs tests


openfe/protocols/openmm_rbfe/_rbfe_utils/multistate.py
```diff
@@ -115,8 +116,7 @@ class creation of LambdaProtocol.
             context, context_integrator = context_cache.get_context(
                                              compound_thermostate_copy)
             # TODO: move to our own MD tasks
-            # feptasks.minimize(compound_thermodynamic_state_copy,
-            #                   sampler_state)
+            #minimize(compound_thermostate_copy, sampler_state)
```
IAlibay, line 119: The original perses code does the minimization here, however in my opinion it's a bit problematic to have it hidden here. It would be better for this to be a) controlled, b) easy to spot in the protocol.

openfe/protocols/openmm_rbfe/_rbfe_utils/multistate.py
```diff
@@ -143,6 +148,79 @@ class creation of LambdaProtocol.
...
+            final_energy = thermodynamic_state.reduced_potential(sampler_state)
+
+        logger.debug('Replica {}/{}: final energy {:8.3f}kT'.format(
+            replica_id + 1, self.n_replicas, final_energy))
+
+        # Clean up the integrator
+        del context
+
+        # Return minimized positions.
+        return sampler_state.positions
```
IAlibay, line 222: This probably can go away for now, but we should try to push something like this to openmmtools. Ideally we need a) optional platform choice for minimisation (that isn't in the context cache), and b) fallback to CPU L-BFGS when GPU optimization doesn't get anywhere.

Also - I think here we want the positions to be reset to their pre GPU minimised values before the CPU L-BFGS fallback. I've had some interesting failures yesterday that indicate that the failed GPU minimisation might take things to a bad configurational state.

openfe/protocols/openmm_rbfe/_rbfe_utils/topologyhelpers.py
```diff
@@ -155,7 +155,7 @@ def _remove_constraints(old_to_new_atom_map, old_system, old_topology,
             new_constraints[atom2] = length
 
     to_del = []
-    for new_index, old_index in old_to_new_atom_map.items():
+    for old_index, new_index in old_to_new_atom_map.items():
```
IAlibay, line 158: As per the tasks board, we need to add tests in a separate PR.

richardjgowers:
test for constraint detection added

IAlibay:
@RiesBen can this gathering stuff not be in a separate PR? It doesn't seem like part of the original bug issue?

IAlibay:
Virtual bond addition doesn't seem to be working for inter-chain things, for some reason in one of my system I seem to be getting a bond between a monomer protein and water O.o

richardjgowers (approved):

openfe/protocols/openmm_rbfe/_rbfe_utils/multistate.py
```diff
@@ -143,6 +148,79 @@ class creation of LambdaProtocol.
             self.create(thermodynamic_states=thermodynamic_state_list,
                         sampler_states=sampler_state_list, storage=reporter)
 
+    def _minimize_replica(self, replica_id, tolerance, max_iterations):
```
IAlibay, line 151: As discussed, let's remove this for now and move it to a separate PR

PR #227, opened 2022-11-17
Rewrite constraint finding

richardjgowers:
this adds removal of constraints disappearing into harmonic bonds, with tests for a toluene to benzene system

it's limited to just H constraints, as it leans heavily on the fact that the way to resolve is to remove the -H from the mapping and that this doesn't create any disconnections (H are degree 1 always)

I think we can probably get rid of the two O(natoms) scans to find hydrogens if we pass the entire `Mapping` object in, as this has references to the molecules (to check what's a hydrogen)

there isn't a way to selectively filter which constraints we iterate over, openmm just gives them all.

### DIFF
--- a/openfe/protocols/openmm_rbfe/_rbfe_utils/topologyhelpers.py
+++ b/openfe/protocols/openmm_rbfe/_rbfe_utils/topologyhelpers.py
@@ -133,40 +133,6 @@ def _remove_constraints(old_to_new_atom_map, old_system, old_topology,
     * Very slow, needs refactoring
     * Can we drop having topologies as inputs here?
     """
-    def get_the_H(A, B, C, D, oldH, newH) -> int:
-        """
-        one of A, B, C, D is a hydrogen, we must figure out which
-        its like a very bad sequel to knives out
-
-        these indices are from a mapping such that:
-          A->C, B->D
-
-        if A==H-> return A
-        elif B==H -> return B
-        elif C==H -> return A (to remove A->C from map)
-        elif D==H -> return B (to remove B->D from map)
-
-        Parameters
-        ----------
-        A, B, C, D: int
-        oldH, newH: set[int]
-
-        Returns
-        -------
-        H: int
-          the index of the hydrogen (or atom that maps to hydrogen)
-        """
-        if A in oldH:
-            return A
-        elif B in oldH:
-            return B
-        elif C in newH:
-            return A
-        elif D in newH:
-            return B
-        else:
-            raise ValueError("Fatal: couldn't find H in constraint")
-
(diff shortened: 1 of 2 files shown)


IAlibay (changes requested):
No time to review this week, but for now we need to put a block on this until we have proper tests of XML diffs against perses.

We need to a) make sure we retain the right behaviour for diverse cases (benzene transformations are too simple as we've seen in the latest p38 problems), b) make sure if we've improved things that we have tests that specifically show this.

IAlibay:
To add here, what I'm suggesting is an optional "slow" tests for maybe a dozen systems from the plb where we check that we get the right behaviour?

richardjgowers:
Yeah, the nice thing about toy systems is there's one offending constraint here and I checked it by hand, so it's a more "analytical" unit test, but regression tests would also be required. And yeah, we spoke about xml regression tests in the call today.

IAlibay:
The issue with sticking to this one case is that we know that this specific test case is also easy to get wrong (indeed relying on this transformation is why we even had one of the bugs), so I'd be keen for us not to rely on it. something with a C-X to C-H change would be good here.

richardjgowers:
@IAlibay toluene to benzonitrile (both directions) test added

richardjgowers:
I'm going to double check how Ivan is doing this in perses, I think his approach is more elegant if it can pass the same tests

openfe/protocols/openmm_rbfe/_rbfe_utils/topologyhelpers.py
```diff
@@ -130,39 +133,93 @@ def _remove_constraints(old_to_new_atom_map, old_system, old_topology,
     * Very slow, needs refactoring
     * Can we drop having topologies as inputs here?
     """
+    def get_the_H(A, B, C, D, oldH, newH) -> int:
```
IAlibay, line 136: so a curveball towards generalisability here, what if we have a constrained bond that doesn't involve a hydrogen?

IAlibay: For the current case, is there a way we could just check and raise an error? I don't particularly think we need to support this edge case just know to fail it.

richardjgowers: yeah ok, I can add this tomorrow

richardjgowers:
@IAlibay ok I added a test where a constraint changes length and neither atom involved is a Hydrogen, we get an error message complaining about the indices.  (the non aromatic C-C in benzonitrile / toluene is a different length)

IAlibay (approved):
That should work thanks, I've not had a chance to fully compare against Ivan's PR but I think the error should at least cover us.

PR #237, opened 2023-01-10
Absolute free energy protocol (no restraints)

IAlibay:
The aim of this is to add support for calculating hydration free energies.


openfe/protocols/openmm_afe/afe_settings.py
```diff
@@ -0,0 +1,373 @@
...
+* Add support for restraints
+* Improve this docstring by adding an example use case.
+
+"""
+from typing import Optional
+from pydantic import validator
+from openff.units import unit
+from gufe import settings
+
+
```
IAlibay, line 21: @richardjgowers this is pretty much what I'm proposing as to how the RBFE settings should look.

Once we get those settings updated I can go ahead and switch to using a similar parameterisation method as what's used in this protocol to make sure we can have arbitrary small molecules.

IAlibay:
Aside from quickrun test failures and problems building docs (gufe settings is throwing errors), this is good to go.

There's TODOs I want to introduce, but they can be done in separate PRs (i.e. the core functionality is there now).

openfe/protocols/openmm_afe/afe_settings.py
```diff
@@ -0,0 +1,386 @@
...
+    NOTE
+    ----
+    * No solvation will happen if a SolventComponent is not passed.
+    """
+    solvent_model = 'tip3p'
+    """
+    Force field water model to use.
+    Allowed values are; `tip3p`, `spce`, `tip4pew`, and `tip5p`.
+    """
```
lilyminium, line 69: Out of curiosity, where does this load parameters from? OpenMM?

IAlibay: This is the solvent model used for addSolvent, which I believe is using XML files baked into OpenMM. We then override the FF using whatever user provided XML in the SystemGenerator settings (in practice it has to be from openmmforcefields).

PR #241, opened 2023-01-19
update mapping visualization of bond changes

### LINKED ISSUE #239: atom mapping bonds seem red too often
In the mapping visualisation below, 19:14 is in the mapping, which is marked blue as a transmutation.  The bond is coloured red as not mapped, but arguably should be blue as something which is changing?

index

richardjgowers:
Fixes #239

previously all bonds involving changes were highlighted red

now deleted bonds show as red, altered bonds show as blue

I've included two screenshots showing the difference.  I think now the ring breaking is easier to see

index
index2

### DIFF
--- a/openfe/tests/utils/test_visualization.py
+++ b/openfe/tests/utils/test_visualization.py
@@ -69,11 +69,11 @@ def benzene_phenol_mapping(benzene_transforms, maps):
     return mapping, mol1, mol2
 
 
-@pytest.mark.parametrize('molname, atoms, elems, bonds', [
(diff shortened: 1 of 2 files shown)


richardjgowers:
one thing that I'm still not quite happy about is that atom 20(left)/21(right) is unchanged, but this is hard to see behind the bond highlight.

richardjgowers:
A more simple example, what the vis looks like with and without element changes allowed.  

index3
index4

RiesBen (approved):

PR #245, opened 2023-01-22
Maximal network planner

dwhswenson:
This adds a ligand network planner to create as many edges as the mappers can propose. This isn't the same as all-to-all, since some changes might be beyond that a mapper can create a mapping for, but it will be as close to all-to-all as we can propose.

It will be particularly useful in developing networks and in debugging. Already, this is really just a refactor of stuff that had been part of the minimal spanning graph.

Tasks:
- [x] add network planner
- [x] refactor minimal spanning graph to reuse this code
- [x] docstrings
- [x] unit tests

I was using this already for some debugging stuff; thought it was better to push the partially completed code up. If someone else wants to take it to the finish line with docstrings and tests, please do.


openfe/setup/ligand_network_planning.py
```diff
@@ -74,6 +74,26 @@ def generate_radial_network(ligands: Iterable[SmallMoleculeComponent],
...
+    scorer: Optional[Callable[[LigandAtomMapper], float]] = None,
+    # allow_disconnected=True
+):
+    nodes = list(ligands)
+
+    # First create a network with all the proposed mappings (scored)
+    mapping_generator = itertools.chain.from_iterable(
+        mapper.suggest_mappings(molA, molB)
+        for molA, molB in itertools.combinations(nodes, 2)
+        for mapper in mappers
```
richardjgowers, line 89: will this generate multiple edges between two nodes for multiple mappers?

dwhswenson: Yes. In fact, this is already tested (by way of the minimum spanning graph, which uses this in its first stage). This will try to create all the edges possible, with all the mappers that are provided.

richardjgowers: Hmm, I would have expected it to instead choose the edge with the best score amongst the mappers... but I can also see this way round too.

richardjgowers: @RiesBen / @IAlibay would be interested what you'd expect from this and/or is there value in producing a network with multiple edges from different mappers (if so maybe these two different behaviours are controlled by a kwarg)

IAlibay: So the only case I personally can think of where this might be useful, is when you end up trying to benchmark various mapping methods against each other. That being said, without any kind of annotation of what the mapper was, it ends up being rather useless.

It's unclear to me what is the intent of this method. Is it:

1. Create a network that just has all the possible connections that could ever be generated (which this does, but will need better annotations).
2. Create as fully connected a network as possible (which this doesn't but could potentially use this method to then filter by score)
3. Something else

If the aim is the first one, then we will simply need a second method that just filters through the results of this method to generate a connected network with a single edge per node pair.

richardjgowers (changes requested):
tests

openfe/tests/setup/test_network_planning.py
```diff
@@ -90,6 +90,47 @@ def test_radial_network_failure(atom_mapping_basic_test_files):
         )
 
 
+@pytest.mark.parametrize('with_scorer', [True, False])
+@pytest.mark.parametrize('extra_mapper', [True, False])
```
IAlibay, line 94: Whilst I'm a man that loves his parametrize, I do wonder if this might be easier to read if it was 3 tests 😅

openfe/setup/ligand_network_planning.py
richardjgowers: ok we're going with multiple and it's documented so all good

openfe/setup/ligand_network_planning.py
```diff
@@ -75,33 +75,69 @@ def generate_radial_network(ligands: Iterable[SmallMoleculeComponent],
...
+def generate_maximal_network(
+    ligands: Iterable[SmallMoleculeComponent],
+    mappers: Iterable[LigandAtomMapper],
+    scorer: Optional[Callable[[LigandAtomMapper], float]] = None,
+    # allow_disconnected=True
+):
+    """Create a network with all possible proposed mappings.
+
+    This will attempt to create (and optionally score) all possible mappings
+    ($N(N-1)/2$ for each mapper given). This is typically used as the
```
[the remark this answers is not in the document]
richardjgowers: the language is "attempt to" but maybe it should be explicit that if an edge is unmappable then you'll get missing edges

openfe/tests/setup/test_network_planning.py
richardjgowers: this is probably fine, it's one test doing the "same thing"

openfe/setup/ligand_network_planning.py
```diff
@@ -75,33 +75,69 @@ def generate_radial_network(ligands: Iterable[SmallMoleculeComponent],
...
-      the ligands to include in the Network
+      the ligands to include in the LigandNetwork
     mappers : Iterable[LigandAtomMapper]
       the AtomMappers to use to propose mappings.  At least 1 required,
       but many can be given, in which case all will be tried to find the
       lowest score edges
     scorer : Scoring function
       any callable which takes a LigandAtomMapping and returns a float
-    """
+     """
```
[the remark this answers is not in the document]
dwhswenson: Copy-paste using visual mode in vim. Must have caught a leading space by accident. Fixed.

openfe/tests/setup/test_network_planning.py
dwhswenson: Was in a hurry to get to some evening plans. 😆 But I'm still happy to leave as-is.

PR #248, opened 2023-01-27
adding 3D structure Mapping Vis to LigandMapping

RiesBen:
Here the visualization functions are added in order to visualize the mapping in 3D.

new dependency: py3DMol

New feature:
```
lomap_mapper = LomapAtomMapper()
lomap_mapping = next(lomap_mapper.suggest_mappings(SmallMoleculeComponent(molA) , SmallMoleculeComponent(molB)))
lomap_mapping.show_mapping3D()

```
image

### DIFF
--- a/openfe/utils/visualization_3D.py
+++ b/openfe/utils/visualization_3D.py
@@ -10,7 +10,7 @@ from matplotlib.colors import rgb2hex
 
 from gufe.mapping import AtomMapping
 
-
+ 
 def draw_mapping_on_3Dstructure(
     edge: AtomMapping,
     spheres: bool = True,
@@ -18,6 +18,7 @@ def draw_mapping_on_3Dstructure(
     style: str = "stick",
     shift: Tuple[float, float, float] = (10, 0, 0),
 ) -> py3Dmol.view:
+
     """
     Render relative transformation edge in 3D using py3Dmol.
 
--- a/openfe/utils/visualization_3D.py
+++ b/openfe/utils/visualization_3D.py
@@ -1,16 +1,21 @@
-import py3Dmol
 import numpy as np
 from numpy.typing import NDArray
 from typing import Tuple, Union, Optional
-
 
 from rdkit import Chem
 from rdkit.Geometry.rdGeometry import Point3D
 from matplotlib import pyplot as plt
 from matplotlib.colors import rgb2hex
 
+try:
+    import py3Dmol
+except ImportError:
+    pass    # Don't throw  error, will happen later
+
+
 from gufe.mapping import AtomMapping
 
+from openfe.utils import requires_package
 
 def _get_max_dist_in_x(atom_mapping: AtomMapping) -> float:
     """helper function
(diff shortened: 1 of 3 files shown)


openfe/setup/atom_mapping/ligandatommapping.py
```diff
@@ -2,12 +2,15 @@
 # For details, see https://github.com/OpenFreeEnergy/openfe
 import json
 import gufe
-from typing import Any, Optional
 import numpy as np
+from typing import Any, Optional
 from numpy.typing import NDArray
 
+from py3Dmol import view
```
IAlibay, line 9: We'll probably want to make this an optional dependency like perses. py3dmol can be quite heavy with all the node.js things

openfe/setup/atom_mapping/ligandatommapping.py
```diff
@@ -147,6 +150,23 @@ def draw_to_file(self, fname: str, d2d=None):
...
+        ----------
+        show_atomIDs : bool, optional
+            show the atomIDs of the Molecules, by default True
+
+        Returns
+        -------
+        view
+            the visualization of the mapping on the 3D structure of the components, can be further adjusted
+        """
+        view = draw_mapping_on_3Dstructure(self, show_atomIDs=show_atomIDs)
```
IAlibay, line 167: Was draw_mapping_3d not long enough :P

IAlibay (changes requested):
Just the one comment re: optional deps. Do we have tests for these visualization methods? Might be worth adding one if possible, at the very least something that flushes the code path will tell us if things break horribly.

IAlibay:
o also mypy is annoyed for some reason

openfe/setup/atom_mapping/ligandatommapping.py
RiesBen: hihi

RiesBen:
> Just the one comment re: optional deps. Do we have tests for these visualization methods? Might be worth adding one if possible, at the very least something that flushes the code path will tell us if things break horribly.

added a smoke test for this... hm not very nice but maybe sufficient?

richardjgowers (changes requested):
this might need a rework due to some refactoring I'm about to do

RiesBen:
@richardjgowers HA, now I want to move this to kartograf! :P what do you think? :)

richardjgowers:
@RiesBen nah this is useful for *all* mapping objects, so it's openfe :P

RiesBen:
I tried :)

openfe/utils/visualization_3D.py
```diff
@@ -0,0 +1,186 @@
...
+                "center": {"x": p2.x, "y": p2.y, "z": p2.z},
+                "radius": 0.6,
+                "color": color,
+                "alpha": 0.8,
+            }
+        )
+
+
+@requires_package("py3Dmol")
+def draw_mapping_on_3Dstructure(
```
richardjgowers, line 111: the only thing I don't like is the name... `draw_3D_mapping`?  `show_3D_mapping`?

richardjgowers (dismissed):

openfe/utils/visualization_3D.py
RiesBen: hehe ok now you and Irfan are two :) let's go with show_3D_mapping. :)

richardjgowers (approved):

PR #263, opened 2023-02-17
add custom codec for OpenMM Units in RBFE Protocol

richardjgowers:
fixes `quickrun ... -o` not working because of unserialisable Quantities

### DIFF
--- a/openfe/tests/setup/test_openmm_equil_protocols.py
+++ b/openfe/tests/setup/test_openmm_equil_protocols.py
@@ -1,5 +1,6 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/openfe
+import json
 import gufe
 import pytest
 from unittest import mock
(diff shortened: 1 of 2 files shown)


openfe/tests/setup/test_openmm_equil_protocols.py
```diff
@@ -751,3 +753,20 @@ def test_constraints(tyk2_xml, tyk2_reference_xml):
             assert a.get('p1') == b.get('p1')
             assert a.get('p2') == b.get('p2')
             assert float(a.get('d')) == pytest.approx(float(b.get('d')))
+
+
+def test_custom_openmm_unit_codec():
```
dwhswenson, line 758: Note that this test doesn't check the case that you're starting analysis without having imported this `Protocol`. There's a potential issue there. We might dodge it in this specific case, but it is likely to be a problem in a more general case.

Possible workaround proposed here: https://github.com/OpenFreeEnergy/gufe/pull/68#issuecomment-1267441746

dwhswenson (commented):
Did you check that it actually fixes enough to get to the next stage in the pipeline? I'm not convinced this does anything other than push to the problem to analysis.

PR #265, opened 2023-02-21
Silence root logging (on perses imports)

dwhswenson:
This provides a context manager intended to wrap imports that use `logging.basicConfig`.

`logging.basicConfig` is aimed at users. Libraries shouldn't use it. This temporarily attaches a handler to the root logger (which makes `basicConfig` a no-op). It then resets the root logger handlers, in case someone in user-land had set them previously.

I haven't tested that this fixes annoyances of overly-verbose logging (specifically due to perses) while waiting for the upstream fix to be released; but this should be a solution.

### DIFF
--- a/openfe/setup/atom_mapping/perses_mapper.py
+++ b/openfe/setup/atom_mapping/perses_mapper.py
@@ -9,8 +9,12 @@ The MCS class from Perses shamelessly wrapped and used here to match our API.
 from openmm import unit
 from openfe.utils import requires_package
 
+from ...utils.silence_root_logging import silence_root_logging
 try:
-    from perses.rjmc.atom_mapping import AtomMapper, InvalidMappingException
+    with silence_root_logging():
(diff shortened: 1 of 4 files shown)


dwhswenson:
@richardjgowers : this is the solution I was proposing to block other things from setting logging levels on import. I think you had stuff set up to quickly check if this will silence things in the use case where it was annoying; if so, can you check that?

I can add unit tests here, but an E2E test may be a bit of a headache to set up. (Also fine with merging this without unit tests; this is just a hack to make it easy to work around an upstream problem, and something we might throw away when we no longer need it.)

mikemhenry:
Our plan is to get a release out on 2023-02-28 which should then make this fix irrelevant, but I think it would be better to get this merged in instead of wait. @richardjgowers can you test to see if this PR fixes things for you?

richardjgowers (approved):
yep this seems to improve the verbosity of quickrun

PR #266, opened 2023-02-21
Storage migration: Add to `openfe`

dwhswenson:
This adds the storage files that had been in `gufe` into `openfe`. Specifically, this includes:

* `metadatastore.py`
* `resultclient.py`
* `resultserver.py`

and relevant tests.

We may consider renaming the client/server approach; these names were because originally this would be shared with other projects like `alchemiscale` as part of `gufe`, but from an `openfe` perspective, we might prefer "client" to just be `Storage` and "server" to be something like `Backend`.

Sibling PR on `gufe`: OpenFreeEnergy/gufe#131

### DIFF
--- a/openfe/storage/resultserver.py
+++ b/openfe/storage/resultserver.py
@@ -1,0 +1,65 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/gufe
+import warnings
+
+from typing import ClassVar
+
+from gufe.storage.errors import (
+    MissingExternalResourceError, ChangedExternalResourceError
+)
+
+
+class ResultServer:
+    """Class to manage communication between metadata and data storage.
+
+    At this level, we provide an abstraction where client code no longer
+    needs to be aware of the nature of the metadata, or even that it exists.
+    """
+    def __init__(self, external_store, metadata_store):
+        self.external_store = external_store
+        self.metadata_store = metadata_store
+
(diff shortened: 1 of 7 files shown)


openfe/tests/storage/conftest.py
```diff
@@ -0,0 +1 @@
+from gufe.tests.conftest import *
```
richardjgowers, line 1: I'm not completely happy with this, it's a bit annoying to have to jump between two packages' conftests (which are already a bit of an opaque pytest magic), but I think we're due a testdata reorganisation in general

openfe/storage/resultserver.py
```diff
@@ -0,0 +1,65 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/gufe
```
richardjgowers, line 2: wrong link here, but I suspect there's a few files either missing or with the wrong license header since we're switching things quite liberally

richardjgowers (approved):

openfe/tests/storage/conftest.py
dwhswenson: IMO, we should generally be doing exactly this: reusing `gufe`'s fixtures. Otherwise we're doubling maintenance burden on fixtures (not to mention increasing confusion, since fixtures aren't consistent between the packages). This isn't just about the data, but about the more complicated fixtures that are more than just loading data from a file.

PR #267, opened 2023-02-23
using gufe Settings in RBFE Protocol

richardjgowers:


openfe/tests/protocols/test_openmm_equil_protocols.py
IAlibay, line 1: Would it be possible to have a test that checks the default settings against a serialised version of the settings?

My take here is that whilst we _should_ know that things are changing upstream, it would be easy for us to forget to change / review its impact downstream. At least if it fails here we can go "yeah ok, that's fine" (or not).

IAlibay (changes requested):
Still digging upstream for a few things but here's a first comment - we need to be more vigilant in picking up cases where settings changes happen, otherwise we'll one day walk into a minefield.

IAlibay (changes requested):
Still digging upstream for a few things but here's a first comment - we need to be more vigilant in picking up cases where settings changes happen, otherwise we'll one day walk into a minefield.

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
```diff
@@ -570,25 +509,33 @@ def _default_settings(cls) -> RelativeLigandProtocolSettings:
...
-                equilibration_length=2.0 * unit.nanosecond,
-                production_length=5.0 * unit.nanosecond,
+            protocol_settings=RelativeLigandProtocolSettings(
+                system_settings=SystemSettings(
+                    constraints='HBonds'
+                ),
+                topology_settings=TopologySettings(
+                    forcefield={'protein': 'amber99sb.xml',
+                                'ligand': 'openff-2.0.0.offxml',
+                                'solvent': 'tip3p.xml'},
```
IAlibay, line 529: this is a good moment to update this:
```suggestion
                    forcefield={'protein': 'amber/ff14SB.xml',
                                'ligand': 'openff-2.0.0.offxml',
                                'solvent': 'amber/tip3p_standard.xml'
                                'ions': 'amber/tip3p_HFE_multivalent.xml',},
```

IAlibay: Indeed those are the forcefield settings found under `OpenMMSystemGeneratorFFSettings`.

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
```diff
@@ -570,25 +509,33 @@ def _default_settings(cls) -> RelativeLigandProtocolSettings:
...
-            system_settings=SystemSettings(
-                constraints='HBonds'
-            ),
-            topology_settings=TopologySettings(
-                forcefield={'protein': 'amber99sb.xml',
-                            'ligand': 'openff-2.0.0.offxml',
-                            'solvent': 'tip3p.xml'},
+        return settings.Settings(
+            settings_version=1,
+            forcefield_settings=settings.OpenMMSystemGeneratorFFSettings(),
```
IAlibay, line 517: Why are we setting forcefields here and then again below  under protocol settings but using twto different objects?

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
```diff
@@ -264,31 +263,16 @@ class BarostatSettings(BaseModel):
     class Config:
         arbitrary_types_allowed = True
 
-    pressure = 1 * unit.bar
     frequency = 25 * unit.timestep
```
IAlibay, line 266: It might make more sense to just move this to integrator settings (renamed as `barostat_frequency`), in reality this is just a "how often should my integration loop stop and do an MC barostat move).

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
```diff
@@ -838,19 +762,19 @@ def run(self, dry=False, verbose=True, basepath=None) -> dict[str, Any]:
 
         #  1. Get smirnoff template generators
         smirnoff_stateA = SMIRNOFFTemplateGenerator(
-            forcefield=settings.topology_settings.forcefield['ligand'],
+            forcefield=topology_settings.forcefield['ligand'],
```
IAlibay, line 765: switch to `forcefield_settings.small_molecule_forcefield`?

IAlibay (changes requested):
I'm not going to block over the restructuring / merging of leftover setting classes (that can be dealt with later if necessary).

I am however going to block over the duplication of settings between what gets defined under `settings.forcefield_settings`, and elsewhere. That's probably a nightmare waiting to happen.

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
```diff
@@ -1166,7 +1082,7 @@ def _execute(
         outputs = self.run(basepath=mypath)
 
         return {
-            'repeat_id': self.repeat_id,
-            'generation': self.generation,
+            'repeat_id': self.inputs['repeat_id'],
+            'generation': self.inputs['generation'],
```
IAlibay, line 1086:
```suggestion
            'repeat_id': self._inputs['repeat_id'],
            'generation': self._inputs['generation'],
```

IAlibay (changes requested):
found this out the hard way this evening

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
```diff
@@ -803,15 +356,29 @@ def run(self, dry=False, verbose=True, basepath=None) -> dict[str, Any]:
...
+        thermo_settings: settings.ThermoSettings = prototol_settings.thermo_settings
+        alchem_settings: AlchemicalSettings = prototol_settings.alchemical_settings
+        system_settings: SystemSettings = prototol_settings.system_settings
+        solvation_settings: SolvationSettings = prototol_settings.solvation_settings
+        sampler_settings: AlchemicalSamplerSettings = prototol_settings.alchemical_sampler_settings
+        sim_settings: SimulationSettings = prototol_settings.simulation_settings
+        timestep = prototol_settings.integrator_settings.timestep
+        mc_steps = prototol_settings.integrator_settings.n_steps.m
+
+        # is the timestep good for the mass?
```
IAlibay, line 376: Is there a way to do this at the validator level? It feels like there should be, but I couldn't find a way to use a validator to validate across two pydantic classes as attributes of a model.

IAlibay: The answer _should_ be to use a root_validator in the pydantic settings child class.

_However_, applying validators on nested models is all kinds of broken in pydantic right now (it's meant to be fixed in the v2.0 release). We might just have to keep there here until v2.0 gets released properly.

richardjgowers: yeah I think we won't be able to get validators to do all the heavy lifting, but they'll at least do a good level of sanity checking

IAlibay:
Requires #293 for CI to pass

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
```diff
@@ -41,428 +40,18 @@
...
-    @validator('collision_rate')
-    def must_be_inverse_time(cls, v):
-        if not v.is_compatible_with(1 / unit.picosecond):
-            raise ValueError("collision_rate must be in inverse time "
-                             "(i.e. 1/picoseconds)")
-        return v
-
-
-class SimulationSettings(BaseModel):
-    """Settings for simulation control, including lengths, writing to disk,
```
IAlibay, line 354: For tomorrow - I just realised, removing these is probably going to make it impossible to look at the docstring programmattically right? Should we keep them in and then also have the inline docstrings?

IAlibay (changes requested):
Going to block on this question.

IAlibay:
This is how it looks if you interrogate a settings class right now:

image

openfe/protocols/openmm_rbfe/__init__.py
```diff
@@ -5,7 +5,10 @@
 
 from .equil_rbfe_methods import (
     RelativeLigandProtocol,
-    RelativeLigandProtocolSettings,
     RelativeLigandProtocolResult,
     RelativeLigandProtocolUnit,
 )
+
+from .equil_rbfe_settings import (
```
richardjgowers, line 12: since Protocol depends on Settings, would it make more sense to import this first?

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
```diff
@@ -803,15 +356,29 @@ def run(self, dry=False, verbose=True, basepath=None) -> dict[str, Any]:
...
+        sim_settings: SimulationSettings = prototol_settings.simulation_settings
+        timestep = prototol_settings.integrator_settings.timestep
+        mc_steps = prototol_settings.integrator_settings.n_steps.m
+
+        # is the timestep good for the mass?
+        if forcefield_settings.hydrogen_mass < 3.0:
+            if timestep > 2.0 * unit.femtoseconds:
+                errmsg = (f"timestep {timestep} too large for "
+                          "hydrogen mass {forcefield_settings.hydrogen_mass}")
+                raise ValueError(errmsg)
```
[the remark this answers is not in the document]
richardjgowers: or do we call it `-ffast-science`?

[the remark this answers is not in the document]
IAlibay: I think this definitely falls on the "this is a terrible idea" line. However, we can probably do a major review of how we validate things for 0.8?

openfe/protocols/openmm_rbfe/__init__.py
IAlibay: do you mean moving it up the import list? sure

richardjgowers: yes, it looks like a circular import?  I guess they don't exist any more

openfe/protocols/openmm_rbfe/equil_rbfe_methods.py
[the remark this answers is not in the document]
richardjgowers: yeah, probably a fun chat to have, how do we want the safety rails to behave

IAlibay (approved):
happy with the status of this, will defer to Richard for approval / merge

IAlibay:
```
Object of type RelativeLigandProtocolSettings is not JSON serializable
```

:/ yeah I'm giving up on trying to get my mush brain around that one. Settings seem to serialise fine, but maybe I'm missing something obvious.

richardjgowers:
right I think we're gucci, I'll merge this once green

PR #272, opened 2023-03-16
quickrun --directory => --work-dir; cwd as default

dwhswenson:
We don't have proper tests for this, because we don't have a test protocol that deposits files. I think that's in progress in openfreeenergy/gufe#150, though -- marking this WIP until we can reuse that for testing here.


openfecli/commands/quickrun.py
```diff
@@ -23,11 +23,11 @@ def _format_exception(exception) -> str:
 @click.argument('transformation', type=click.File(mode='r'),
                 required=True)
 @click.option(
-    'directory', '-d', default=None,
+    '--work-dir', '-d', default=None,
```
RiesBen, line 26: how about the short option, as `-wd`

RiesBen: I like :)

dwhswenson: > how about the short option, as `-wd`

Short options should be single characters, following the [POSIX conventions](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html), with long arguments following the [GNU extension](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html).

Single letters prevent ambiguity when stacking short arguments (e.g., `program -abc` instead of `program -a -b -c`) or when using short opts without a space (e.g., `program -ffilename.txt` instead of `program -f filename.txt`).

PR #274, opened 2023-03-17
Remove tolerance checks

IAlibay:
Pretty much in the name - tolerance checks are biting us more than they are being useful.

Ideally the tolerance should be built into the mapper (so if the mapper doesn't work and things break, it's on the mapper not the protocol).


openfe/protocols/openmm_rbfe/_rbfe_utils/topologyhelpers.py
```diff
@@ -374,8 +374,8 @@ def set_and_check_new_positions(mapping, old_topology, new_topology,
         Positions of the alchemically changing molecule in the "new" alchemical
         state.
     tolerance : float
-        Maximum allowed deviation along any dimension (x,y,z) in mapped atoms
-        between the "old" and "new" positions. Default 0.5.
+        Warning threshold for deviations along any dimension (x,y,z) in mapped
```
IAlibay, line 377: Keeping this in because it's a nice diagnostic, but am completely removing it as a user option.

richardjgowers (changes requested):
this can go in after the refactor in #267

IAlibay:
> this can go in after the refactor in #267

Sounds good, tbh I opened it so we can have that argument early - I remember there being several schools of thought on this one.

Plus I wanted to double check the behaviour because I'm pretty sure there's something a bit dodgy with it currently.

RiesBen (approved):
In favor of this :)

openfe/protocols/openmm_rbfe/_rbfe_utils/topologyhelpers.py
```diff
@@ -395,11 +395,12 @@ def set_and_check_new_positions(mapping, old_topology, new_topology,
...
     # loop through all mapped atoms and make sure we don't deviate by more than
-    # tolerance - not super necessary, but it's a nice sanity check that should
-    # eventually make it to a test
+    # tolerance - not super necessary, but it's a nice sanity check
     for key, val in mapping['old_to_new_atom_map'].items():
-        if np.any((new_pos_array[val] - old_pos_array[key]) > tolerance):
-            errmsg = f"mapping {key} : {val} deviates by more than {tolerance}"
-            raise ValueError(errmsg)
+        if np.any(
+            np.abs(new_pos_array[val] - old_pos_array[key]) > tolerance):
```
IAlibay, line 401: HAHA! I knew there was a bug!

IAlibay:
Depends on #267

openfe/protocols/openmm_rbfe/_rbfe_utils/topologyhelpers.py
richardjgowers: I think this is another thing I'd want to put under the "what do our safety rails do" category

richardjgowers (approved):

PR #281, opened 2023-03-28
Add periodicity setting tests

IAlibay:
Simple enough one, just a small sanity check to ensure that we are getting the right periodicity for the right set of settings.


RiesBen (approved):
cool :)

PR #285, opened 2023-04-03
Fix json issues with tests

IAlibay:
I'll admit it, I manually just edited the JSON, it was a lot faster.


IAlibay:
this is a deeper thing, too late in the day to work out why CI passes locally...

PR #286, opened 2023-04-03
Gufe fixup

richardjgowers:


IAlibay (approved):
We also have an openfe docs build issue that's also happening with the latest gufe change, I got partway there through fixes on the AFE PR (i.e. also use pydantic autodoc here), but there's still an off unit issue to resolve.

PR #287, opened 2023-04-03
Let's try to not import with eyes open

IAlibay:


IAlibay:
halves the conda build time, seems like it's a good move ahead!

IAlibay:
even better, quarters the build time!

richardjgowers (approved):

PR #288, opened 2023-04-03
Temporarily fix docs building

IAlibay:
Just testing the one thing re: mocks, otherwise we'll just mock everything and build docs that way.


IAlibay:
@mikemhenry I think this is good to go for a "minimal" fix (i.e. it'll still raise warnings that might be problematic for your PR, but at least docs build now). It's a good halfway to fixing things if you want to merge this.

mikemhenry (approved):

PR #294, opened 2023-04-04
simple duration printout

RiesBen:
This pr brings simple duration timings to the cmd line tools:
- plan_rbfe_network
- plan_rhfe_network
- quickrun

adds this line to the output: `Duration: 0:00:30.134822`
it is a super small code change, that adds a decorator to the cli tools.

### DIFF
--- a/openfecli/commands/plan_rbfe_network.py
+++ b/openfecli/commands/plan_rbfe_network.py
@@ -7,6 +7,7 @@
 """
 
 import click
+from datetime import datetime
 from typing import List
 from openfecli.utils import write
 from openfecli import OFECommandPlugin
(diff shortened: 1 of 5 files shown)


openfecli/commands/plan_rbfe_network.py
```diff
@@ -145,6 +148,9 @@ def plan_rbfe_network(
         folder_path=OUTPUT_DIR.get(output_dir),
     )
 
+    end_time = datetime.now()
```
IAlibay, line 151: May I suggest using a decorator for this?

RiesBen: problem is, thenyou can not align it anymore within the output. but maybe that is the way...

IAlibay: Don't we have a utils area?

RiesBen: I think I got an idea

richardjgowers (changes requested):
so currently `write` will print to stdout, but this will make stdout noisy if you're also using that for the actual output of the command

mikemhenry (changes requested):

richardjgowers (approved):
I opened openfreeenergy/gufe#168 which I think is a better system that @mikemhenry is alluding to.  This is good enough and a nice improvement for 0.7 though

PR #304, opened 2023-04-05
fix execute_DAG usage following gufe update

richardjgowers:


IAlibay (approved):
🎉

IAlibay:
Don't we need the changes to the protocol that passes ctx.shared to `_execute` too?

IAlibay:
> Don't we need the changes to the protocol that passes ctx.shared to `_execute` too?

Guess not, if CI works then we're good!

PR #307, opened 2023-04-05
add fetch-depth to CI version check

IAlibay:
Nothing big, doesn't look like we had universal wheels on here either.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


dwhswenson (approved):
Looks fine, but any particular reason to fetch the entire history? Is that needed?

IAlibay:
tbh it's partially lazyness on trying to figure out how deep you have to get to pickup enough for version to be right

IAlibay:
yeah 🤷🏽 I'm sure there's a happy medium, but it's also a 1s checkout so my incentive to try to work it out is pretty low

mikemhenry:
When I saw it was 1 second still, I wasn't too worried about it

PR #308, opened 2023-04-06
[WIP] Fix  single file installer 

mikemhenry:


.github/workflows/installer.yaml
```diff
@@ -2,96 +2,74 @@ name: Make single-file installers
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - "fix/installer"
```
mikemhenry, line 7: will remove before merge

devtools/installer/construct.yaml
mikemhenry, line 1: This is the file where we specify what we want in our installer

mikemhenry: Should I add `notebook` and `jupyterlab`?

openfe/tests/storage/test_resultclient.py
mikemhenry, line 1: This is from https://github.com/OpenFreeEnergy/openfe/pull/321 (needed for the same reason)

IAlibay (approved):
Makes sense from what I can see. I am taking a bit of a "if it works it works" approach on this one, since I'm not super versed on the single file installers (Richard echoed a similar take in today's call so please merge when ready @mikemhenry)

PR #310, opened 2023-04-06
rbfe -> rfe rename

### LINKED ISSUE #275: Rename rbfe to rfe
Everywhere we use the word "rbfe", technically it's just a plain "rfe", since sometimes we don't do "binding". It might be worth considering just renaming everything at some point?

IAlibay:
Fixes #275 

Whilst we're making breaking changes...

The proposal here is that RBFE is too specific when it's all pretty much "relative free energy" calcs (could be RHFE, could be RBFE, etc..). So RFE seems like a middle ground of not too specific but specific enough to make sense.


richardjgowers (commented):
@IAlibay Do you also want to think on the word Ligand?

IAlibay:
Yeah we can drop Ligand too

IAlibay:
there we go

IAlibay:
fun story, sed doesn't like certain git files so you need to use a mix of grep and sed

PR #317, opened 2023-04-06
Explanation docs (initial work and structure)

dwhswenson:
As discussed in the meeting today, this is a rough outline of the "explanation" docs (per diataxis). The approach I've been taking is to think of it as stripping down layers of abstraction, so we start out with the very abstract stuff and get more into the details as we go on.

Basically, the goal is "here's the simplest way to do it -- if that's too simple for your needs, here's the next simplest." These docs should also have lots of links to how-tos as we add them; the goal here is to start from the big picture.

The rest of the documentation will be stubs; not intended to be completed as part of this PR.


dwhswenson:
Outline/toctree is now as discussed in meeting today and as outlines in Google Doc (see slack), with the exception that I haven't combined API and CLI reference into a Reference section. That will be part of #325.

To-do list updated in main post.

richardjgowers (approved):

PR #319, opened 2023-04-07
Don't exclude py3.10 macos runners

IAlibay:
With OpenMM 8 out, this should be ok to do now.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,9 +29,6 @@ jobs:
         python-version:
           - "3.9"
           - "3.10"
-        exclude:
-          - os: 'macos'
-            python-version: '3.10'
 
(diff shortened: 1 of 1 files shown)


richardjgowers (changes requested):
macos is already the slowest runner, we don't need to test both py3.9 and py3.10

IAlibay:
> macos is already the slowest runner, we don't need to test both py3.9 and py3.10

did we fix things on arm64-osx? we could use the cirrus runners, they are blazing fast

IAlibay:
Should be good for re-review @richardjgowers - just bumped the runner to 3.10 is all.

Once merged we'll need to cycle the "required checks" in the repo settings.

richardjgowers (approved):

PR #323, opened 2023-04-07
[WIP] fix packaging

### LINKED ISSUE #322: Fix packaging [include tests + data]
Currently if you build openfe you won't get the majority of the tests or the data. What we want to do is release a package which users can test post-installation (as per the current instructions in our docs).

We'll need:
- Including test data in the manifest
- Adding all the necessary init files
- Adding a test that does a pure pip install against main and runs the tests

IAlibay:
Fixes #322 

recursive-includes are heavy handed and I hate them, I'd prefer do individual includes, but that's a lot more error prone I guess?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+recursive-include openfecli/tests/data/ *
+recursive-include openfe/tests/data/ *


mikemhenry:
It is more error-prone, but if we are testing this, then we should get an error when we try and run the tests missing some data files.

.github/workflows/ci.yaml
```diff
@@ -73,3 +73,26 @@ jobs:
...
+
+      - name: "Install"
+        run: python -m pip install .
+
+      - name: "Tests"
+        # Move far out of the openfe src directory
+        working-directory: ../../
+        run: |
+          pytest --pyargs openfe.tests
+          pytest --pyargs openfecli.tests
```
mikemhenry, line 98:
```suggestion
          pytest -v --pyargs openfe.tests
          pytest -v --pyargs openfecli.tests
```

I find pytest without `-v` a bit hard to read, or at least I prefer this version of the output.

mikemhenry (changes requested):

IAlibay:
> It is more error-prone, but if we are testing this, then we should get an error when we try and run the tests missing some data files. 

that's very true, I'm happier to stick with normal includes if folks don't mind!

openfe/tests/storage/conftest.py
```diff
@@ -1 +1,4 @@
-from gufe.tests.conftest import *
+from gufe.tests.conftest import (
+    absolute_transformation, complex_equilibrium,
```
IAlibay, line 2: Fun story, benzene_modifications is session scoped, and rdkit mols in gufe and smallmoleculecomponents in openfe.

Because of the way the discovery order works locally vs when doing it off an installed package, we weren't picking up issues related to that one fixture being overriden until now.

mikemhenry: amazing catch

IAlibay: would be if I could actually fix it :(

openfe/tests/storage/conftest.py
```diff
@@ -1 +1,102 @@
-from gufe.tests.conftest import *
+import pytest
```
IAlibay, line 1: This is not the cleanest way to do this, but was the easiest way I could work out to get us out scope clashes 🤷🏽

.github/workflows/ci.yaml
```diff
@@ -75,3 +75,39 @@ jobs:
           file: coverage.xml
           fail_ci_if_error: False
           verbose: True
+
+  package-tests:
```
mikemhenry, line 79: Thoughts on splitting this into it's own yaml? If the `package-tests` job only runs on schedule, then we can just make something like `package-tests.yaml`

IAlibay: Can do, I believe the other job also runs on schedule, but I do see the appeal of a cleaner "independent jobs per yaml" approach.

mikemhenry (approved):

PR #326, opened 2023-04-07
Speed up doc building

mikemhenry:


mikemhenry:
Docs build time on this pr: Build took 142 seconds 
Docs build time on main: Build took 938 seconds

dwhswenson (approved):
Changes LGTM, and can't imagine this will break CI (which is still running).

.github/workflows/ci.yaml
```diff
@@ -44,6 +44,8 @@ jobs:
         with:
           environment-file: environment.yml
           environment-name: openfe
+          cache-env: true
```
IAlibay, line 47: That's quite neat, might steal that for MDA

IAlibay (approved):

.github/workflows/ci.yaml
mikemhenry: dude it is amazing, I have it cache env and the downloads so that if there is an env change, at least the download cache will help.

mikemhenry: it is smart enough to be per-platform as well so it will cache mac, linux, and win builds separately, cache TTL is 1 day so it shound't get so stale it won't catch issues with env solving if upstream changes

mikemhenry:
Before we merge this I want to make sure `-n auto` doesn't slow things down

mikemhenry:
Cool, looks like on github we only get 2 threads anyway, so with `-n auto` if we ever run on a beefy runner we don't need to do anything.

PR #327, opened 2023-04-10
pull AlchemicalNetwork into top namespace

richardjgowers:


IAlibay (approved):
Are we happy with its presence in setup and setup.mapping?

richardjgowers:
@IAlibay yeah I think it's fine, this is one of those white lie imports to make everything import from a single source (rather than remembering gufe for X, openfe for Y etc)

IAlibay:
fair, tbh we can clean up our namespace later if it ever gets too much (it's not exactly a big API break if we're encouraging folks to import from the top level anyways)

PR #332, opened 2023-04-10
Swap LangevinSplittingDynamicMove to LangevinDynamicsMove

IAlibay:
I believe perses did this at some point last year: https://github.com/choderalab/perses/blob/main/perses/app/setup_relative_calculation.py

Performance-wise it yields:
- ~ 5-6x speed up in vacuum calcs
- ~ 50% performance increase in solvent calcs
- ~ couple % in complex calcs 

## Validation

To check that everything runs fine, I did a comparison of the benzenes RHFE set. When comparing DDGs between v0.6 and a hypothetical v0.8 with this integrator swap, we find all values to be within error of each other.

DDG

We do note however, that when comparing just the gas phase results, one of the edges has a > 1 kcal/mol deviation. This however turns out to not be linked to this integrator swap, but the change from HMR using 3.0 amu to 4.0 amu (results not shown, but effectively if I change the mass this deviation goes away). The error cancels out so it's fine here, but it's something we should be careful about (see #363) especially when we do AFEs.

DG-vacuum

(scales make it hard to see, but the red data point is what you're looking at above)


IAlibay:
The performance improvements here are substantial enough that we should get this merged earlier than later.

PR #333, opened 2023-04-10
Add pytest slow test filter

IAlibay:
Pretty much doing what's in the pytest docs + some fun actions trickery


.github/workflows/ci.yaml
```diff
@@ -63,8 +63,10 @@ jobs:
           micromamba list
 
       - name: "Run tests"
+        env:
+          slow: ${{ fromJSON('{"false":"","true":"--runslow"}')[github.event_name == 'schedule'] }}
```
mikemhenry, line 67: I love this

mikemhenry: If this works, I am going to steal it since this pattern comes up all the time

IAlibay: I'll admit it's not fully original - may have been inspired by github forums answers 😅 

That being said, if you like this, you'll love this: https://github.com/MDAnalysis/MDAKits/blob/main/.github/workflows/gh-ci-cron.yaml#L44

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -133,12 +133,11 @@ def test_dry_run_ligand(benzene_system, toluene_system,
         assert sampler.is_periodic
 
 
-@pytest.mark.parametrize('method', ['repex', 'sams', 'independent'])
 def test_dry_run_complex(benzene_complex_system, toluene_complex_system,
```
IAlibay, line 136: so adding this as a slow test does make CI go down to ~ couple of minutes (at least on Linux), however it does mean not having recorded coverage for the "protein" bits of the Protocol. It's ~ 1%, so I don't know if we care that much.

conftest.py
```diff
@@ -0,0 +1,24 @@
+# This code is part of OpenFE and is licensed under the MIT license.
```
IAlibay, line 1: This is ugly but the only way we can appropriately call pytest at the root level directory and have additional options picked up (otherwise we'd have to call pytest openfe.tests and pyest openfecli.tests separately).

Interestingly, this is the recommended approach by ptyest: https://docs.pytest.org/en/latest/how-to/writing_plugins.html#plugin-discovery-order-at-tool-startup :/ ... doesn't make it any prettier though

IAlibay: @mikemhenry does bring up the good point that this will probably not get packaged if it's in root, which would break the idea of doing this anyways. Shall we move it back to openfe.tests?

mikemhenry: I vote we move back to `openfe.tests` so they get packaged. Then we can tell users to run `pytest --pyargs openfecli.tests --pyargs openfe.tests -v -n auto` and that will test everything. Then if we add more burn in tests that we want people to run, they can add the `--runslow` arg

mikemhenry: And then on our main CI we just need to add `openfe/tests` to the pytest invocation (if we are not using the pyargs method)

IAlibay: pytest --pyargs invocation won't work with the flag it seems (this is a known issue), my suggestion is that we either a) turn the flag into `--fast` so that by default we already run all the tests, or b) switch to using env variables (boring, but at least it'll likely work?)

.github/workflows/ci.yaml
```diff
@@ -63,8 +63,10 @@ jobs:
           micromamba list
 
       - name: "Run tests"
+        env:
+          slow: ${{ fromJSON('{"false":"","true":"--runslow"}')[github.event_name != 'schedule'] }}
```
richardjgowers, line 67: can stuff like this get a comment?  it's not clear what's going on here to me

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
richardjgowers: pretty sure we can mark individual parameters as slow here, so you could keep repex as always on, then test the other two when slow tests are enabled.

IAlibay: tbh the other two cases (sams and independent) don't cover anything that's not covered in the vacuum and solvent tests, so I'd advocate for removing them and keeping repex.

My main question is whether we want to take the 4 mins bullet for this one repex test (which will give us coverage for protein paths).

richardjgowers: depends on your viewpoint of coverage.  internally we know these are doing the same thing which is your point, but externally it looks like two different codepaths so we probably ought to test both.  I'd probably wrap all three in a slow call?

IAlibay:
Ok - it's now got both the option of using a flag `--runslow`, for the lazy folks like me that don't like always having to set an env var on/off, and an env var (OFE_SLOW_TESTS) for when using `pytest --pyargs` invocations, or calling tests purely in the root directory (i.e. `pytest .` at the top level directory).

mikemhenry (approved):

PR #336, opened 2023-04-11
Fix protocol roundtrip

richardjgowers:


richardjgowers:
@IAlibay I ran into `pydantic.errors.ConfigError: field "FooSettings" not yet prepared so type is still a ForwardRef, you might need to call M.update_forward_refs().` which I think we can fix with removing the future import, but I don't have time to figure out all the implications of that

IAlibay:
> @IAlibay I ran into `pydantic.errors.ConfigError: field "FooSettings" not yet prepared so type is still a ForwardRef, you might need to call M.update_forward_refs().` which I think we can fix with removing the future import, but I don't have time to figure out all the implications of that

Ok makes sense, I see the issue's been targetting for v0.8!

PR #339, opened 2023-04-12
fix issue where we would get a key error if `OFE_SLOW_TESTS` was unset

mikemhenry:


openfe/tests/conftest.py
```diff
@@ -26,7 +26,7 @@ def pytest_configure(config):
 
 def pytest_collection_modifyitems(config, items):
     if (config.getoption("--runslow") or
-        os.environ['OFE_SLOW_TESTS'].lower() == 'true'):
+        os.getenv("OFE_SLOW_TESTS", default="false").lower()):
```
IAlibay, line 29: don't you still need to evaluate if it's equal to something? (i.e. it doesn't evaluate to a bool right?)

IAlibay, line 29:
```suggestion
        os.getenv("OFE_SLOW_TESTS", default="false").lower() == 'true'):
```

IAlibay (approved):

mikemhenry:
Yes! Silly copy and paste typo

PR #351, opened 2023-04-18
Use pip instead of conda for sphinx-toolbox

IAlibay:
Hopefully this will fix our sphinx versioning issues


IAlibay:
somethings wrong with the commit history here, cycling close->open

PR #354, opened 2023-04-20
Cleanup contexts upon calculation completion, failure

dotsdl:
This adds calls to `ContextCache.empty` at the end of repex execution, or in the case of execution failure, to avoid leaving behind stale contexts. This can be important for long-running processes executing multiple `ProtocolDAG`s in sequence.

### DIFF
--- a/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
+++ b/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
@@ -118,6 +118,8 @@ class HybridCompatibilityMixin(object):
                      max_iterations=minimization_steps)
             sampler_state_list.append(copy.deepcopy(sampler_state))
 
+        del compound_thermostate, sampler_state
+
         # making sure number of sampler states equals n_replicas
         if len(sampler_state_list) != n_replicas:
             # picking roughly evenly spaced sampler states
@@ -277,8 +279,11 @@ def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: stat
     sampler_state : openmmtools.states.SamplerState
         The posititions and accompanying state following minimization
     """
-    integrator = openmm.VerletIntegrator(1.0) #we won't take any steps, so use a simple integrator
-    context, integrator = cache.global_context_cache.get_context(
+    # we won't take any steps, so use a simple integrator
+    integrator = openmm.VerletIntegrator(1.0)
+    platform = openmm.Platform.getPlatformByName('CPU')
+    dummy_cache = cache.DummyContextCache(platform=platform)
+    context, integrator = dummy_cache.get_context(
         thermodynamic_state, integrator
     )
     try:
@@ -290,8 +295,4 @@ def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: stat
         )
         sampler_state.update_from_context(context)
     finally:
-        # cautiously clear out the global context cache too
-        for context in list(cache.global_context_cache._lru._data.keys()):
-            del cache.global_context_cache._lru._data[context]
-
-        del context, integrator
+        del context, integrator, dummy_cache
--- a/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
+++ b/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
@@ -287,10 +287,13 @@ def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: stat
     context, integrator = cache.global_context_cache.get_context(
         thermodynamic_state, integrator
     )
-    sampler_state.apply_to_context(
-        context, ignore_velocities=True
-    )
-    openmm.LocalEnergyMinimizer.minimize(
-        context, maxIterations=max_iterations
-    )
-    sampler_state.update_from_context(context)
+    try:
+        sampler_state.apply_to_context(
+            context, ignore_velocities=True
+        )
+        openmm.LocalEnergyMinimizer.minimize(
+            context, maxIterations=max_iterations
+        )
+        sampler_state.update_from_context(context)
+    finally:
+        del context, integrator
--- a/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
+++ b/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
@@ -34,7 +34,7 @@ class HybridCompatibilityMixin(object):
         self._hybrid_factory = hybrid_factory
         super(HybridCompatibilityMixin, self).__init__(*args, **kwargs)
 
-    def setup(self, reporter, platform, lambda_protocol,
+    def setup(self, reporter, lambda_protocol,
               temperature=298.15 * unit.kelvin, n_replicas=None,
               endstates=True, minimization_steps=100):
         """
@@ -45,8 +45,6 @@ class HybridCompatibilityMixin(object):
         ----------
         reporter : OpenMM reporter
             Simulation reporter to attach to each simulation replica.
-        platform : openmm.Platform
-            Platform to perform simulation on.
         lambda_protocol : LambdaProtocol
             The lambda protocol to be used for simulation. Default to a default
             class creation of LambdaProtocol.
@@ -84,8 +82,6 @@ class HybridCompatibilityMixin(object):
         thermodynamic_state_list = []
         sampler_state_list = []
 
-        context_cache = cache.ContextCache(platform)
-
         if n_replicas is None:
             msg = (f"setting number of replicas to number of states: {n_states}")
             warnings.warn(msg)
@@ -118,8 +114,6 @@ class HybridCompatibilityMixin(object):
 
             # now generating a sampler_state for each thermodyanmic state,
             # with relaxed positions
-            context, context_integrator = context_cache.get_context(
-                                             compound_thermostate_copy)
             minimize(compound_thermostate_copy, sampler_state,
                      max_iterations=minimization_steps)
             sampler_state_list.append(copy.deepcopy(sampler_state))
@@ -296,4 +290,8 @@ def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: stat
         )
         sampler_state.update_from_context(context)
     finally:
+        # cautiously clear out the global context cache too
+        for context in list(cache.global_context_cache._lru._data.keys()):
+            del cache.global_context_cache._lru._data[context]
+
         del context, integrator
--- a/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
+++ b/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
@@ -36,7 +36,8 @@ class HybridCompatibilityMixin(object):
 
     def setup(self, reporter, lambda_protocol,
               temperature=298.15 * unit.kelvin, n_replicas=None,
-              endstates=True, minimization_steps=100):
+              endstates=True, minimization_steps=100,
+              minimization_platform="CPU"):
         """
         Setup MultistateSampler based on the input lambda protocol and number
         of replicas.
@@ -58,7 +59,9 @@ class HybridCompatibilityMixin(object):
             Whether or not to generate unsampled endstates (i.e. dispersion
             correction).
         minimization_steps : int
-            Number of steps to minimize states.
+            Number of steps to pre-minimize states.
+        minimization_platform : str
+            Platform to do the initial pre-minimization with.
 
         Attributes
         ----------
@@ -114,8 +117,10 @@ class HybridCompatibilityMixin(object):
 
             # now generating a sampler_state for each thermodyanmic state,
             # with relaxed positions
+            # Note: remove once  choderalab/openmmtools#672 is completed
             minimize(compound_thermostate_copy, sampler_state,
-                     max_iterations=minimization_steps)
+                     max_iterations=minimization_steps,
+                     platform_name=minimization_platform)
             sampler_state_list.append(copy.deepcopy(sampler_state))
 
         del compound_thermostate, sampler_state
@@ -257,8 +262,10 @@ def create_endstates(first_thermostate, last_thermostate):
     return unsampled_endstates
 
 
-def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: states.SamplerState,
-             max_iterations: int=100) -> states.SamplerState:
+def minimize(thermodynamic_state: states.ThermodynamicState,
+             sampler_state: states.SamplerState,
+             max_iterations: int=100,
+             platform_name: str="CPU") -> states.SamplerState:
     """
     Adapted from perses.dispersed.feptasks.minimize
 
@@ -273,6 +280,8 @@ def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: stat
         The starting state at which to minimize the system.
     max_iterations : int, optional, default 100
         The maximum number of minimization steps. Default is 100.
+    platform_name : str
+        The OpenMM platform name to carry out the minimization with.
 
     Returns
     -------
@@ -281,7 +290,7 @@ def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: stat
     """
     # we won't take any steps, so use a simple integrator
     integrator = openmm.VerletIntegrator(1.0)
-    platform = openmm.Platform.getPlatformByName('CPU')
+    platform = openmm.Platform.getPlatformByName(platform_name)
     dummy_cache = cache.DummyContextCache(platform=platform)
     context, integrator = dummy_cache.get_context(
         thermodynamic_state, integrator
--- a/openfe/protocols/openmm_rfe/equil_rfe_methods.py
+++ b/openfe/protocols/openmm_rfe/equil_rfe_methods.py
@@ -81,7 +81,13 @@ class RelativeHybridTopologyProtocolResult(gufe.ProtocolResult):
           a Quantity defined with units.
         """
         # TODO: Check this holds up completely for SAMS.
-        dGs = [pus[0].outputs['unit_estimate'] for pus in self.data.values()]
+        # this v
+        dGs = []
+        for pus in self.data.values():
+            dGs.extend([pu.outputs['unit_estimate'] for pu in pus])
+
+        # not this v
+        #dGs = [pus[0].outputs['unit_estimate'] for pus in self.data.values()]
         u = dGs[0].u
         # convert all values to units of the first value, then take average of magnitude
         # this would avoid a screwy case where each value was in different units
@@ -91,7 +97,13 @@ class RelativeHybridTopologyProtocolResult(gufe.ProtocolResult):
 
     def get_uncertainty(self):
         """The uncertainty/error in the dG value: The std of the estimates of each independent repeat"""
-        dGs = [pus[0].outputs['unit_estimate'] for pus in self.data.values()]
+        # this v
+        dGs = []
+        for pus in self.data.values():
+            dGs.extend([pu.outputs['unit_estimate'] for pu in pus])
+
+        # not this v
+        #dGs = [pus[0].outputs['unit_estimate'] for pus in self.data.values()]
         u = dGs[0].u
         # convert all values to units of the first value, then take average of magnitude
         # this would avoid a screwy case where each value was in different units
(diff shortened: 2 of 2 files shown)


dotsdl:
Worked with @jchodera earlier today on this; he suggested this change due to behavior we were seeing in `alchemiscale` compute services on Lilac. This problem may be mitigated by relatively short (24hr) jobs on Lilac, but would become a problem for services running indefinitely (such as on PRP).

I'm seeing behavior on my local test box that suggests contexts are not being cleaned up currently; I have yet to test if this PR addresses this.

dotsdl:
Can confirm that we observe GPU memory exhaustion when running many RBFE `ProtocolDAG`s in sequence:

```
[2023-04-20 06:16:48,756] [vulkan-cc574698-44f9-4d9f-adbd-a9055d83d534] [INFO] '<ProtocolDAG-8c622c4dca6e677be62fdb5199cbb69e>' : '<ProtocolDAGResult-a060faa7540c0d29a7bceff2e1a5f9c9>' : FAILURE : 'ProtocolUnitFailure(lig_ejm_31 lig_ejm_42 repeat 0 generation 0)' : ('OpenMMException', ('Error creating array posDelta: CUDA_ERROR_OUT_OF_MEMORY (2)',))
```

Also end up getting many of these:

```
[2023-04-20 06:19:33,720] [vulkan-1611dd3a-cd5e-4c9f-9661-0930074ec2a1] [INFO] '<ProtocolDAG-623250040607a0675771ae4b3ce4405e>' : '<ProtocolDAGResult-b91a3c75dd6eb5bb56f351f886e44f04>' : FAILURE : 'ProtocolUnitFailure(lig_ejm_42 lig_ejm_43 repeat 0 generation 0)' : ('OpenMMException', ('No compatible CUDA device is available',))
```

dotsdl:
Running a test now on my local box to see if the changes in this PR give stable memory use over time. Will report back on observations.

dotsdl:
This may not yet be solving the issue, but can't conclude yet; will let the test box run overnight and see what state it's in in the morning.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -587,107 +591,110 @@ def run(self, *, dry=False, verbose=True,
             capacity=None, time_to_live=None, platform=platform,
         )
```
mikemhenry, line 593:
```suggestion
# Hold onto your butts, big try block
```
/s

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -587,107 +591,110 @@ def run(self, *, dry=False, verbose=True,
...
-            )
-        elif sampler_settings.sampler_method.lower() == "sams":
-            sampler = _rfe_utils.multistate.HybridSAMSSampler(
-                mcmc_moves=integrator,
-                hybrid_factory=hybrid_factory,
-                online_analysis_interval=sampler_settings.online_analysis_interval,
-                online_analysis_minimum_iterations=sampler_settings.online_analysis_minimum_iterations,
-                flatness_criteria=sampler_settings.flatness_criteria,
-                gamma0=sampler_settings.gamma0,
+        try:
```
IAlibay, line 594: Does the try really have to be all the way up at the integrator generation stage? We shouldn't be failing here ever. Why not just over the if/else on dry run?

IAlibay (changes requested):
Can you move the try except down to only encapsulate the problematic section?

IAlibay:
This is rather odd, I've ran multi-day long DAGs before and never saw that behaviour.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -587,107 +591,110 @@ def run(self, *, dry=False, verbose=True,
...
-            nc = shared_basepath / sim_settings.output_filename
-            chk = shared_basepath / sim_settings.checkpoint_storage
+                # clean up the reporter file
+                fns = [shared_basepath / sim_settings.output_filename,
+                       shared_basepath / sim_settings.checkpoint_storage]
+                for fn in fns:
+                    os.remove(fn)
+        finally:
+            energy_context_cache.empty()
+            sampler_context_cache.empty()
```
richardjgowers, line 689: this feels like it should get (eventually) upstreamed as these context objects are used using a context manager pattern

richardjgowers:
@IAlibay re: your multiple long DAGs, were these in different Python evocations?  I think this might be surfacing as dotson has a single long running process issuing these (which shouldn't be bad), whereas if you were quickrunning in succession you'd have been tearing down the memory footprint between each

IAlibay:
> @IAlibay re: your multiple long DAGs, were these in different Python evocations?  I think this might be surfacing as dotson has a single long running process issuing these (which shouldn't be bad), whereas if you were quickrunning in succession you'd have been tearing down the memory footprint between each

No, recently it was the one big Python loop, so should all be the same process. I'd say the most I tried is 12 tyk2 complex calculations in a loop, so if it's a large scale thing it might explain why we didn't see it.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
IAlibay: I think in the short term solution, it might be good to at least make sure that the context is wiped in multistatesampler.__del__

dotsdl:
Running another testing round on local box with latest changes from power hour discussion. Will report back behavior.

dotsdl:
These changes still appear insufficient to avoid growing system and GPU memory use. As we saw from discussion we observe the problem after ~40 successful runs.

dotsdl:
Addressing your comments now @IAlibay.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
dotsdl: Indeed, upstreaming here: https://github.com/choderalab/openmmtools/pull/690

dotsdl:
Running another round of local tests with latest changes, including what we propose to upstream in https://github.com/choderalab/openmmtools/pull/690.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -637,57 +637,71 @@ def run(self, *, dry=False, verbose=True,
...
+            #sampler_context_cache.empty()
+            # TODO: remove once upstream solution in place: https://github.com/choderalab/openmmtools/pull/690
+            # replace with above
+            for context in list(energy_context_cache._lru._data.keys()):
+                del energy_context_cache._lru._data[context]
+            for context in list(sampler_context_cache._lru._data.keys()):
+                del sampler_context_cache._lru._data[context]
+
+            del sampler_context_cache, energy_context_cache
+            del integrator, sampler
```
IAlibay, line 696: We can't do this here - if we delete the sampler we can't return it on line 691

IAlibay:
@dotsdl - latest changes seem to be preventing memory leaks on my test loop, if you can test it on your end that'd be great.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -635,59 +625,87 @@ def run(self, *, dry=False, verbose=True,
...
+                del energy_context_cache._lru._data[context]
+            for context in list(sampler_context_cache._lru._data.keys()):
+                del sampler_context_cache._lru._data[context]
+            # cautiously clear out the global context cache too
+            for context in list(
+                    openmmtools.cache.global_context_cache._lru._data.keys()):
+                del cache.global_context_cache._lru._data[context]
+
+            del sampler_context_cache, energy_context_cache
+            if not dry:
```
IAlibay, line 699: won't delete the integrator on dry, one would hope that alchemiscale style services wouldn't be running in dry mode anyways?

dotsdl: Yeah, I don't see a scenario where we would use `dry` on `alchemiscale`, so think we're good here.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
dotsdl: Ah good catch!

dotsdl:
@IAlibay haha sometimes it's a good strategy. :grin:

IAlibay:
> @IAlibay haha sometimes it's a good strategy. :grin: 

😅 you mean to say there are other strategies?

In any case, I'll happily merge this & cut a release once we know it's stable.

dotsdl:
I think that was it @IAlibay! Great work! I am no longer seeing rising GPU memory usage, but still observe rising system memory usage. Do you observe the same?

If you don't also observe rising system memory, this may be a symptom of the `SynchronousComputeService`, not this protocol.

IAlibay:
> I think that was it @IAlibay! Great work! I am no longer seeing rising GPU memory usage, but still observe rising system memory usage. Do you observe the same?
> 
> If you don't also observe rising system memory, this may be a symptom of the `SynchronousComputeService`, not this protocol.

Awesome that's one thing gotten. Yeah the host memory leak tracks with what I'm seeing too :/ 

Off the top of my head it's possibly HTF or the generator objects.

IAlibay:
@dotsdl - so with these latest changes I do see host memory creep up, but it plateaus ~ 9 GB in total system memory for three concurrent dag execution loops on my workstation (with a gc.collect() call at the end of each loop). It's hard to know if this is an actual memory leak or just garbage collection being quite lazy when there's lots of RAM left (I still have another 23 GB of RAM free..). For now I'm going to leave it as-is since it seems stable and honestly if we manage to get 12 iterations of a complex calculation in 24h that would be rather amazing performance.

Given it's so late in the day I'm going to just hold off on merging this whilst I quickly do a couple of actual simulations overnight. Just want to check we didn't drastically break stuff somewhere here. If everything looks fine I'll merge / release tomorrow in the US morning.

openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, *args, hybrid_factory=None, **kwargs):
         self._hybrid_factory = hybrid_factory
         super(HybridCompatibilityMixin, self).__init__(*args, **kwargs)
 
-    def setup(self, reporter, platform, lambda_protocol,
+    def setup(self, reporter, lambda_protocol,
```
IAlibay, line 37: creating a context here was absolutely unecessary, so it's been removed

openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
```diff
@@ -283,14 +279,20 @@ def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: stat
     sampler_state : openmmtools.states.SamplerState
         The posititions and accompanying state following minimization
     """
-    integrator = openmm.VerletIntegrator(1.0) #we won't take any steps, so use a simple integrator
-    context, integrator = cache.global_context_cache.get_context(
+    # we won't take any steps, so use a simple integrator
+    integrator = openmm.VerletIntegrator(1.0)
+    platform = openmm.Platform.getPlatformByName('CPU')
+    dummy_cache = cache.DummyContextCache(platform=platform)
```
IAlibay, line 285: using a dummy context cache for the convenience of being able to pass a thermodynamic state object rather than doubling up efforts and doing the 3-4 extra lines here ourselves

openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
```diff
@@ -283,14 +279,20 @@ def minimize(thermodynamic_state: states.ThermodynamicState, sampler_state: stat
...
+    try:
+        sampler_state.apply_to_context(
+            context, ignore_velocities=True
+        )
+        openmm.LocalEnergyMinimizer.minimize(
+            context, maxIterations=max_iterations
+        )
+        sampler_state.update_from_context(context)
+    finally:
+        del context, integrator, dummy_cache
```
IAlibay, line 298: Being a bit overly cautious but it'll do I think.

openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
```diff
@@ -118,12 +114,12 @@ class creation of LambdaProtocol.
 
             # now generating a sampler_state for each thermodyanmic state,
             # with relaxed positions
-            context, context_integrator = context_cache.get_context(
-                                             compound_thermostate_copy)
             minimize(compound_thermostate_copy, sampler_state,
                      max_iterations=minimization_steps)
             sampler_state_list.append(copy.deepcopy(sampler_state))
 
+        del compound_thermostate, sampler_state
```
IAlibay, line 121: I _think_ that should be fine.

IAlibay (approved):

dotsdl:
All good @IAlibay! Will await your conclusions and `openfe` 0.7.3 on the `alchemiscale` end. Let me know if there's anything you'd like me to hit as well.

dotsdl:
Since we're now using the CPU platform for minimization, I need to set `OPENMM_CPU_THREADS=1` in typical usage with stacking `alchemiscale` `SynchronousComputeService`s on a single host. This is fine, but is minimization generally faster to do on 1 CPU core or on the GPU, accounting for amortization of moving things on and off the GPU? Not critical to address now, but wanted to ask.

dotsdl:
Running another burn test on my host here; will report back observations as well.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
jchodera: Also cc https://github.com/choderalab/openmmtools/issues/688

jchodera:
> Since we're now using the CPU platform for minimization, I need to set OPENMM_CPU_THREADS=1 in typical usage with stacking alchemiscale SynchronousComputeServices on a single host. This is fine, but is minimization generally faster to do on 1 CPU core or on the GPU, accounting for amortization of moving things on and off the GPU? Not critical to address now, but wanted to ask.

Minimization is extremely slow on the CPU. I highly recommend doing this on the GPU. 

Some considerations for minimization of replicas:
* If you create/destroy `Context` objects for each minimization, it might incur overhead of a second or so per replica.  
* If you use a `ContextCache` to manage minimization of replicas, the `Context`s may not be reused since the `ContextCache` requires both `System` and `Integrator` to match to re-use the `Context`. That will just tie up a bunch or resources you don't use again if the `Integrator` doesn't match (e.g. minimization is done with `VerletIntegrator` but propagation uses `LangevinMiddleIntegrator`).
* The best thing to do here might be to create a `ContextCache` just for minimization of all replicas, minimize all replicas, and then clean up that cache, being careful that all `Context`s are deleted.

dotsdl:
Yeah I do observe an idle GPU on my test host for some time while minimizations are happening, which may be a bit of a bottleneck. However for production-length runs, this may not end up mattering much, since the minimization may still be very short compared to production runtime.

If it's possible to take @jchodera's suggestions above and run the minimizations on the GPU, that may still be preferable though.

The good news is that so far running 6 workers with a mix of complex and solvent transformation `ProtocolDAG`s going from tyk2, I see each worker topping out at about 2.45GB, or 14.7GB total memory usage. This suggests that the current state of this protocol may now be memory stable!

IAlibay:
> Minimization is extremely slow on the CPU. I highly recommend doing this on the GPU.

@jchodera - For context here, this is only a 100 step pre-minimization to get out of potentially really bad high energy states. The reason we're doing this here is that we were seeing cases of silent overflows when minimizing on GPUs.

Admittedly I think this might have all been tied into the FIRE minimizer issues we recently reported, however our take was that being a bit extra cautious for now might only cost us an extra couple of minutes? Unless you think this might be unreasonably slow for larger systems (TYK2 seems to be fine, but it's not exactly the largest system)?

> Yeah I do observe an idle GPU on my test host for some time while minimizations are happening, which may be a bit of a bottleneck. 

I note that the "proper" minimization (i.e. https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/protocols/openmm_rfe/equil_rfe_methods.py#L652)  on  should be assigned the GPU platform @dotsdl (or at least whatever platform was chosen by default in settings).

> The best thing to do here might be to create a ContextCache just for minimization of all replicas, minimize all replicas, and then clean up that cache, being careful that all Contexts are deleted.

That makes sense thanks! For this PR (just so we can quickly deploy on alchemiscale), I'm going to say let's take the extra couple of seconds performance hit, and then we can follow-up with a fix that does exactly as @jchodera suggests? If anything the performance loss of building / creating contexts is miniscule given we're using a wildly inneficient integrator (see #332 ).

dotsdl:
Observing a slow climb in host memory usage, but I still think this looks tolerable if it scales with task count, not task length, given how artificially short the test tasks are. I ran 240 tasks, 120 each of complex and solvent `Transformation`s. Of the 6 processes, here is the host memory usage with only a handful of tasks remaining:

```
2.79G
2.66G
2.53G
2.45G
2.43G
2.29G
```
This after running for nearly 4 hours, and an almost unnoticeable climb in memory usage from that observed over 2 hours ago.

dotsdl:
> That makes sense thanks! For this PR (just so we can quickly deploy on alchemiscale), I'm going to say let's take the extra couple of seconds performance hit, and then we can follow-up with a fix that does exactly as @jchodera suggests? If anything the performance loss of building / creating contexts is miniscule given we're using a wildly inneficient integrator (see #332 ).

Yes, I think we're clear to proceed with this as-is. Thank you for all the solid sleuthing on this @IAlibay! I think in practice we'll get plenty of throughput out of the implementation as it stands, and we can work on greater efficiency when we're not under the current time pressure.

IAlibay:
Aha! I didn't know OpenMM's localenergyminimizer was this smart: https://github.com/openmm/openmm/blob/master/openmmapi/src/LocalEnergyMinimizer.cpp#L54

Seems like it'll always do a CPU fallback if there's a force clipping issue. In that case let's keep the old behaviour here and just stick to whatever the default platform is for this initial minimization. Once https://github.com/choderalab/openmmtools/pull/672 is merged, we can probably just get rid of this pre-minimization completely.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -635,59 +625,91 @@ def run(self, *, dry=False, verbose=True,
...
+                       shared_basepath / sim_settings.checkpoint_storage]
+                for fn in fns:
+                    os.remove(fn)
+        finally:
+            # close reporter when you're done, prevent file handle clashes
             reporter.close()
+            del reporter
+
+            # clean up the analyzer
+            if not dry:
```
IAlibay, line 686: This needs a bit of a cleanup, but I'll do it in the upcoming refactor.

IAlibay:
Quick test update - we're getting reproducible results for TYK2's ejm_31_ejm_46 edge - within 0.1 kcal/mol. Everything seems fine.

IAlibay:
Alright, thanks for all the work here @dotsdl - let's move towards a release 🎉

jchodera:
>@jchodera - For context here, this is only a 100 step pre-minimization to get out of potentially really bad high energy states. The reason we're doing this here is that we were seeing cases of silent overflows when minimizing on GPUs.
> Admittedly I think this might have all been tied into the FIRE minimizer issues we recently reported, however our take was that being a bit extra cautious for now might only cost us an extra couple of minutes? Unless you think this might be unreasonably slow for larger systems (TYK2 seems to be fine, but it's not exactly the largest system)?

@IAlibay : There are definitely deficiencies with the FIRE minimizer, and we're hoping to be able to address those at some point in the near future. But PME systems will get _very_ slow on the CPU, even for 100 steps. 

Can you elaborate on the kinds of "silent overflows" you ran into? This was with `mixed` precision? I wonder if a simpler gradient descent minimizer that rejects steps that go uphill would be sufficient for your purposes but could still run on the GPU.

jchodera:
> That makes sense thanks! For this PR (just so we can quickly deploy on alchemiscale), I'm going to say let's take the extra couple of seconds performance hit, and then we can follow-up with a fix that does exactly as @jchodera suggests? If anything the performance loss of building / creating contexts is miniscule given we're using a wildly inneficient integrator (see https://github.com/OpenFreeEnergy/openfe/pull/332 ).

This is a reasonable solution! The main concern here was ensuring there were no device resource leaks, not performance.

IAlibay:
> There are definitely deficiencies with the FIRE minimizer, and we're hoping to be able to address those at some point in the near future. But PME systems will get very slow on the CPU, even for 100 steps.

I got muddled a bit here and I had completely forgotten for this PR - it turns out that localenergyminizer will do a default fallback to CPU if the forces overflow, so I ended up sticking with the old behaviour and just passing whatever the user defined platform is.

> Can you elaborate on the kinds of "silent overflows" you ran into? This was with mixed precision? I wonder if a simpler gradient descent minimizer that rejects steps that go uphill would be sufficient for your purposes but could still run on the GPU.

@jchodera: If I remember correctly the issue here was that the minimizer would essentially attempt to minimize but fail to do so. However rather than throwing an exception or directly giving a NaN energy, it would just take the system to an unreasonable energy state, from that point any further attempts to minimize would lead to a system that yields NaNs. I didn't manage to dig into this too much at the time, but avoiding the default `MultiStateSampler.minimize` FIRE -> local minimizer scheme by doing the pre-minimization seperately with `LocalEnergyMinimizer` first worked fine.

I suspect the issue with the FIRE minimizer is that, similar to LocalEnergyMinimizer on non-CPU platforms, the forces overflow but not the energy accumulator leading to this weird behaviour.

(I wish I had taken much better notes when I encountered it, I'll try to reproduce this again once things a bit less hectic).

PR #355, opened 2023-04-20
Fixes ProtocolResult aggregation for get_estimate, get_uncertainty

dotsdl:
When gathering multiple independent `ProtocolDAGResult`s, we need to gather up estimates from all the `ProtocolUnitResult`s under each repeat index, not just the first one.


richardjgowers (changes requested):
I think the real fix here is to not make repeat ids sequential from 0 in `_create`, so that they don't clash in gather (see #356).

The changes here would (eventually) mean we're not treating generations correctly (they would require a concatenation of the timeseries)

IAlibay:
@dotsdl I'm going to close this PR since we went with #356

PR #356, opened 2023-04-20
openmm_rfe: make repeat_ids non 0 assuming

richardjgowers:
currently _create() assumes there will be a single DAG in flight for a given Transformation.  So repeats are numbered from 0 sequentially.  This breaks if you issue multiple DAGS from the same Transformation, as there will be multiple repeat_id=0 (each from different DAGs).  Instead, number repeat IDs randomly (since the 0 based numbering was arbitrary anyway)

re: #355


IAlibay (approved):
Yeah ok, if that id ever gets used for anything but internal bookkeeping, we'll need a cleaner solution for users.

PR #358, opened 2023-04-21
Fix test incompatibility

dwhswenson:
I think this fixes the weird incompatibility we were seeing between something in the protocol tests and the storage tests.

I still don't understand why it worked when we ran based on files, instead of on the installation (unless maybe the order runs differently?) but it is clear that modifying a `GufeTokenizable` after it has been created would cause problems for the round-trip tests.

Also, I finally found an excuse to use the `copy_with_replacements` method I've claimed will be so useful!

Currently running a final local check to ensure that this fixes; will request review when I can confirm.


IAlibay (approved):
lgtm, thanks!

dwhswenson:
Yep, this fixes!

`pytest --pyargs openfe.tests.protocols --pyargs openfe.tests.storage` fails on `main`, passes in this branch.

Issue #342, opened 2023-04-12
Weird failures when running tests in docker image


mikemhenry:
Not sure why, but these tests pass in our CI but fail when ran inside a docker image, all other tests pass (including `gufe` tests). 

As a workaround, we just skip these when running tests inside a docker image.

```
=================================== FAILURES ===================================
_ TestResultClient.test_store_load_network_same_process[benzene_variants_star_map] _

self = <openfe.tests.storage.test_resultclient.TestResultClient object at 0x7f72af55a1f0>
request = <FixtureRequest for <Function test_store_load_network_same_process[benzene_variants_star_map]>>
fixture = 'benzene_variants_star_map'

    @pytest.mark.skipif(os.getenv("OFE_DOCKER_BUILD", default="false").upper() == "TRUE", reason="Test fails in docker image, see issue #xxx")
    @pytest.mark.parametrize("fixture", ["benzene_variants_star_map"])
    def test_store_load_network_same_process(self, request, fixture):
        network = request.getfixturevalue(fixture)
>       self._test_store_load_same_process(network,
                                           "store_network",
                                           "load_network")

/opt/conda/lib/python3.9/site-packages/openfe/tests/storage/test_resultclient.py:206: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/conda/lib/python3.9/site-packages/openfe/tests/storage/test_resultclient.py:158: in _test_store_load_same_process
    reloaded = load_func(obj.key)
/opt/conda/lib/python3.9/site-packages/openfe/storage/resultclient.py:237: in load_network
    return self._load_gufe_tokenizable("setup", key)
/opt/conda/lib/python3.9/site-packages/openfe/storage/resultclient.py:207: in _load_gufe_tokenizable
    return recursive_build_object_cache(gufe_key)
/opt/conda/lib/python3.9/site-packages/openfe/storage/resultclient.py:192: in recursive_build_object_cache
    recursive_build_object_cache(key)
/opt/conda/lib/python3.9/site-packages/openfe/storage/resultclient.py:192: in recursive_build_object_cache
    recursive_build_object_cache(key)
/opt/conda/lib/python3.9/site-packages/openfe/storage/resultclient.py:202: in recursive_build_object_cache
    obj = key_decode_dependencies(dct, registry)
[...]

PR #359, opened 2023-04-21
Remove test skipping

mikemhenry:
Liked we talked about, this should make skipping the tests that fail when called with `--pyargs` the default, but our CI "opts in" to running the tests so they are still tested, but when end-users run the tests, everything passes since the tests will be skipped.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -66,6 +66,8 @@ jobs:
         env:
           # Set the OFE_SLOW_TESTS to True if running a Cron job
           OFE_SLOW_TESTS: ${{ fromJSON('{"false":"false","true":"true"}')[github.event_name == 'schedule'] }}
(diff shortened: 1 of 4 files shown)


dwhswenson:
I think I fixed the problem with tests failing on `--pyargs` in #358?

IAlibay:
😅 yeah I think @dwhswenson was too fast sorry @mikemhenry !

dwhswenson:
Could convert this PR into an opportunity to remove the `OFE_DOCKER_BUILD` / `SKIP_SOME_TESTS` env var, though?

mikemhenry:
Haha no worries! And yes I can remove those args now!

IAlibay (approved):

IAlibay:
we might have further issues (see #342 but at those specific tests seem to be running fine now)

PR #364, opened 2023-04-25
lomap mapper updates for 2.3.0

richardjgowers:
force kwarg usage for LomapAtomMapper (avoids kwargs accidentally getting set in arg usage)

add seed and shift kwargs (added in Lomap 2.2.0 and 2.3.0 respectively)

### DIFF
--- a/openfe/setup/atom_mapping/lomap_mapper.py
+++ b/openfe/setup/atom_mapping/lomap_mapper.py
@@ -14,9 +14,13 @@ class LomapAtomMapper(LigandAtomMapper):
     time: int
     threed: bool
     max3d: float
+    element_change: bool
+    seed: str
+    shift: bool
 
-    def __init__(self, time: int = 20, threed: bool = True,
-                 max3d: float = 1000.0, element_change: bool = True):
+    def __init__(self, *, time: int = 20, threed: bool = True,
+                 max3d: float = 1000.0, element_change: bool = True,
+                 seed: str = '', shift: bool = True):
         """Wraps the MCS atom mapper from Lomap.
 
         Kwargs are passed directly to the MCS class from Lomap for each mapping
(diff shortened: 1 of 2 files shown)


openfe/setup/atom_mapping/lomap_mapper.py
```diff
@@ -14,9 +14,13 @@ class LomapAtomMapper(LigandAtomMapper):
...
     max3d: float
+    element_change: bool
+    seed: str
+    shift: bool
 
-    def __init__(self, time: int = 20, threed: bool = True,
-                 max3d: float = 1000.0, element_change: bool = True):
+    def __init__(self, *, time: int = 20, threed: bool = True,
+                 max3d: float = 1000.0, element_change: bool = True,
+                 seed: str = '', shift: bool = False):
```
richardjgowers, line 23: this `shift=False` will give different behaviour to before, but might be better suited to our current available methods?

RiesBen (approved):
lgtm, keeps shifting as default :)

RiesBen (approved):
lgtm, keeps shifting as default :)

openfe/setup/atom_mapping/lomap_mapper.py
```diff
@@ -14,9 +14,13 @@ class LomapAtomMapper(LigandAtomMapper):
     time: int
     threed: bool
     max3d: float
+    element_change: bool
+    seed: str
```
IAlibay, line 18: In most cases the likelihood is that the mapper will be used in sucessing on multiple similar molecules. Could it be possible to add an option to populate and cache the seed into the mapper?

richardjgowers: you could probably make the Mapper slowly learn the seed, but it wouldn't be a good way to do this (the learning rate would depend a lot on the order you used the mapper).  I'm going to say out of scope?

IAlibay: Sure (it's why I didn't block over it 🙃 ), that being said it would be a good one to at least document somewhere? Possibly in ExampleNotebooks

PR #371, opened 2023-04-27
update package versions

mikemhenry:
Will merge this in after I upload the new installers


mikemhenry:
I need to make the action add the package to the release so I don't need to download and upload it manually

mikemhenry:
I've attached the installers to the release so this is good now to merge.

IAlibay:
we should consider if a good old `sed` might help automate this task

PR #373, opened 2023-04-28
`test` CLI command

dwhswenson:
Just a little start on this. Seems to work for me locally.

A couple of questions:

1. Which "section" of the CLI should this be listed in, if at all? Currently I have it hidden, so it doesn't show in `openfe -h` (it will show its help if you type `openfe test -h`, though). NB: we can also change up the sections we have listed, and I think that's probably a good idea anyway.

2. Do we want to add options to pass through some pytest options, like `-k`? That would make it possible to test, too. Otherwise I'm afraid we'd get into testception.

### DIFF
--- a/openfecli/commands/test.py
+++ b/openfecli/commands/test.py
@@ -1,0 +1,40 @@
+import click
+from openfecli import OFECommandPlugin
+
+import pytest
+import os
+
+from openfecli.utils import write
+
+@click.command(
+    "test",
+    short_help="Run the OpenFE test suite"
+)
+@click.option('--long', is_flag=True, default=False,
+              help="Run additional tests (takes much longer)")
+def test(long):
+    """
+    Run the OpenFE test suite. This first checks that OpenFE is correctly
+    imported, and then runs the main test suite, which should take several
+    minutes. If given the ``--long`` flag, this will include several tests
+    that take significantly longer, but ensure that we're able to fully run
+    in your environment.
+    """
+    default = ["-v"]
+    pyargs = ["--pyargs", "openfe", "--pyargs", "openfecli"]
+    old_env = dict(os.environ)
+    os.environ["OFE_SLOW_TESTS"] = long
+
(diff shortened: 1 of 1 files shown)


richardjgowers (changes requested):
I wouldn't worry about making it hidden, `openfe test` isn't too hard to figure out.  I wouldn't complicate this by trying to implement further options to `pytest`, the full pytest command isn't hard to access if you want that sort of control

richardjgowers:
@dwhswenson if I run this locally I get two failures: 
```
======================================================= short test summary info =======================================================
ERROR code/openfe/openfecli/tests/test_cli.py::TestCLI::test_command_sections - UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 3173: ordinal not in range(128)
ERROR code/openfe/openfecli/tests/test_cli.py::TestCLI::test_get_installed_plugins - UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 3173: ordinal not in range(128)
```

Running pytest manually for openfecli doesn't trigger these failures.

dwhswenson:
@richardjgowers I can't reproduce those test failures. Can you confirm that when you say "running pytest locally" you mean running via `pytest --pyargs openfecli`, and that the content of your installed `openfecli/commands/` matches the repo (no extra files; no diff in `__init__`, etc?) Can you also check the output of `openfe --help`?

Failing the `test_command_sections` test makes me suspect a seriously borked installation. Please note the test itself:

https://github.com/OpenFreeEnergy/openfe/blob/1743cfe8a90da77a5422c90c16bf4380a6d896e3/openfecli/tests/test_cli.py#L51-L57

And the line of code it is testing:

https://github.com/OpenFreeEnergy/openfe/blob/1743cfe8a90da77a5422c90c16bf4380a6d896e3/openfecli/cli.py#L17-L18

richardjgowers (approved):

PR #378, opened 2023-04-30
Moist test

richardjgowers:


openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -786,3 +788,27 @@ def test_constraints(tyk2_xml, tyk2_reference_xml):
             assert a.get('p1') == b.get('p1')
             assert a.get('p2') == b.get('p2')
             assert float(a.get('d')) == pytest.approx(float(b.get('d')))
+
+
+@pytest.mark.slow
```
IAlibay, line 793: As per my comments last week, can we please not have tests that run the integrator in the standard slow tests? I need to be able to dry run things regularly when I work on things but if I have to also run a simulation that adds a massive additional burden.

My take is that you should make this a "runintegrator" flag or similar. It's only really necessary because we assume that some folks will have platforms that can't run an integrator.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -348,19 +348,27 @@ def run(self, *, dry=False, verbose=True,
         equil_time = sim_settings.equilibration_length.to('femtosecond')
         equil_steps = round(equil_time / timestep)
 
+        logger.debug(f'Timestep is {timestep}')
```
IAlibay, line 351: I don't understand why the logger debug is repeating what's in settings? What are you specifically trying to capture here?

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -786,3 +788,27 @@ def test_constraints(tyk2_xml, tyk2_reference_xml):
...
+def test_moist_test(benzene_vacuum_system, toluene_vacuum_system,
+                    benzene_to_toluene_mapping, tmpdir):
+    # this test actually runs MD
+    # if this passes, you're 99% likely to have a good time
+    # these settings are a small sim, that has enough eq that it doesn't occaisionally crash
```
IAlibay, line 798: You'll occasionally crash anyways - we see this in practice, random seeds can be problematic even when things should all be ok - I'd stick some kind of rerun fixture similar to https://stackoverflow.com/questions/52967150/can-i-retry-for-failed-tests-in-pytest

IAlibay (changes requested):

IAlibay:
Also you'll need to temporarily enable the run slow env variable to see if it actually can run on gh actions

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
IAlibay: I'll go back on this for now if we can make this more of a "minimal" test - i.e. a methane self transformation. Assuming a stable initial conformation, that should "always" be stable (since there are no transforming atoms happening)

IAlibay: We should also enforce OPENMM_CPU_THREADS=1

richardjgowers: I've added a fixture for threads=1, it doesn't seem to make any difference for me locally however.  The tests take 20s/5s for CPU/CUDA respectively, I think this is in line/faster than other slow tests, so it's probably ok to not label them differently.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
richardjgowers: I was debugging the rounding of steps to check how small I could make the test.  Yes this is duplicate information from the settings, but it's useful to have here if you want to debug these kinds of things.

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
IAlibay: Can you undo the env variable and check here? The gh actions CPUs are very very old

richardjgowers:
@ialibay tests ran in 1m47 with the "slow" test
`openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_openmm_run_engine[CPU] 
[gw0] [ 95%] PASSED openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_openmm_run_engine[CPU] 
`

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
IAlibay: Looks like it ran pretty quickly, we could just leave it in the non slow tests tbh

IAlibay (approved):
Looks good, although can you either drop opencl or put a skip for macos?

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -786,3 +788,57 @@ def test_constraints(tyk2_xml, tyk2_reference_xml):
...
+            del os.environ['OPENMM_CPU_THREADS']
+        else:
+            os.environ['OPENMM_CPU_THREADS'] = previous
+
+
+@pytest.mark.flaky(reruns=3)  # pytest-rerunfailures; we can get bad minimisation
+@pytest.mark.parametrize('platform', ['CPU', 'CUDA', 'OpenCL'])
+def test_openmm_run_engine(benzene_vacuum_system, platform, available_platforms,
+                           set_openmm_threads_1, tmpdir):
+    if os.environ['CI'] and os.environ['RUNNER_OS'] == 'macOS' and platform == 'OpenCL':
```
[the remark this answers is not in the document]
IAlibay: We should probably just not allow OpenCL tbh, it's a wildly untested platform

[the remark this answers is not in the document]
richardjgowers: Is it that OpenCL will never work?

If we don't want to use OpenCL ever,  raise an issue and implement that in the Protocol if that's the case.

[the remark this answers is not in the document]
IAlibay: For macOS I belive it will _never_ work - that's an OS specific thing (OpenCL got deprecated a long while ago).

OpenCL is generally untested - we _should_ test it at some point, but I just don't have the capacity to do so.

PR #381, opened 2023-05-01
Update gather.py - protocol failure robustness

RiesBen:
adding checks for estimate and uncertainty fields, otherwise code crashes later. (Origin: A failed job produces a result, that might contain None fields.)

Discussable how to integrate this, this is just a first attempt ;)

### DIFF
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -21,8 +21,10 @@ def load_results(f):
 
 
 def get_names(result):
-    # Result to list of ligand names
-    return tuple(list(result['unit_results'].values())[0]['name'].split()[:2])
+    # Result to tuple of ligand names
+    nm = list(result['unit_results'].values())[0]['name']
+    toks = nm.split()
+    return toks[0], toks[2]
 
 
 def get_type(f):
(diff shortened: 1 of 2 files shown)


dwhswenson:
A failed job should produce a result that is a `ProtocolDAGFailure`, right? So I think we want to do the following:

* issue a warning with the filename and the short message of the error, which can be extracted from the JSON
* at the end (not for each failure) trigger something that gives instructions for how to rerun a failed edge. Note that this will probably overwrite the failure, so the failed run should be moved if they want to keep that data.
* we should probably insert *an* answer for this; I would say NaN. This way it gives users a "WTF" moment if they ignored the warning.
* we should also probably give instructions / a CLI command to extract the complete traceback from the JSON from a failed run

openfecli/commands/gather.py
```diff
@@ -87,6 +87,9 @@ def dp2(v: float) -> str:
         result = load_results(result_fn)
         if result is None:
             continue
+        elif result['estimate'] is None or result['uncertainty'] is None:
+            print("WARNING: Calculations for "+result_fn+" did not finish succesfully!")
```
mikemhenry, line 91:
```suggestion
            print(f"WARNING: Calculations for {result_fn} did not finish succesfully!")
```
I agree with @dwhswenson but if we do it this way, I'd prefer an f-string.

RiesBen:
agree with @dwsenson, this is more a quick fix and heads-up, that there is a issue, we need to solve :)

PR #384, opened 2023-05-03
CLI: ligand network viewer

dwhswenson:
We have this code sitting around to do an interactive visualization of a ligand network. No reason to make it so hard to find!


openfe/utils/atommapping_network_plotting.py
```diff
@@ -173,14 +173,16 @@ def plot_atommapping_network(network: LigandNetwork):
     """
     return AtomMappingNetworkDrawing(network.graph).fig
 
-
-if __name__ == "__main__":
-    import sys
+def main(filename):
```
IAlibay, line 176: This might speak to how rarely I actually do work on main, is calling this `main` that common? It feels a bit undescriptive for a method name.

IAlibay (approved):
lgtm (I'm sorta taking a "if it works it works stance here")

openfe/utils/atommapping_network_plotting.py
dwhswenson: I mean this is taken from the C world, which is also kind of echoed in Python, right? I have a module that is executable via `python -m openfe.blahblahblah`. I want to make the actual code that gets run there testable (at least in a smoke test). So under the `if __name__ == "__main__":`, I just call `main()`

dwhswenson: @richardjgowers : I don't think you actually want what I understand from your suggestion, which is something like this:

```python
# in openfe/utils/plot_atommapping_network

def plot_atommapping_network(network: LigandNetwork):
    ...

def _plot_network(filename: PathLike):
    ...
```

```python
# in openfecli/commands/ligand_network_viewer

from openfe.utils.plot_atommapping_network import _plot_network
```

Instead, in [894fd35](https://github.com/OpenFreeEnergy/openfe/pull/384/commits/894fd353fc1e6b23d957daadffb43f6ca05f9452) I just removed the ability to run `python -m openfe.utils.plot_atommapping_network file.graphml` and moved the functionality of `main` (first loading the file, then running the matplotlib) to the CLI. Breaks backwards compatibility, but I don't think anyone will care. The point of this PR was to provide a more natural way to do invoke that.

PR #388, opened 2023-05-04
Enable online-analysis by default

IAlibay:
It'll write out a yaml file every 200 MCMC steps (so by default 200 ps).

I set the target error to 0.0 that way there shouldn't be any early termination by default.


openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -811,7 +811,7 @@ def set_openmm_threads_1():
             os.environ['OPENMM_CPU_THREADS'] = previous
 
 
-@pytest.mark.slow
+#@pytest.mark.slow
```
IAlibay, line 814: re-enable before merging.

dwhswenson (approved):

IAlibay:
blergh keys.. always keys

IAlibay:
@richardjgowers I don't know if you were trying to hold off on merging things for the v0.8 release, so I'll wait on you before merging.

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -169,19 +169,20 @@ class Config:
...
+    online_analysis_interval: Optional[int] = 200
     """
     Interval at which to perform an analysis of the free energies.
     At each interval the free energy is estimate and the simulation is
     considered complete if the free energy estimate is below
     ``online_analysis_target_error``. If set, will write a yaml file with
     real time analysis data. Default `None`.
     """
-    online_analysis_target_error = 0.1 * unit.boltzmann_constant * unit.kelvin
+    online_analysis_target_error = 0.0 * unit.boltzmann_constant * unit.kelvin
```
richardjgowers, line 180: are these two settings that every 200 steps you check if the target error is below 0.0?  There's a `unit.timestep` for denoting steps if that's the case, else it's `unit.picosecond` maybe?

richardjgowers: Or more generally `Interval at which` is ambiguous and could mean either simulated time or timesteps so let's rephrase

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -169,19 +169,20 @@ class Config:
...
     """
     Target error for the online analysis measured in kT. Once the free energy
     is at or below this value, the simulation will be considered complete.
-    Default 0.1 * unit.bolzmann_constant * unit.kelvin
+    Default 0.0 * unit.bolzmann_constant * unit.kelvin, i.e. no early
+    termination will occur.
```
richardjgowers, line 185: If we can, suggesting a sensible default for when people do want to use this feature would be helpful.  Was 0.1 a reasonable value when using this feature?

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -169,19 +169,23 @@ class Config:
     or `independent` (independently sampled lambda windows).
     Default `repex`.
     """
-    online_analysis_interval: Optional[int] = None
+    online_analysis_interval: Optional[int] = 200
```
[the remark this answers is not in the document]
IAlibay: how about this?

IAlibay:
Going ahead and merging so that it doesn't hold back tutorials etc...

PR #393, opened 2023-05-05
Update gather.py - output file refactor

RiesBen:
slight outformat modificatios - making data  a bit more accessible:
- adding name content additionally seperatly for easy filtering and access
- adding unit to uncertainty header.
- not using +- for easier parsing


dwhswenson (changes requested):
CLI help (docstring) needs to be updated to reflect these changes.

Also, can we reformat to keep to 80 columns?

richardjgowers (approved):

PR #395, opened 2023-05-05
Fix handling of `ThreeParticleAverageSite` virtual sites

### LINKED ISSUE #394: Copying OpenMM VirtualSite particles from old/new system to hybrid fails
## Background

We want to copy non-alchemical virtual sites from one system to the hybrid one (for things like tip4pew waters):

https://github.com/OpenFreeEnergy/openfe/blob/d7bf903b769d4cdbfd4723739faa25b6e35d7ce6/openfe/protocols/openmm_rfe/_rfe_utils/relative.py#L601-L604

## Problem

The way this is currently done throws a `Exception: the System object does not own its corresponding OpenMM object` error. I guess that makes sense since the VirtualSite is attached to the existing System.

It _doesn't_ look like a deepcopy is the solution (or at least that'll throw you a pickling error).

IAlibay:
Towards #394


IAlibay:
The only test cases I can find where this will affect users "today" are for `ThreeParticleAverageSite` (when using `tip4pew`, `opc`, and `tip4pfb` from `openmmforcefields.amber`).

OpenFF also needs to handle `LocalCoordinationSite`, but we have no test cases for it currently. My take is that we should fix this now, and revisit it later once OpenFF has something ready for us to test against.

IAlibay:
Interesting - I manually checked a serialized version of a sytem XML generated with this and it looks fine. **However**, I get NaNs when used with the default protocol.

Looks like we're missing a call to computeVirtualSites in the MCMC move.

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -180,10 +216,10 @@ def test_dry_run_complex(benzene_complex_system, toluene_complex_system,
         stateB=toluene_complex_system,
         mapping={'ligand': benzene_to_toluene_mapping},
     )
-    unit = list(dag.protocol_units)[0]
+    dag_unit = list(dag.protocol_units)[0]
```
IAlibay, line 219: renaming a lot of these because it technically overrides `openff.units.unit` which I've found to be annoying at times (when you try to do further processing later on in the test)

IAlibay:
Simulations on RHFE set with tip4p-ew seems fine. I'd probably go ahead and say it's ok?

v8-ABFEs
vsites-ABFEs

mikemhenry (approved):

openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -2441,3 +2486,17 @@ def omm_hybrid_topology(self):
...
+        Checks the hybrid system and tells us if we have any virtual sites.
+
+        Returns
+        -------
+        has_virtual_sites : bool
+        """
+        has_virtual_sites = False
+        for ix in range(self._hybrid_system.getNumParticles()):
+            if self._hybrid_system.isVirtualSite(ix):
+                return True
```
richardjgowers, line 2502: you need a `else: return False` at the end here right?

PR #396, opened 2023-05-05
Improving our landing page

IAlibay:
I'm sure folks will have many opinions here - for now I'm seeing if we can get a scipy-like landing page going that actually makes sense to newcomers.


IAlibay:
Ok - I'm not going to claim by any means that this is the most amazing landing page, but it's better than what we had before. I'm not aiming for this to be the final iteration for this, I just want it to look "ok" for the workshop.

dwhswenson (approved):
Basic idea definitely seems fine, and passable for workshop. Thoughts to consider before merge (not necessarily blockers):

1. I'm not sure that we need direct links to CLI reference, and if we have a direct CLI link, I'd expect it to link to `guide/cli.html` (at least, if I had seen `guide/cli.html` and came back looking for it, I would be surprised this linked elsewhere)
2. Ditto on protocol reference, which, TBH, is currently a pretty ugly page.
3. Curious how this would look with a grid of 4 across the page. That might shrink it enough to fit above the fold on a laptop (especially if we drop sections as mentioned in 1 and 2).
4. Not the fault of this PR, but as we get toward something that looks better, the fact that we need CSS cleanup to match our colors becomes more and more obvious. I'm not sure there's time before workshop, though.

IAlibay:
> I'm not sure that we need direct links to CLI reference, and if we have a direct CLI link, I'd expect it to link to guide/cli.html (at least, if I had seen guide/cli.html and came back looking for it, I would be surprised this linked elsewhere)

This points out a deep problem with the way our docs currently exist - there's absolutely no flow. As a brand new user, I have 3 different entry points for the CLI and I wouldn't know where to look for what I need. I've changed it to guide/cli, but we need to rethink how we present this imho.

> Ditto on protocol reference, which, TBH, is currently a pretty ugly page.

The protocol reference is there because of the above too. If I ask folks to look for settings it would be quasi impossible for them to actually find out where the settings page is. I'm keeping it here for now because I want them to be able to actually be able to find it.

> Curious how this would look with a grid of 4 across the page. That might shrink it enough to fit above the fold on a laptop (especially if we drop sections as mentioned in 1 and 2).

The grids are screen size dependent, see: https://github.com/OpenFreeEnergy/openfe/pull/396/files#diff-8d068e8797e88947c320f79e856c3e16a72b730124a8f9d7031e2c4680dfa534R17

I can bump it up to 4 for larger screens. On smaller screens it becomes really squished.

IAlibay:
I'm going to go ahead and merge this, we can iterate on it later if needed.

PR #397, opened 2023-05-06
Fix nesting of shared directory for checkpoint

dwhswenson:
Per docs on `MultistateReporter`, ["this should NOT be a full path, and instead just a filename"](https://github.com/choderalab/openmmtools/blob/abd4011f1c756722ab8671097335c0e895f5741f/openmmtools/multistate/multistatereporter.py#L93).

Going to add tests to `test_openmm_run_engine` to check this -- that is the appropriate "moist" test, right?

### DIFF
--- a/openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
+++ b/openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
@@ -811,7 +811,7 @@ def set_openmm_threads_1():
             os.environ['OPENMM_CPU_THREADS'] = previous
 
 
-# @pytest.mark.slow
+@pytest.mark.slow
 @pytest.mark.flaky(reruns=3)  # pytest-rerunfailures; we can get bad minimisation
 @pytest.mark.parametrize('platform', ['CPU', 'CUDA'])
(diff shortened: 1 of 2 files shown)


IAlibay:
> that is the appropriate "moist" test, right?

Yes

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -811,7 +811,7 @@ def set_openmm_threads_1():
             os.environ['OPENMM_CPU_THREADS'] = previous
 
 
-@pytest.mark.slow
+# @pytest.mark.slow
```
dwhswenson, line 814: Revert this before merge

dwhswenson:
Ready for review. Turned on the slow test for this, turn it back off before merge.

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
dwhswenson:
```suggestion
@pytest.mark.slow
```

richardjgowers (approved):
checked locally, works

PR #398, opened 2023-05-06
Silence `openmmtools.multistate` API warnings

dwhswenson:
As some of you may be aware, "The openmmtools.multistate API is experimental and may change in future releases."

I suppose it's a good thing for us to know. That said, we probably don't need to announce this to our CLI users 3 times during `quickrun`.

Also, this PR makes it so that our "you don't have a GPU" warning looks a little better.

### DIFF
--- a/openfe/utils/logging_filter.py
+++ b/openfe/utils/logging_filter.py
@@ -1,0 +1,8 @@
+import logging
+
+class MsgIncludesStringFilter:
+    def __init__(self, string):
+        self.string = string
+
+    def filter(self, record):
+        return not self.string in record.msg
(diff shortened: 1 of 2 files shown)


openfe/utils/logging_filter.py
```diff
@@ -0,0 +1,8 @@
+import logging
+
+class MsgIncludesStringFilter:
```
richardjgowers, line 3: document what this is used for

richardjgowers (changes requested):

openfe/utils/logging_filter.py
dwhswenson: done

richardjgowers (approved):

PR #401, opened 2023-05-06
Make docs reasonable to be user facing

IAlibay:


dwhswenson:
If you're doing this kind of cleanup, it might be worth merging my open docs PR first. I think it mainly needs this kind of stuff. I'm more concerned about dealing with conflicts than about getting those last 2 to-dos in.

richardjgowers (approved):

IAlibay:
@richardjgowers there's still some work to be done here on updating plan_* docstring we can either merge this as-is and deal with the rest later or deal with it here

PR #399, opened 2023-05-06
Add ligand network output to CLI network planners

dwhswenson:
This adds output of the `LigandNetwork` to the file `$OUTPUT_DIR/ligand_network.graphml` in the `plan-*-network` CLI commands.

I also slightly re-nested, re-adding the `transformations/` dir that was removed in #382.

PR #402, opened 2023-05-07
Remove mapper from plan-*-network command

dwhswenson:
This is for two reasons:

1. We're seeing bad networks (disconnected core components) with the current code.

2. Rather than offer users a choice here, this means that (at least for now) `plan-*-network` is just our recommendation of best practices. We were already doing this with everything else (scorer, network type); mapper was previously an exception to the rule.

@IAlibay : If this is our best practices, could you give guidance on the recommended invocation of `LomapAtomMapper`?

Marking as WIP until after #399 is in, at which point I'll more easily be able to validate that this dodges the bad mappings we were getting before.


openfecli/commands/plan_rbfe_network.py
```diff
@@ -126,7 +126,7 @@ def plan_rbfe_network(
     write("")
 
     write("Using Options:")
-    mapper_obj = MAPPER.get(mapper)()
+    mapper_obj = LomapAtomMapper(element_change=False)
```
IAlibay, line 129:
```suggestion
    mapper_obj = LomapAtomMapper(time=20, threed=True, element_change=False, max3d=1)
```
This is what we use in openfe-benchmarks by default.

openfecli/commands/plan_rhfe_network.py
```diff
@@ -120,7 +120,7 @@ def plan_rhfe_network(molecules: List[str], output_dir: str, mapper: str):
     write("")
 
     write("Using Options:")
-    mapper_obj = MAPPER.get(mapper)()
+    mapper_obj = LomapAtomMapper(element_change=False)
```
IAlibay, line 123:
```suggestion
    mapper_obj = LomapAtomMapper(time=20, threed=True, element_change=False, max3d=1)
```

IAlibay (changes requested):
I think we'll also need to disable the element change stuff that @RiesBen added into the LigandAtomMapper class - if it's providing us with issues let's just use the standard Lomap behaviour and not try to mix the two (especially if we're making folks use the network planner in any tutorials).

IAlibay (changes requested):
I think we'll also need to disable the element change stuff that @RiesBen added into the LigandAtomMapper class - if it's providing us with issues let's just use the standard Lomap behaviour and not try to mix the two (especially if we're making folks use the network planner in any tutorials).

openfecli/commands/plan_rhfe_network.py
dwhswenson: This fails for the RHFE set we're using. MCVE: get `benzenes_RHFE.sdf` from `openfe fetch rhfe-tutorial`. Then:

```python
import openfe
from rdkit import Chem
from openfe.setup.ligand_network_planning import generate_minimal_spanning_network
from openfe.setup.atom_mapping.lomap_scorers import default_lomap_score

supp = Chem.SDMolSupplier("benzenes_RHFE.sdf", removeHs=False)
mols = [openfe.SmallMoleculeComponent.from_rdkit(mol) for mol in supp]
mapper = openfe.LomapAtomMapper(time=20, threed=True, element_change=False, max3d=1)
generate_minimal_spanning_network(mols, [mapper], default_lomap_score)
```

```pytb
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[18], line 9
      7 mols = [openfe.SmallMoleculeComponent.from_rdkit(mol) for mol in supp]
      8 mapper = openfe.LomapAtomMapper(time=20, threed=True, element_change=False, max3d=1)
----> 9 generate_minimal_spanning_network(mols, [mapper], default_lomap_score)

File ~/Dropbox/omsf/src/openfe/openfe/setup/ligand_network_planning.py:153, in generate_minimal_spanning_network(ligands, mappers, scorer)
    151 missing_nodes = set(network.nodes) - set(min_network.nodes)
    152 if missing_nodes:
--> 153     raise RuntimeError("Unable to create edges to some nodes: "
    154                        + str(list(missing_nodes)))
    156 return min_network

RuntimeError: Unable to create edges to some nodes: [SmallMoleculeComponent(name=lig_7)]
```

Seems to work at `max3d=2.0`?

IAlibay: Are we sure we're using a set that can have a whole network generated from? Did we remove the problematic ligands?

dwhswenson: I'm using whatever @RiesBen put in the easy campaign demo scripts in the ExampleNotebooks repo. If there's a better place for a set of RHFE ligands, let me know and I can switch. But I think these correspond roughly to the results he gave (same number, at least). My network was different from his (expected), so I had to fake in that some of the empty files were from a different edge (empty files, so no one would notice!)

IAlibay: The benzenes set is a radial network not a minimum spanning tree, this is the cause of this issue I think (the central ligand should be ligand 12 iirc).

dwhswenson: Aha... well, network planner uses MST. We can rewrite for radial, but it is a bad approach: completely bakes in the idea that small molecules MUST have a name (at least for the CLI). And do we want to say that our recommended best practice is radial?

Any ideas for workarounds here?

IAlibay: honestly I have no idea - I assume somehow the results you have were generated by the CLI? are they MST?

PR #409, opened 2023-05-08
Rbfe fetch assets

richardjgowers:
I've checked locally, this gives an output that `openfe gather` can work with and matches what `plan-rbfe` will produce


dwhswenson (approved):
lgtm, thanks! I'll test them out once they're in

PR #411, opened 2023-05-09
Progress bar for maximal network

dwhswenson:
Creating a `LigandNetwork` is something that we expect to be an interactive experience. But creating the maximal network (a starting point for other networks) scales as $N^2$ times the cost of an atom mapping, which isn't necessarily cheap. So this ends up being the bottleneck in many setups.

This PR adds a `progress` parameter to the maximal network planner. Users can interface with it in 3 ways:

1. Set `True` (default): It uses a `tqdm` progress bar that only shows up if the calculation takes more than 1.5 seconds.
2. Set `False`: no progress bar
3. Provide a custom progress bar that takes an iterable and spits out the values, after whatever side-effect illustrates progress.

For UX of our CLI `plan-*-network` tools, this is a huge win. Instead of seeing a screen that looks like it is hanging, we see an active progress bar. The same will be true for notebook usage of network planners.

But the downside of this is that we're sticking output into deep library code. One option would be for network planners that initially generate the maximal network to also take a `progress` parameter, and pass it down. Another would be to convert network planners to classes (planned) and have them carry a `progress` attribute. (We created a `Progress` API for OPS, and tools that needed it used it as a mix-in. That might be the route we select in the end here.)

This PR technically should add `tqdm` to the list of our requirements, although it is already in our stack via `openff-toolkit` → `openff-interchange` → `panedr` → `pyedr` → `tqdm`.


openfe/setup/ligand_network_planning.py
```diff
@@ -79,11 +81,12 @@ def generate_maximal_network(
     ligands: Iterable[SmallMoleculeComponent],
     mappers: Iterable[LigandAtomMapper],
     scorer: Optional[Callable[[LigandAtomMapper], float]] = None,
+    progress: Union[bool, Callable[[Iterable], Iterable]] = True,
     # allow_disconnected=True
 ):
     """Create a network with all possible proposed mappings.
 
-    This will attempt to create (and optionally score) all possible mappings
+    This pt to create (and optionally score) all possible mappings
```
richardjgowers, line 89: Typo

dwhswenson: looks like I ate a few characters there. fixed!

dwhswenson:
Long term, I think a more OOP solution will be better here. But that's already somewhat in the works.

PR #419, opened 2023-05-11
adding style to the docs

RiesBen:
a css file for the docs, to make them OpenFE stylish.


dwhswenson (approved):
Looks very good, IMO. Definitely an improvement over what we had, and we can iterate on this.

I'll note a few things here, but nothing that should be blocking:

* when there's a toctree that only has one level (no nesting), it isn't immediately obvious to me that those are links. See Core Models, Simulation Setup.
* we may want to play with the CSS on the code sections
* we *will* want to figure out better CSS for `<select>` elements (see Installation), although these default to being OS/Browser specific, so this may be a pain.

PR #427, opened 2023-05-12
Remove top level settings class entry

### LINKED ISSUE #413: Remove openfe.protocols.openmm_rfe.equil_rfe_settings.RelativeHybridTopologyProtocolSettings from settings docs
It's not giving a correct overview of the settings (missing thermo and force field settings).

IAlibay:
Fixes #413

### DIFF
--- a/docs/reference/api/openmm_rfe.rst
+++ b/docs/reference/api/openmm_rfe.rst
@@ -16,14 +16,6 @@ cases (such as when doing gas phase calculations) where you will need to tweak
 some of the following settings.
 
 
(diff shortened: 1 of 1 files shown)


richardjgowers (changes requested):
not sure I agree with this one.  I would rather we include the inherited settings members?

IAlibay:
> not sure I agree with this one. I would rather we include the inherited settings members?

how

IAlibay:
@richardjgowers - now you have BaseSettings (points to nothing) and a ThermoSettings entry that has a broken link.

IAlibay:
Fixed by much bettter people.

PR #431, opened 2023-05-13
all to not water for writing

IAlibay:
This should avoid the 3TB outputs


dwhswenson (changes requested):
gufe keys

richardjgowers:
We need a script which is gufe key amnesty...

IAlibay:
> gufe keys

done

dwhswenson:
I'm kind of concerned that we don't have a test that this breaks.... or would this break the moist tests, and we're just not running them?

IAlibay:
the question I'm going to ask here - why would you expect this to break a test?

richardjgowers (approved):
lgtm

dwhswenson (approved):

PR #437, opened 2023-06-05
add  fetchables for showcase notebook inputs

mikemhenry:


openfecli/fetchables.py
```diff
@@ -5,6 +5,18 @@
...
                     "OpenFreeEnergy/ExampleNotebooks/main/")
 
+RBFE_SHOWCASE = URLFetcher(
+    resources=[
+        (_EXAMPLE_NB_BASE + "openmm_rbfe", "inputs/ligands.sdf"),
+        (_EXAMPLE_NB_BASE + "openmm_rbfe", "inputs/Jnk1_ligands.sdf"),
+        (_EXAMPLE_NB_BASE + "openmm_rbfe", "inputs/181L_mod_capped_protonated.pdb"),
+    ],
+    short_name="rbfe-showcase",
+    short_help="Inputes needed for the RBFE Showcase Notebook",
```
dwhswenson, line 15:
```suggestion
    short_help="Inputs needed for the RBFE Showcase Notebook",
```

But really, is there any reason not to also include the notebook itself? I suppose technically this will overwrite something on colab, but should overwrite with the same thing, no? Including the notebook makes this useful for regular users, too.

mikemhenry:
@dwhswenson ready for review :tada: 
Right now it works for the getting the showcase on binder, I don't want it to pull down the notebook since with the way it is setup, the notebook + files pulled down won't work together.

mikemhenry:
Also once this gets merged in, I will need to cut a new release + make a new installer for our colab installer

openfecli/fetchables.py
```diff
@@ -5,6 +5,20 @@
 _EXAMPLE_NB_BASE = ("https://raw.githubusercontent.com/"
```
richardjgowers, line 5: If we change the url here to be against a tag/commit then we can't accidentally break these. If we are patching the files, it makes more sense to be intentional about updating this url to reflect that.

dwhswenson: @richardjgowers: Is the intent for `try.openfree.energy` to only update on releases? Like, a typo fix in  notebook will only be reflected on a release of `openfe`? 

If we do that, I'd recommend setting up some automation to check that that this is the most up-to-date. "Be intentional about updating the URL" sounds to me like "we're going to forget to update the URL."

openfecli/fetchables.py
```diff
@@ -5,6 +5,20 @@
...
+    resources=[
+        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "ligands.sdf"),
+        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "Jnk1_ligands.sdf"),
+        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "181L_mod_capped_protonated.pdb"),
+        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "benzene.sdf"),
+        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "styrene.sdf"),
+    ],
+    short_name="rbfe-showcase",
+    short_help="Inputs needed for the RBFE Showcase Notebook",
+    section="Tutorials",
```
dwhswenson, line 18: Seems to me that, without the actual notebook, this is only intended for internal use. If so, change the section name (to anything, really, but I default to using `"hidden"`) so that it doesn't show in help.

openfecli/fetchables.py
mikemhenry: With the installer hard-coded + the URL using a tagged release, then we are pretty set in keeping `try.openfree.energy` stable. Even if we push a new release and have API changes, the notebook will still be using the "old" installer, so it will stay stable and just have to remember to update it.

mikemhenry:
We should hold off merging this until people are happy with https://github.com/OpenFreeEnergy/ExampleNotebooks/pull/69 (just in case I need to make changes to this PR)

openfecli/fetchables.py
richardjgowers: Yes I think `try.` should be off releases not HEAD

richardjgowers (approved):

PR #441, opened 2023-06-08
rename water molecules so addSolvent doesn't replace with ion

richardjgowers:


richardjgowers:
it took us a few months of running these sims before we hit the issue, so we'd really have to force a random seed to repro.  `_addIons` uses `random.choice` so we could keep printing current seeds till we get a bad replacement, then use that seed as our regression test?

IAlibay (approved):
months :P we merged that refactor last week!

But yeah that's fair. Is it worth just testing that the final hybrid topology has no residuess named WAT?

richardjgowers:
@IAlibay we've been calling `addSolvent` in an indentical way for a while, so this might have happened before; we only noticed this issue because it screwed up something unrelated (residue tracking) in our code

IAlibay:
> @IAlibay we've been calling `addSolvent` in an indentical way for a while, so this might have happened before; we only noticed this issue because it screwed up something unrelated (residue tracking) in our code

ah ok I see what you mean

PR #445, opened 2023-06-09
Proposed path for handling Settings schema changes for a protocol

dotsdl:


openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -527,3 +535,24 @@ class Config:
...
     # Simulation run settings
     simulation_settings: SimulationSettings
+
+
+
+def from_dict(dct):
+
+    dct = dict(dct)  # make a copy
+
+    if 'splitting' in dct.get('integrator_settings'):
```
IAlibay, line 545: Questions:

1 - Is this versioned or do we just keep adding entries that get removed over time?
2 - Thoughts on being noisy, i.e. warning folks that the settings they are getting aren't the setttings on disk?
3 - What about if we have non-optional new entries? (i.e. there's no defined default)

dwhswenson: Basically, this is a minimal example. We discussed these points as options that could be added here.

> Is this versioned or do we just keep adding entries that get removed over time?

This is the non-versioned hack. The better solution is to introduce proper versioning, but this was just the output of a working session to see an overview.

> Thoughts on being noisy, i.e. warning folks that the settings they are getting aren't the setttings on disk?

Absolutely, when relevant, as it is in this example. This is tied to a specific protocol, so in this example, you could check the provided splitting against what will be used. If there's a difference you could, e.g., error with an option to just give a warning if some env var is set. 

For some cases (name of option was changed but default value was not), there would be no need for this.

> What about if we have non-optional new entries? (i.e. there's no defined default)

I think this is an issue where we would just have to error. But since it is handled here, we could give a much more user-friendly error (tell version of the code they need to load their file).

IAlibay: Is the initial openfe & gufe version informattion accessible on deserialisation? It'd be good to be able to point folks to a version difference.

richardjgowers:
done in https://github.com/OpenFreeEnergy/gufe/pull/227

PR #450, opened 2023-06-12
Add PDB serialization in RFE protocol

IAlibay:
This will let us actually look at our systems properly.


openfe/protocols/openmm_utils/utils.py
```diff
@@ -0,0 +1,71 @@
...
+      input Topology.
+
+    Returns
+    -------
+    top : app.Topology
+      Subsampled Topology.
+    """
+    top = app.Topology()
+    chains = []
+    residues = []
```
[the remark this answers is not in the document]
IAlibay: I was always under the impression that `.add()` was significantly slower than creating a list and then creating the set after the fact?

richardjgowers:
@IAlibay is this mdtraj downstream?  I don't understand where mdtraj is coming into the picture here

IAlibay:
> @IAlibay is this mdtraj downstream? I don't understand where mdtraj is coming into the picture here

Upstream - we use MDTraj as part of the HTF class, particularly for generating Topologies.

openfe/protocols/openmm_utils/utils.py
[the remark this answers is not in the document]
IAlibay: Huh I guess not.

openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -2346,21 +2347,99 @@ def _create_mdtraj_topology(self):
             at1_uniq = at1_hybrid_idx in self._atom_classes['unique_new_atoms']
             at2_uniq = at2_hybrid_idx in self._atom_classes['unique_new_atoms']
             if at1_uniq or at2_uniq:
-                if atom1.index in self._atom_classes['unique_new_atoms']:
-                    atom1_to_bond = added_atoms[atom1.index]
+                if at1_uniq:
```
IAlibay, line 2350: Bonds weren't being added properly here, which was causing issues with the roundtripping of these MDTraj topologies.

Turns out we didn't need to roundtrip, but we should really fix this whilst we're here.

openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -2346,21 +2347,99 @@ def _create_mdtraj_topology(self):
...
-                    atom2_to_bond = atom2
+                    old_idx = self._hybrid_to_old_map[at2_hybrid_idx]
+                    atom2_to_bond = hybrid_topology.atom(old_idx)
 
                 hybrid_topology.add_bond(atom1_to_bond, atom2_to_bond)
 
         return hybrid_topology
 
 
+    def _create_hybrid_topology(self):
```
IAlibay, line 2367: So we can't do openmm -> MDTraj -> openmm because indices will get renumered.

Instead here we just directly create a hybrid Topology from the input old / new topologies

openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -2346,21 +2347,99 @@ def _create_mdtraj_topology(self):
...
+                atom_list.append(list(self._old_topology.atoms())[idx])
+            else:
+                idx = self._hybrid_to_new_map[pidx]
+                atom_list.append(list(self._new_topology.atoms())[idx])
+
+        # Now we loop over the atoms and add them in alongside chains & resids
+        
+        # Non ideal variables to track the previous set of residues & chains
+        # without having to constantly search backwards
+        prev_res = None
```
IAlibay, line 2396: This is ugly, I just ran out of steam on how to do this without lots of complexity below. I'm sure it's a simple thing I'm not thinking of.

openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -2489,8 +2568,14 @@ def omm_hybrid_topology(self):
         Returns
         -------
         hybrid_topology : simtk.openmm.app.Topology
+
+
+        .. versionchanged:: OpenFE 0.11
```
richardjgowers, line 2573: not sure we're tracking changes until 1.0, can probably not bother with these till then

openfe/protocols/openmm_utils/utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # First pass - get chain & resid info
+    for at in topology.atoms():
+        if at.index in subsample_indices:
+            residues.add(at.residue.index)
+            chains.add(at.residue.chain.index)
+
+    # Now let's actually add things
+    for chain_id, chain in enumerate(topology.chains()):
+        if chain.index in chains:
```
richardjgowers, line 47: if you instead do `if not chain.index in chains: continue` you can avoid going 6 layers deep on the indentation and running out of known universe

openfe/protocols/openmm_rfe/_rfe_utils/relative.py
IAlibay: I only added this here so we can keep track of the changes we made to HTF - that way when things finally get moved into OpenMMTools, it'll be easier to quickly identify which features changed from perses.

IAlibay:
image
Well I messed up something somewhere...

IAlibay:
image
When your code actually sorts properly 🤦🏽

IAlibay:
Ok we have two options here:

1. We stick with an MDTraj experience - we use the fixed MDTraj hybrid topology creation and directly subsample the trajectory from it.
  - Advantages:
    - It's backwards compatible
    - Fewer lines of code changes
  - Disadvantages:
    - It squashes the unique atoms from both endstates into one residue, making it hard to work out which atoms belong to which end state.
2. We switch to using an OpenMM hybrid Topology.
  -  Advantages:
      - The unique atoms can be placed in separate residues, making it easier to track / analyse after the fact.
  - Disadvantages:
     - Not backwards compatible
     - More lines of code changes

openfe/protocols/openmm_rfe/_rfe_utils/relative.py
IAlibay: This doesn't have as much utility now, but the _old_ roundtrip method is still incorrect - at least this one doesn't renumber indices (although the residue will be different than those created by the MDTraj trajectory).

I would like to keep this around so that we can switch to this in 1.x?

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -523,6 +525,12 @@ def run(self, *, dry=False, verbose=True,
             checkpoint_storage=sim_settings.checkpoint_storage,
         )
 
+        #  b. Write out a PDB containing the subsampled hybrid state
+        traj = mdtraj.Trajectory(
```
IAlibay, line 529: I removed the new utils bits (for now at least) and opted for this since it's the best backwards compatibility we're going to get.

IAlibay (commented):
For the sake of getting this done, I've opted for the MDTraj option for now. It's not amazing, but at least it's fewer lines of code.

I would advocate for a change to a separated out residue for unique atoms in 1.x though.

openfe/utils/handle_trajectories.py
```diff
@@ -0,0 +1,207 @@
...
+    cell_lengths.units = 'angstrom'
+    cell_angles = ncfile.createVariable(
+        'cell_angles', 'f8', ('frame', 'cell_spatial')
+    )
+    cell_angles.units = 'degree'
+    
+    return ncfile
+
+
+def _get_unitcell(
```
IAlibay, line 123: So the coordinates are coming out reasonable for the individual molecules, but it looks like there's some kind of imaging issue? I'm not sure that I didn't mess up the unitcell conversion tbh.

image

IAlibay:
Right it's missing some tests and it'll need working out what's up with the centering. But more or less everything else is sorted out. I'm going to leave it to you @richardjgowers ?

richardjgowers (approved):

PR #451, opened 2023-06-12
test newest version of openmmtools

mikemhenry:


mikemhenry:
We have a change coming in openmmtools where online analysis interval needs to be a multiple of the checkpoint interval, so we should make our default match. I will need to update the key checks but before I did that I wanted to make sure people were happy with this.

dwhswenson (changes requested):
If changing defaults, you need to update gufe keys

mikemhenry:
> If changing defaults, you need to update gufe keys

> I will need to update the key checks but before I did that I wanted to make sure people were happy with this.

richardjgowers (changes requested):
We've got checks elsewhere that check that these steps are compatible with each other so we fail early

IAlibay:
> We've got checks elsewhere that check that these steps are compatible with each other so we fail early

We do? I thought we didn't because cross pydantic models checks were broken until the next major release?

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -275,13 +275,13 @@ class Config:
...
     """
-    online_analysis_interval: Optional[int] = 200
+    online_analysis_interval: Optional[int] = 250
     """
     MCMC steps (i.e. ``IntegratorSettings.n_steps``) interval at which
     to perform an analysis of the free energies.
 
     At each interval, real time analysis data (e.g. current free energy
-    estimate and timing data) will be written to a yaml file named 
+    estimate and timing data) will be written to a yaml file named
```
IAlibay, line 284: Can you update this to make it clear of the new limitation? i.e. must be a multiple of X

IAlibay (changes requested):

openfe/protocols/openmm_rfe/equil_rfe_settings.py
mikemhenry: Yes!

mikemhenry:
> We've got checks elsewhere that check that these steps are compatible with each other so we fail early


What do you mean? Before I made the change, this is the error we got `FAILED openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_ligand_overlap_warning - ValueError: Online analysis interval: 200, must be a multiple of the checkpoint interval: 250` which is coming from openmmtools

mikemhenry:
Actually, that validation check is hard to do since they are two different settings objects. I think I could add a validator to `RelativeHybridTopologyProtocolSettings` But I am thinking that could be a seperate PR?

IAlibay:
> Actually, that validation check is hard to do since they are two different settings objects. I think I could add a validator to `RelativeHybridTopologyProtocolSettings` But I am thinking that could be a seperate PR?

Last I checked it won't work, until the next major release. It's a deeper issue with pydantic as-is.

richardjgowers:
@IAlibay @mikemhenry I was talking about the checks here: https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/protocols/openmm_utils/settings_validation.py#L40

These are fancy validators that are checking that other steps are divisible by each other etc.  Would it make sense to add another check here, or is the openmmtools error message enough?  (it's pretty clear tbh)

Also would it be better here to make the online analysis interval expressed in checkpoints?  E.g. if you set it to 2 it would run the online analysis every other checkpoint.  Not sure if that's more/less intuitive

IAlibay:
checkpoints as a unit is less intuitive to me, re: early validation, yeah I guess we could, the openmmtools message is rather clear though so it wouldn't be a priority imho

richardjgowers (approved):

mikemhenry:
I will make an issue so we can talk about it, the "problem" is the check will have to happen in our settings object that is a collection of other settings, since the setting is spread across two different settings objects, so the validator will have to live in the aggregate

dwhswenson (approved):
Once we're not pinned to a dev version of openmmtools, this should be good to go.

Issue #448, opened 2023-06-12
CI caches gufe install


richardjgowers:
our CI build cache seems to not reinstall HEAD of gufe but caches it.  Probably need to salt the definition of gufe@HEAD so it is pip installed for each build

mikemhenry:
Resolved with #452

PR #452, opened 2023-06-12
Install GUFE after we restore cache to alwasys get a fresh version

mikemhenry:
Resolves issue #448


mikemhenry:
After going through the different workflows, I think only `mypy` and our main CI workflow are affected.

richardjgowers (approved):

PR #454, opened 2023-06-13
allow single AtomMapper to be passed to network planners

richardjgowers:


openfe/setup/ligand_network_planning.py
```diff
@@ -49,6 +50,8 @@ def generate_radial_network(ligands: Iterable[SmallMoleculeComponent],
       If no scorer is supplied, the first mapping provided by the iterable
       of mappers will be used.
     """
+    if isinstance(mappers, AtomMapper):
```
IAlibay, line 53: having to repeat this is pretty ugly :/ could we not wrap it up into a decorator?

richardjgowers: nah, it makes it hard to read (the function is in two places).  It's two lines, it's fine

dwhswenson: Again, a call for Konnektor (or however the German spells it), where we could put the real work in a private method and the public method could handle this.

openfe/tests/setup/test_network_planning.py
```diff
@@ -133,6 +138,29 @@ def scoring_func(mapping):
...
+    def scorer(mapping):
+        return 1.0 / len(mapping.componentA_to_componentB)
+
+    network = openfe.ligand_network_planning.generate_minimal_spanning_network(
+        ligands=ligands,
+        mappers=mappers,
+        scorer=scorer,
+    )
+
+    assert isinstance(network, openfe.LigandNetwork)
```
dwhswenson, line 161: Add an assertion on the number of edges? (can be a `MultiDiGraph`, but shouldn't actually be multi).

dwhswenson (approved):
Minor changes suggested. Otherwise good by me.

PR #456, opened 2023-06-14
cofactors from CLI

### LINKED ISSUE #442: moist test with EG5
has crystal waters and cofactors

richardjgowers:
fixes #442

### DIFF
--- a/openfecli/parameters/molecules.py
+++ b/openfecli/parameters/molecules.py
@@ -59,6 +59,7 @@ def load_molecules(file_or_directory):
 def molecule_getter(user_input, context):
     return load_molecules(user_input)
 
+
 MOL_DIR = Option(
     "-M",
     "--molecules",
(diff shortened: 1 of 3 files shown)


RiesBen (approved):
the basic code changes look good to me! 

But we should add a test system with cofactors :)

openfecli/parameters/molecules.py
```diff
@@ -69,3 +69,10 @@ def molecule_getter(user_input, context):
     ),
     getter=molecule_getter,
 )
+
+COFACTORS = Option(
```
IAlibay, line 73: Two things:

1. This behaviour is difference than "MOL_DIR", do we expect folks to sometimes pass a directory of molecule files for spectator molecules?
2. This might be a bit late, but the fact that we use -c and -M and -p is just annoying.

dwhswenson: `-m` is already taken by a single-molecule version that is used elsewhere. Maybe capital letters when they're expected to contain multiple objects

IAlibay: was `-l` not available? (l for ligand)

richardjgowers:
Ok I've gone for `-C` for cofactors.  Also allows multiple files to be given, so `... -C file1.sdf -C file2.sdf` etc

PR #465, opened 2023-06-19
Update installer and container images to python 310

mikemhenry:


mikemhenry:
If we build python 3.10 images, then we can support google colab "natively" (the conda google colab installer requires python=3.10, so I have to manually create the installer to update the env, with this change, I can just point to our releases)

richardjgowers (approved):

PR #468, opened 2023-06-21
Reconfigure stdout to be line-buffered in quickrun

dwhswenson:
It looks like we're having an issue that, if a job is killed by a queueing system, some of the buffered output might not get written to the job's output file. This should force the buffer to flush on newline, which can hopefully solve some user confusion.

I considered putting this in the CLI `main`, but decided the one-liner can go in specific commands that need it. It should only be needed for commands that will be long-running (on a cluster).

### DIFF
--- a/openfecli/commands/quickrun.py
+++ b/openfecli/commands/quickrun.py
@@ -51,10 +51,14 @@ def quickrun(transformation, work_dir, output):
     """
     import gufe
     import os
+    import sys
     from gufe.protocols.protocoldag import execute_DAG
     from gufe.tokenization import JSON_HANDLER
     from openfe.utils.logging_filter import MsgIncludesStringFilter
     import logging
+
+    # avoid problems with output not showing if queueing system kills a job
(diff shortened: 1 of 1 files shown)


dwhswenson:
@mikemhenry : Could you check if this does, in fact, solve the problem? I was thinking you could use something like the `shorten_runs.py` script to make units shorter, then set some even-shorter wall time. Rather than running this code, you could also make a test run like that and compare with `PYTHONUNBUFFERED` unset and set: ideally, you can regularly reproduce when unset, and never reproduce when it is set!

mikemhenry:
Reference here: https://docs.python.org/3/library/io.html#io.TextIOWrapper (I was trying to understand how setting something abut buffering to `True` would reduce buffering, but it flushes the buffer on `\n`)

mikemhenry:
RE: testing

I will have to make a script that does something like sleep and print in a loop because it looks like dumping all the debug stuff has made everything show me the progress as I look at it

richardjgowers (approved):

PR #470, opened 2023-06-22
[WIP: don't review] New storage in network planner CLI

dwhswenson:
This will add a flag to use our new storage to output an alchemical network in the CLI network planners. For now, this option is hidden (doesn't show in `--help`), so we have full freedom to change the UX later. That way we can dogfood new storage with actual release code.

As of pushing the PR, nothing is actually hooked into storage. Just giving a heads-up that I'll be working on it.


atravitz:
closing, since this functionality will be covered by https://github.com/OpenFreeEnergy/openfe/pull/1833

PR #475, opened 2023-06-30
Added DCO

mikemhenry:


IAlibay (approved):

dwhswenson:
Isn't there something that integrates with GitHub so that your first contribution to a repo needs a CLA signature, and future things don't (with the understanding that sign it once means you've signed it for future contributions as well)? I'm not a fan of having a PR template which we then regularly ignore.

Basically, my opinion is that we should be dogfooding the external contributor experience as much as possible. I don't mind if my next PR to OpenFE says "please sign the CLA." I do get annoyed if every PR I open I have to check this box.

mikemhenry:
@dwhswenson I found this:
https://github.com/dcoapp/app
and looking around, most bots/automation require a sign-off on each commit, which sounds more annoying than checking a box

I couldn't find a native solution that github provides. For the pull request template they have query params, but it doesn't look like "first time" is an option https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/using-query-parameters-to-create-a-pull-request

I know there is a thing in the web UI that says "this is your first time contributing, read the license" or something, but I couldn't find a way to get access to that or something

IAlibay:
Would a PR comment for first time contributors work?

mikemhenry:
Like in a comment, something like "If you have contributed to the project before, you may remove this"?

IAlibay:
> Like in a comment, something like "If you have contributed to the project before, you may remove this"?
> 
>  

No, we can use gh action tooling to work out of its a first time contributor and then just either modify the PR lead comment or a later comment with the necessary wording they need to agree to.

richardjgowers (approved):

PR #479, opened 2023-07-05
Force no oechem

richardjgowers:


richardjgowers:
@IAlibay this is a pain to test inside the protocol but I've got some eg5 edges working with oechem present with this patch.

IAlibay:
I mean this is fine, but I don't understand why we needed to vendor this?

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -679,7 +680,9 @@ def run(self, *, dry=False, verbose=True,
     def _execute(
         self, ctx: gufe.Context, **kwargs,
     ) -> dict[str, Any]:
-        outputs = self.run(scratch_basepath=ctx.scratch, shared_basepath=ctx.shared)
+        with without_oechem_backend():
```
IAlibay, line 683: Actually we might need to review this, I suspect we'll run into trouble when we start offering folks the ability to choose what charge derivation method they can have (for example if they want elf charges)

IAlibay (approved):
Let's go with this for now, but we'll need to make it optional eventually (probably when it's fixed in openmmforcefields) - that way folks can choose different flavours of charge eventually.

PR #482, opened 2023-07-06
Network planner functions from indices/names

richardjgowers:


richardjgowers:
@IAlibay is this what you had in mind for convenience functions?

Then at what point are these classmethods not functions?

IAlibay:
Yeah that's pretty much what I'm thinking. It's silly small but it prevents us having to tell folks to write their own every time.

openfe/setup/ligand_network_planning.py
```diff
@@ -210,3 +211,95 @@ def generate_minimal_spanning_network(
...
+        raise ValueError(f"Duplicate names: {dupe_names}")
+
+    try:
+        ids = [(nm2idx[nm1], nm2idx[nm2]) for nm1, nm2 in names]
+    except KeyError:
+        badnames = [nm for nm in itertools.chain.from_iterable(names)
+                    if nm not in nm2idx]
+        available = [ligand.name for ligand in ligands]
+        raise KeyError(f"Invalid name(s) requested {badnames}.  "
+                       f"Available: {available}")
```
dwhswenson, line 260: Not sure whether the long list of available names may swamp the short list of bad ones in practice. However, this might be a good place for fuzzy string matching, if we ever want to add that for usability help: "Unknown ligand 'benzalderp'. Did you mean 'benzaldehyde'?"

Not a blocker, but might be a useful todo to note in code.

dwhswenson (approved):
Looks fine to me. Only real question is whether we want to encourage index-based usage -- I think name-based is likely to be useful, but index-based sounds like an easy place to make mistakes.

Doesn't mean it shouldn't be an option, but maybe we don't advertise it heavily....

IAlibay (approved):
lgtm

PR #485, opened 2023-07-11
Use common core option when using the Lomap Atom Mapper

richardjgowers:
Gives speed up of over 10x

### DIFF
--- a/openfe/setup/atom_mapping/ligandatommapper.py
+++ b/openfe/setup/atom_mapping/ligandatommapper.py
@@ -14,8 +14,6 @@ class LigandAtomMapper(gufe.AtomMapper):
     Subclasses will typically implement the ``_mappings_generator`` method,
     which returns an iterable of :class:`.LigandAtomMapping` suggestions.
     """
-    _no_element_changes: bool = False #TODO: to be removed
-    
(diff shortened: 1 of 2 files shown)


richardjgowers:
I think there's a better way to do this eventually, but this is such a nice win that I'd rather take this victory today then do the better way in time

openfe/setup/atom_mapping/ligandatommapper.py
```diff
@@ -14,8 +14,6 @@ class LigandAtomMapper(gufe.AtomMapper):
     Subclasses will typically implement the ``_mappings_generator`` method,
     which returns an iterable of :class:`.LigandAtomMapping` suggestions.
     """
-    _no_element_changes: bool = False #TODO: to be removed
```
IAlibay, line 17: Huh was this even ever used by anything?

IAlibay (approved):
LGTM, it's not exactly the least elegant thing in our stack tbh.

openfe/setup/atom_mapping/ligandatommapper.py
richardjgowers: It's the stub of something that was already removed

PR #488, opened 2023-07-13
Plan rbfe network radial

richardjgowers:


dwhswenson (changes requested):
I'm still not a fan of this special treatment of star maps. I think the better solution is either `plan-rbfe-starmap` or a more generic approach based on our plugin infrastructure. But since we won't make the time to do that, I've got some suggestions on improving this.

openfecli/plan_alchemical_networks_utils.py
```diff
@@ -36,10 +36,14 @@ def plan_alchemical_network_output(
...
     transformations_dir.mkdir(parents=True, exist_ok=True)
 
+    filenames = []
     for transformation in alchemical_network.edges:
         transformation_name = transformation.name or transformation.key
         filename = f"{transformation_name}.json"
+        filenames.append(filename)
         transformation.dump(transformations_dir / filename)
+
+    for filename in sorted(filenames):
```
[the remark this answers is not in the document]
richardjgowers: it's annoying when these are unsorted as it's harder to read/check what's been produced, I can sort as we get them out of the network

openfecli/commands/plan_rbfe_network.py
```diff
@@ -155,7 +166,25 @@ def plan_rbfe_network(
     write("\tMapping Scorer: " + str(mapping_scorer))
 
     # TODO:  write nice parameter
-    ligand_network_planner = generate_minimal_spanning_network
+    if radial:
```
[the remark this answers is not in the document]
richardjgowers: I raise a warning now, is there a proper way to do this through click?

openfecli/commands/plan_rbfe_network.py
```diff
@@ -82,9 +84,18 @@ def plan_rbfe_network_main(
...
     default="alchemicalNetwork",
 )
+@click.option(
+    "--radial", required=False, default=False, type=bool, is_flag=True,
+    help="Toggle on generation of a radial network, otherwise uses minimal-spanning-tree",
+)
+@click.option(
+    "--radial_hub", required=False, default=None, type=str,
+    help="For radial networks, the name of the ligand to use as the center",
+)
```
RiesBen, line 94: Could make sense, to implement these options in parameters and only import them here.

like MOL_DIR, that way you could share the implementation for rhfes an rbfes  and both clis have the same guaranteed cli experience.

openfecli/commands/plan_rhfe_network.py
```diff
@@ -68,8 +70,17 @@ def plan_rhfe_network_main(
...
     default="alchemicalNetwork",
 )
+@click.option(
+    "--radial", required=False, default=False, type=bool, is_flag=True,
+    help="Toggle on generation of a radial network, otherwise uses minimal-spanning-tree",
+)
+@click.option(
+    "--radial_hub", required=False, default=None, type=str,
+    help="For radial networks, the name of the ligand to use as the center",
+)
```
RiesBen, line 80: see above

openfecli/utils.py
```diff
@@ -67,3 +64,37 @@ def wrapper(*args, **kwargs):
...
+            raise click.BadParameter(f"Ambiguous ligand name ({radial_hub}) given as radial_hub")
+        elif not candidates:
+            raise click.BadParameter(f"ligand name '{radial_hub}' not found")
+        central_lig = candidates[0]
+    else:
+        central_lig = molecules[0]
+    # avoid hub to hub edges
+    molecules = [m for m in molecules if m is not central_lig]
+    write(f"\tUsing {central_lig} as radial hub")
+    ligand_network_planner = partial(generate_radial_network,
```
RiesBen, line 97: oh cool I  did not know about partial :) But I like it! :)

RiesBen (changes requested):
I like this addition to the CLI tools here! :) Only have one comment on the PR.

richardjgowers:
Probably superseded by #583 and #592

richardjgowers:
-

PR #495, opened 2023-07-19
Adds MLE estimates on gather command

richardjgowers:

### DIFF
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -20,6 +20,7 @@ def _get_column(val):
         col = np.floor(log10)
     return int(col)
 
+
 def format_estimate_uncertainty(
     est: float,
     unc: float,
@@ -113,6 +114,7 @@ def _get_ddgs(legs):
 
     return DDGs
 
+
 def _write_ddg(legs, writer):
     DDGs = _get_ddgs(legs)
     writer.writerow(["ligand_i", "ligand_j", "DDG(i->j) (kcal/mol)",
@@ -128,6 +130,7 @@ def _write_ddg(legs, writer):
                                                           bind_unc)
             writer.writerow([ligA, ligB, DDGhyd, hyd_unc])
 
+
 def _write_raw_dg(legs, writer):
     writer.writerow(["leg", "ligand_i", "ligand_j", "DG(i->j) (kcal/mol)",
                      "uncertainty (kcal/mol)"])
@@ -139,6 +142,7 @@ def _write_raw_dg(legs, writer):
             else:
                 m, u = format_estimate_uncertainty(m.m, u.m)
             writer.writerow([simtype, *ligpair, m, u])
+
 
 def _write_dg_mle(legs, writer):
     import networkx as nx
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -92,7 +92,9 @@ def gather(rootdir, output):
     """
     from collections import defaultdict
     import glob
+    import networkx as nx
     import numpy as np
+    from cinnabar.stats import mle
 
     def dp2(v: float) -> str:
         # turns 0.0012345 -> '0.0012', round() would get this wrong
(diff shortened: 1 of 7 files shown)


openfecli/commands/gather.py
```diff
@@ -141,24 +142,77 @@ def dp2(v: float) -> str:
...
+        for node, f, df in zip(g.nodes, f_i, df_i):
+            ligname = idx_to_nm[node]
+            MLEs.append((ligname, f, df))
+
+    output.write('measurement\ttype\tligand_i\tligand_j\testimate (kcal/mol)'
+                 '\tuncertainty (kcal/mol)\n')
+    # 5a) write out MLE values
+    for ligA, DG, unc_DG in MLEs:
+        DG, unc_DG = dp2(DG), dp2(unc_DG)
+        output.write(f'DGbind({ligA})\tDG(MLE)\tZero\t{ligA}\t{DG}\t{unc_DG}\n')
```
richardjgowers, line 201: @hannahbaumann this will print a line like:

```
measurement type ligand_i ligand_j estimate (kcal/mol) uncertainty (kcal/mol)
DGbind(lig_1234) DG(mle) Zero lig_1234 -4.2 0.23
```

would this make sense to you?  "zero" is meant to be the zero state reference point, not sure if there's a better label for that

IAlibay: It's not a zero state dG though? Unless you normalise against experiment your reference state is the first ligand passed to your MLE

IAlibay: In any case, if it's possible to inspect the original SMCs for stored experimental values this becomes a lot more useful. Otherwise what I would do is return a list of values with the leading entry being the 0 kcal/mol entry. That way it's clear what the central point is.

hannahbaumann: Yes, I think from that print statement I wouldn't know what zero means, so what the reference is and therefore what meaning the ("relative") dG has. Or is that described elsewhere?

dwhswenson (commented):
Rather than trying to cram everything into one table, why don't we make `gather` take flags to indicate what data to output? We're really offering 3 different tables. Instead of trying for force them into a single table, why not just own it and offer 3 tables?

Gather is also pretty fast, isn't it? So I don't think we save that much time by only walking the filesystem once.

Something like:

```bash
$ openfe gather --report ddg results/
$ openfe gather --report dg results/  # MLE, this is probably the default
$ openfe gather --report leg results/  # raw leg DG
```

If the theory is that we're expecting someone to parse this after, the current approach requires understanding what the special columns in our format mean, which, IMO, takes more teaching than a CLI option that switches between more intuitive output formats for each type of data.

To implement: https://click.palletsprojects.com/en/8.1.x/options/#choice-options

dwhswenson:
Most of the missing coverage would get hit if we had RHFE example data to test. But still is still good enough that codecov isn't complaining. Should be ready for review.

dwhswenson:
Raising my own concerns on this as of now:

* in the MLE DG estimate, reference ligand doesn't seem to show exactly 0.0 value. This makes things ambiguous and suggests a problem. I'm highly concerned about this.
* I haven't seen anyone do a proper cinnabar plot of numbers from this vs. experiment. There's no check in anything in this PR to ensure that our numbers are actually accurate.

I'm gone for a week as of now (well, a few of you might be able to get me in the next 13 hours.) So I'm logging these concerns so someone else can carry them through ad fix problems in the PR.

hannahbaumann:
Here a cinnabar plot of the DG values (after centralizing as done in cinnabar). I would say this looks reasonable?!
[]([](DG))

openfecli/tests/commands/test_gather.py
```diff
@@ -4,79 +4,105 @@
...
+        with resources.path('openfecli.tests.data', 'rbfe_results.tar.gz') as f:
             t = tarfile.open(f, mode='r')
             t.extractall('.')
 
         yield
 
+_EXPECTED_DG = b"""
+ligand	DG(MLE) (kcal/mol)	uncertainty (kcal/mol)
+lig_ejm_31	-0.09	0.05
+lig_ejm_42	0.7	0.1
```
richardjgowers, line 39: the sig fig here seems to be inconsistent.  i.e. `0.7` != `0.70`

dwhswenson: Implementation according to https://github.com/OpenFreeEnergy/openfe/pull/502#discussion_r1273261215, which I think is the right was to handle this. We don't have a fixed "sig figs of our estimate," so we say "well, we're correct up to the first digit of error, which will vary for each row of the table."

Assuming a fixed number of sig figs for the estimate gives rounded estimates outside the actual error bars, as described in https://github.com/OpenFreeEnergy/openfe/pull/502#issue-1819029716.

openfecli/tests/commands/test_gather.py
```diff
@@ -4,79 +4,105 @@
...
 def results_dir(tmpdir):
     with tmpdir.as_cwd():
-        with resources.path('openfecli.tests.data', 'results.tar.gz') as f:
+        with resources.path('openfecli.tests.data', 'rbfe_results.tar.gz') as f:
             t = tarfile.open(f, mode='r')
             t.extractall('.')
 
         yield
 
+_EXPECTED_DG = b"""
```
richardjgowers, line 36: @hannahbaumann are you happy with how these three outputs look?

hannahbaumann: Yes, that looks good to me!

richardjgowers:
@dwhswenson re: there's no ligand at DG=0.0, I think that's an error in the cinnabar MLE docs, I think it returns values with a mean of 0.0, rather than one ligand arbitrarily being 0.0

PR #498, opened 2023-07-19
fix issue with new notebook version

mikemhenry:

### DIFF
--- a/devtools/installer/construct.yaml
+++ b/devtools/installer/construct.yaml
@@ -14,7 +14,7 @@ specs:
   - conda
   - jupyterlab
   - mamba
-  - notebook
+  - notebook <7
   - openfe=={{ environ["VERSION"] }}
   - pip
   - pytest


mikemhenry:
Got this error in CI:
`/home/runner/openfeforge/bin/.widgetsnbextension-post-link.sh: line 1: /home/runner/openfeforge/bin/jupyter-nbextension: No such file or directory` 

It was working last week, looks like there was a bump in the major version of `notebook` so that may be the cause, still investigating

mikemhenry:
https://github.com/OpenFreeEnergy/openfe/actions/runs/5532705672

Looks like it worked!

mikemhenry:
This one is ready for review!

dwhswenson (approved):
Looks good as a band-aid, but:

1. Is this something that is going to bite us elsewhere? (Thinking ExampleNotebooks in particular).
2. From pinning, I assume this means something in our stack needs to update to meet to new major version changes. Do we know who we're waiting on?

mikemhenry:
The error I got was:
`/home/runner/openfeforge/bin/.widgetsnbextension-post-link.sh: line 1: /home/runner/openfeforge/bin/jupyter-nbextension: No such file or directory` 

But the actual `jupyter-nbextension` package was the same in the CI that passed last week and the one the failed today. So I agree it is a band-aid and I really don't have a good answer to your questions. The good news :tm: is that with the way the installer works, since it gets tested, we know that it will work if someone uses it.

Issue #464, opened 2023-06-19
Create copy of release assets (installers) without version name


dwhswenson:
This is how miniforge does it: two copies of the asset; one with version name, one without.. Note that the link https://github.com/OpenFreeEnergy/openfe/releases/latest/download/OpenFEforge-0.10.1-Linux-x86_64.sh already works (as long as 0.10.1 is latest); if we can avoid including the version name in the artifact, we can have a URL that never changes in the docs.

mikemhenry:
Solved with #497 

It isn't automated yet and that will be my next step, so we can keep this issue open

PR #503, opened 2023-07-25
remove version name and upload files

mikemhenry:

Still need to keep #464 but this will make it so our "latest" trick for the release works and now you don't need to download the assets. All that we have left to do is figure out how we want to automate this being triggered when we get a new conda-forge release


dwhswenson (approved):
LGTM. Maybe add a comment to mention that much of this is heavily influenced by the miniforge workflows? (just with the thought that if we ever have a problem in the future, we know where to go to see if they already solved the problem)

Otherwise, merge at will!

mikemhenry:
Good idea, I will add a link to the inspo

PR #505, opened 2023-07-27
Add system probe

mikemhenry:


mikemhenry:
Shout out to ChatGPT for helping me write the tests and document strings

mikemhenry:
I wasn't able to test this on a multi GPU system

mikemhenry:
Also the tests are really just to keep the coverage up, since I needed to mock it so we can run the CI anywhere. I did have it setup to catch the `no nvidia-smi` error, but then if we ran the CI on a system with a GPU, that error wouldn't be raised. So I settled on mocking the output. 

Same kind of story for the hostname, can't really hard code something and anything I used to ge the host name (like `socket.gethostname`) is just the same as the function.

mikemhenry:
@dwhswenson let me know if this covers our needs. 

I had an idea to make a `SystemProbe` object that you could then attach different `Probes` to so that it could be extended in all sorts of ways, but that felt maybe overengineered since writing a function to get the info you want, then adding it to the `_probe_system` return dict isn't too bad.

openfe/utils/system_probe.py
```diff
@@ -0,0 +1,141 @@
+import logging
+import socket
+import subprocess
+
+
+def _get_hostname() -> str:
```
dwhswenson, line 6: not sure this one-liner needs its own function

openfe/utils/system_probe.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+
+    GPU_QUERY = "--query-gpu=gpu_uuid,gpu_name,compute_mode"
+
+    try:
+        nvidia_smi_output = subprocess.check_output(
+            ["nvidia-smi", GPU_QUERY, "--format=csv"]
+        ).decode("utf-8")
+    except FileNotFoundError:
+        logging.info(
```
dwhswenson, line 73:
```suggestion
        logging.debug(
```

this is probably debug level in this context (we issue an info-level statement when we're actually running; so we don't need to say it here as well)

openfe/utils/system_probe.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Stuff info we asked for into dict with UUID as key
+        for i in range(1, len(header)):
+            field_name = header[i].strip()
+            gpu_info[gpu_uuid][field_name] = data[i].strip()
+
+    return gpu_info
+
+
+def _probe_system() -> dict:
```
dwhswenson, line 99: seems like this should be a public method?

dwhswenson (changes requested):
1. Can we reformat this to max 88 chars (preferably PEP8's 79)? The wrap length here is very inconsistent.
2. I'll probably go through and make a couple minor changes in the tests. Basically, we should have one patch function that covers all the patching needs, which we reuse, and I'd rather use actual `Mock`s instead of mock functions.
3. We might as well include a few other items here. In particular, I'd like some memory info: soft/hard limits from `resource.getrlimit(resource.RLIMIT_AS)`, to start. If we don't mind making `psutil` a requirement (it is already in our stack via `openff-toolkit -> notebook -> ipykernel`), then I'd like total and available from `psutil.virtual_memory()`.

openfe/utils/system_probe.py
mikemhenry: I didn't know if we wanted to commit it to our public API at this stage, is this something that actually even belongs in opefe or should put this whole thing into a `gufe.util` thing so any protocol writer (or perhaps gufe by default) has the ability to use it.

openfe/utils/system_probe.py
mikemhenry: Mostly did this so if we decided to hash it, or later change it to `socket.getfqdn` we could more easily

mikemhenry:
> Can we reformat this to max 88 chars (preferably PEP8's 79)? The wrap length here is very inconsistent.

:facepalm: yes of course, I ran it though black but I wasn't even thinking about the doc strings, sorry about that! 

> I'll probably go through and make a couple minor changes in the tests. Basically, we should have one patch function that covers all the patching needs, which we reuse, and I'd rather use actual Mocks instead of mock functions.

That would be great, I'd like to see how that is done, I wasn't sure if any of these mocks should be fixtures as well.

> We might as well include a few other items here. In particular, I'd like some memory info: soft/hard limits from resource.getrlimit(resource.RLIMIT_AS), to start. If we don't mind making psutil a requirement (it is already in our stack via openff-toolkit -> notebook -> ipykernel), then I'd like total and available from psutil.virtual_memory().

I focused on just what we talked about because it is easy to make this an endless list. I was thinking `psutil` as well but didn't want to add the dep but it sounds like it is in our stack (should have checked that). While we are on the subject of things to add then:

* GPU Memory
* Resource limits 
* GPU temp/power usage (could help to see if we are running into hardware or over subscribed GPU) 
* GPU Driver version
* Cudatoolkit version
* Free disk space (could help catch that issue)
* Free system memory (would help catch (maybe) if we get hit with the OOM killer)

Thoughts? None of those things are expensive to check/get and would help us figure out if there is an issue. I also don't think any of that info is sensitive.

mikemhenry:
Also thanks for the review! I always learn a lot from these.

mikemhenry:
Before I add tests for everything and doc strings,
@IAlibay @dwhswenson 

Are we happy with what this grabs? Anything missing?

dwhswenson:
I've had some weird experiences with `df`; enough that I'm not sure I always trust it to be authoritative (on shared systems, it was sometimes inconsistent on what was mounted; on my local, it splits one physical drive into a bunch of virtuals, all with 1TB, so some reporting like 1% usage on a disk that is actually 75% full). But seems like an okay place to start. The rest looks like a reasonable list to me!

mikemhenry:
Yes, `df` can have issues... kinda big rabbit hole for how to figure out the free disk space as you get into "exotic" file systems. I went with `df` since it will be fast and hopefully enough to show when a node's `/tmp` is at `100%` and giving us issues. 

In a perfect world, we would need to know all the partitions/mount points where we are doing file IO, then use the utility from that file system(s) to get free disk space

openfe/utils/system_probe.py
```diff
@@ -0,0 +1,124 @@
...
+import os
+import socket
+import subprocess
+
+import psutil
+
+
+def _get_disk_usage() -> dict[str, dict[str, str]]:
+
+    output = subprocess.check_output(["df", "-h"]).decode("utf-8")
```
IAlibay, line 11: Forgive my naive question - if we 're already using psutil, is there a reason we can't just call `psutil.disk_usage` and divide things up as necessary?

Edit: or even `shutil`!

IAlibay (commented):
Just the one thing from me otherwise lgtm!

openfe/utils/system_probe.py
mikemhenry: This is a good question, but all of those require a path, and we don't necessarily know where these files will be written, but df will show disk usage at all mount points

mikemhenry:
This got a little tricky since osx and linux don't have the same attributes for different psutil classes (which makes sense), so to get the tests to pass on OSX some other bits needed to be mocked. This did catch a bug where system_probe would throw an error on osx, which now it shouldn't.

IAlibay:
For an initial implementation, I wouldn't be mad if we just recognised osx and went "nope". After all no one is running osx on compute.

dwhswenson:
> For an initial implementation, I wouldn't be mad if we just recognised osx and went "nope". After all no one is running osx on compute.

I'd like a little osx support, so that I can at least test my follow-up PR locally.

mikemhenry:
As is, running on OSX should work the same except one key will be missing from the dictionary for the rlimit

mikemhenry:
Still getting 
`E           AttributeError: <class 'psutil.Process'> does not have the attribute 'rlimit'`

But I thought the `create=True` here https://github.com/OpenFreeEnergy/openfe/pull/505/files#diff-4286617d23538f23239369cc84f13b9f77577a53d3791d3b6663fa2faba11497R73 would make that attribute when we patch it https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

I can also mark the mocking tests to skip on OSX, but still run the smoke test since that will make sure the system probe works on OSX without having to solve the puzzle of how to get it to mock on OSX for testing

mikemhenry:
I'll just skip those tests on OSX but keep the smoke test (which should work, if it fails then I will fix it)

mikemhenry:
Okay @dwhswenson this one is ready for review now!

openfe/tests/utils/test_system_probe.py
```diff
@@ -0,0 +1,319 @@
...
+    # single patch to fully patch the system
+    patch_hostname = patch("socket.gethostname", Mock(return_value="mock-hostname"))
+
+    patch_psutil_Process_as_dict = patch(
+        "psutil.Process.as_dict",
+        Mock(
+            return_value={
+                "pid": 1590579,
+                "status": "running",
+                "exe": "/home/winry/micromamba/envs/openfe/bin/python3.10",
```
dwhswenson, line 52: I was expecting `/home/mmh/` here... who's this `winry`? 🤔

dwhswenson (approved):
Tiny suggestion for an improvement on the `__main__`; otherwise looks good to me. I suspect we'll iterate a bit on the contents and structure of these dicts, so I'm fine leaving all of this as hidden API. The logging function I write that uses this will be public API (but logging output is, IMO, never guaranteed under SemVer, so putting it in public API only promises that a function of that name exists.)

openfe/tests/utils/test_system_probe.py
mikemhenry: hehe, I cleaned it out and thought I would put a reference to something :smile:

mikemhenry:
>The logging function I write that uses this will be public API (but logging output is, IMO, never guaranteed under SemVer, so putting it in public API only promises that a function of that name exists.)

Sounds good! I figured we should let the dust settle before it gets into our public API (and decide if it it is under `openfe.utils.system_probe` or something we want to pull into `opnfe.system_probe`)

mikemhenry:
I was using pretty print for a while, but got rid of it when it sent me down this rabbit hole
image


I had thought I messed up the parsing of the GPU card name since it came out like:
```
'name': 'NVIDIA '
'GeForce '
'RTX '
'2060',
```
But pprint split that string into seperate lines at the ` `

Issue #507, opened 2023-08-01
Tests in test_openmm_equil_rfe_protocols.py fail with pytest=7.4.0


hannahbaumann:
I created a new conda env from the environment.yml file and got mulitple errors in the test_openmm_equil_rfe_protocols.py, e.g.

file /Users/hannahbaumann/openfe/openfe/tests/protocols/test_openmm_equil_rfe_protocols.py, line 38
  def test_append_topology(benzene_complex_system, toluene_complex_system):
file /Users/hannahbaumann/openfe/openfe/tests/protocols/conftest.py, line 29
  @pytest.fixture
  def benzene_complex_system(benzene_modifications, T4_protein_component):
E       fixture 'benzene_modifications' not found

Downgrading pytest to 7.3.1 seems to fix the issue.

IAlibay:
Yeah I think it was OpenFF TK and then by correspondence ambertools. Should be ok to upgrade now.

IAlibay:
So interestingly py311 is working fine on CI, but I am also about to replicate this 7.4.0 issue on a clean install.

dwhswenson:
1. Is it only with py3.11 and pytest 7.4.0?
2. Is it only for developer installs?
3. What happens if you take the session scoping off of the `benzene_modifications` fixture?

IAlibay:
Only tried on developer installs with 7.4.0 + py3.11, I'll see if I can reproduce later.

My issue was with another fixture (not benzene_modifications, but I didn't note down exactly which one).

PR #508, opened 2023-08-01
see if py311 works okay

mikemhenry:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,14 +28,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu', 'macos']
+        os: ['ubuntu']
         python-version:
           - "3.9"
           - "3.10"
           - "3.11"
-        exclude:
+        include:
           - os: 'macos'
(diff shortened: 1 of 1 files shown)


mikemhenry:
Inspired by https://github.com/OpenFreeEnergy/openfe/issues/507

dwhswenson (approved):
Looks like 3.11 works. I vote to include it in our testing matrix. Also means that wasn't the problem in #507.

mikemhenry:
Okay if we only want to run a job on OSX, which python version do we want it to be? I can do that on this PR.

mikemhenry:
Maybe just the latest? So python 3.11 on osx but skip the others? Or do we want to keep the oldest and the newest and just skip 3.10?

IAlibay:
> Maybe just the latest? So python 3.11 on osx but skip the others? Or do we want to keep the oldest and the newest and just skip 3.10?

3.11 only is fine, that's what folks will pick up on a clean install.

IAlibay (approved):

mikemhenry:
Got all the branch checks setup now as well

PR #509, opened 2023-08-01
[WIP] Add flaky markers to flaky tests on macOS GitHub actions runners

dwhswenson:
This PR will add flaky markers to tests that we see being flaky on macOS in GitHub action. I'm leaning toward just keeping this PR going for a bit (I expect there's more than one flaky test in this part of the code; I just haven't tracked which ones tend to be problematic). However, if people really want to have a separate PR for each one, we can do that too.


IAlibay:
@dwhswenson this has sat around for a while and it doesn't seem like we've been suffering from the issue anymore, do you mind if I close this PR?

richardjgowers:
I think this flaky marker doesn't work if the runners are crashing anyway(?)

PR #512, opened 2023-08-02
Migrate common settings to omm_utils

IAlibay:
This doesn't really do much but both @hannahbaumann and I are working on branches that rely on this based on some initial design on what's reusable in #237.

Since things are moving on the settings side of things, I thought it'd be best to just do this in a separate commit so that it's easier to deal with merge conflicts later.


hannahbaumann (approved):

IAlibay:
> > is there a reason the key should change if the import location changes?
> 
> Yep. Key for a `Protocol` comes from the hash of the JSON dump of the serialization of settings. In order to deserialize settings, the serialized form needs to know where we import the settings from. So moving things around is expected to break keys. Aliasing (via imports) will allow backward compatibility, but will still lead to a different key.

Do we think this is a blocker or as long as the settings themselves don't change we're still backwards compatible on deserialisation?

Issue #510, opened 2023-08-01
Generate network from indices will create non connected graphs if passed extra ligands


IAlibay:
In hindsight this makes sense, but it's definitely not a behaviour that users will be expecting.

I would expect folks to at least get a warning about this? Probably even an error.

IAlibay:
Personally I think that if you're passing in a list of edges to a method that is meant to then turn them into a network, you probably should say "hey buddy, you sure about this?".

I.e. the expectation is that this method takes in some kind of pre-defined network, I'd probably even expect to have a way to represent disconnected nodes in the list?

dwhswenson:
I think a warning is reasonable here; probably not an error (since something that takes us from `LigandNetwork` to `AlchemicalNetwork` can recover at that point.) That said, we error on MST if a node is disconnected (IMO, that one should probably error by default, but maybe give an option to quiet that into a warning. Same might be work here.)

richardjgowers:
probably a warning, I made the connected check here: https://github.com/OpenFreeEnergy/gufe/pull/214 then adding the warning into these generator methods is trivial

PR #514, opened 2023-08-04
network_from_indices raises warning on disconnected network

richardjgowers:
fixes issue #510

removes LigandNetwork in favour of implementation in gufe

### DIFF
--- a/openfe/setup/ligand_network_planning.py
+++ b/openfe/setup/ligand_network_planning.py
@@ -6,6 +6,7 @@ from typing import Iterable, Callable, Optional, Union
 import itertools
 from collections import Counter
 import functools
+import warnings
 
 import networkx as nx
 from tqdm.auto import tqdm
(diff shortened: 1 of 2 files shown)


IAlibay:
Did we not fully switch to using `gufe`'s LigandNetwork? https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/ligand_network.py

richardjgowers:
looks like we forgot to remove it from here when we migrated it

IAlibay:
Huh, did graphml update and now things are zero indexed?

IAlibay:
pinging @dwhswenson on graphml failures here

dwhswenson:
It looks to me like changes to the `LigandNetwork` in https://github.com/OpenFreeEnergy/gufe/pull/195 affected order of resulting sets in serialization. The template files in this repository are no longer accurate. Note: Among those changes, we no longer include version of the software that generated it in the GraphML, which would also cause a problem here if the order had not.

You can fix this by either:

1. Updating the template files by running the `tests/dev/serialization_test_templates.py` script and moving the output `network_template.graphml` to the `tests/data/serialization` directory; or
2. Removing the entirety of the `tests/setup/test_network.py` file -- it should have already been transferred over to gufe, right? There's no need to keep 2 copies of the tests. While you're at it, remove the template GraphML file, and the parts of the code for the `serialization_test_templates`dealing with that.

IMO, reducing duplication (choice 2) is probably the better path.

PR #521, opened 2023-08-15
Remove softcore electrostatics entry

IAlibay:
This is a source of much confusion, but as far as I can tell we don't softcore our electrostatics, nor do we have the ability to do so easily in OpenMM (at least there's no way to change the functional form for PME).

### DIFF
--- a/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
+++ b/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
@@ -86,9 +86,6 @@ class HybridTopologyFactory:
                  softcore_alpha=0.5,
                  softcore_LJ_v2=True,
                  softcore_LJ_v2_alpha=0.85,
-                 softcore_electrostatics=True,
-                 softcore_electrostatics_alpha=0.3,
-                 softcore_sigma_Q=1.0,
                  interpolate_old_and_new_14s=False,
                  flatten_torsions=False,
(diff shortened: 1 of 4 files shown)


richardjgowers (approved):

openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -172,14 +163,6 @@ def __init__(self,
             self._check_bounds(softcore_LJ_v2_alpha, "softcore_LJ_v2_alpha")
             self._softcore_LJ_v2_alpha = softcore_LJ_v2_alpha
 
-        self._softcore_electrostatics = softcore_electrostatics
-        if self._softcore_electrostatics:
-            self._softcore_electrostatics_alpha = softcore_electrostatics_alpha
-            self._check_bounds(softcore_electrostatics_alpha,
-                               "softcore_electrostatics_alpha")
-            self._softcore_sigma_Q = softcore_sigma_Q
-            self._check_bounds(softcore_sigma_Q, "softcore_sigma_Q")
```
richardjgowers, line 181: so these attributes were never used?

IAlibay: As far as I can tell, I did some digging and there's no way to modulate the nonbondedforce either.

PR #531, opened 2023-08-24
[WIP] Analysing free energies

IAlibay:
Here's a notebook that does most analyses we said we were interested in: https://gist.github.com/IAlibay/ab621caa349cdd32d8dbb4b59d37f0dd

I've added here a set of extra methods to get the relevant information to make this. The main thing that's currently missing is the correct method to get the subsampled data for the forward and reverse analysis (I'm not exactly sure how folks wanted to access that, i.e. if just having some kind of "get free energy with this truncation of samples" was better, etc...).

Feel free to jump in here and add / fix things (I've not added tests since I couldn't work out if we should store a full trajectory here or not).

### DIFF
--- a/openfe/protocols/openmm_rfe/equil_rfe_methods.py
+++ b/openfe/protocols/openmm_rfe/equil_rfe_methods.py
@@ -175,6 +175,151 @@ class RelativeHybridTopologyProtocolResult(gufe.ProtocolResult):
 
         return np.std(vals) * u
 
+    def get_overlap_matrices(self):
+        """
+        Return estimates of the overlap matrix between the states
+        for all repeats.
+
+        Returns
+        -------
+        matrices : List[npt.NDArray]
+          List of overlap matrices for each repeat
+        eigenvalues : List[npt.NDArray]
+          List of sorted eigenvalues of the overlap matrix of each repeat
+        scalar : List[npt.NDAarray]
+          List of scalars (one minus the largest nontrivial eigenvalue)
+          for each matrix
+        """
+        # Loop through and get the repeats and get the matrices
+        matrices = []
+        eigenvalues = []
+        scalar = []
+        for pus in self.data.values():
+            nc = pus[0].outputs['nc']
+            chk = pus[0].outputs['last_checkpoint']
+            reporter = multistate.MultiStateReporter(
+                      storage=nc, checkpoint_storage=chk
+                    )
+            analyzer = multistate.MultiStateSaplerAnalyzer(reporter)
+            # Note this might not work for pymbar 4
+            overlap = analyzer.mbar.computeOverlap()
+            matrices.append(overlap['matrix'])
+            eigenvalues.append(overlap['eigenvalues'])
+            scalar.append(overlap['scalar'])
+            analyzer.clear()
+
+        return matrices, eigenvalues, scalar
+
+    def get_transition_matrices(self):
+        """
+        Returns the transition matrices of replica mixing for all repeats.
+
+        Returns
+        -------
+        matrices : List[npt.NDArray]
+          List of transition matrices for each repeat
+        eigenvalues : List[npt.NDArray]
+          List of sorted eigenvalues of the transition matrix of each repeat
+        """
+        matrices = []
+        eigenvalues = []
+        for pus in self.data.values():
+            nc = pus[0].outputs['nc']
+            chk = pus[0].outputs['last_checkpoint']
+            reporter = multistate.MultiStateReporter(
+                      storage=nc, checkpoint_storage=chk
+                    )
+            analyzer = multistate.MultiStateSaplerAnalyzer(reporter)
+            mixing_stats = analyzer.generate_mixing_statistics()
+            matrices.append(mixing_stats.transition_matrix)
+            eigenvalues.append(mixing_stats.eigenvalues)
+            analyzer.clear()
+        return matrices, eigenvalues
+
+    def get_replica_states(self):
+        """
+        Returns the timeseries of replica states for each repeat.
+
+        Returns
+        -------
+        replica_states : List[npt.NDArray]
+          List of replica states for each repeat
+        """
+        replica_states = []
+        for pus in self.data.values():
+            nc = pus[0].outputs['nc']
+            chk = pus[0].outputs['last_checkpoint']
+            reporter = multistate.MultiStateReporter(
+                      storage=nc, checkpoint_storage=chk
+                    )
+            states = reporter.read_replica_thermodynamic_states()
+            replica_states.append(states)
+
+        return replica_states
+
+    def equilibration_iterations(self):
+        """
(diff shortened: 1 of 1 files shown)


openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -0,0 +1,619 @@
...
+import warnings
+
+import matplotlib.pyplot as plt
+import numpy as np
+import numpy.typing as npt
+from openmmtools import multistate
+from openff.units import unit, ensure_quantity
+
+
+class MultistateEquilFEAnalysis:
```
richardjgowers, line 17: I'm not really a fan of another class.  Anything class-like should probably get rolled into the existing `ProtocolUnitResult`, anything that is useful for working with .nc file's energies should just be a function.

openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -0,0 +1,619 @@
...
+            DF_ij = r['Delta_f']
+            dDF_ij = r['dDelta_f']
+
+        DG = DF_ij[0, -1] * analyzer.kT
+        dDG = dDF_ij[0, -1] * analyzer.kT
+
+        return (ensure_quantity(DG, 'openff').to(return_units),
+                ensure_quantity(dDG, 'openff').to(return_units))
+
+    def get_equil_free_energy(self) -> tuple[unit.Quantity, unit.Quantity]:
```
richardjgowers, line 198: how is this different from here: https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/protocols/openmm_rfe/equil_rfe_methods.py#L646

openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -0,0 +1,619 @@
...
+            '_equilibration_iterations': self.equilibration_iterations,
+            '_replica_states': self.replica_states}
+
+        if hasattr(self, '_exchange_matrix'):
+            results_dict['_replica_exchange_statistics'] = self.replica_exchange_statistics
+
+        return results_dict
+
+    def __del__(self):
+        self.close()
```
richardjgowers, line 419: this isn't required, the analyzer already does this

richardjgowers (changes requested):

openfe/protocols/openmm_utils/multistate_analysis.py
IAlibay: The way the method works in openmmtools currently is too restrictive, this is kind of a placeholder to eventually pass through bootstrapping etc.. as options.

openfe/protocols/openmm_utils/multistate_analysis.py
IAlibay: removed

openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -0,0 +1,619 @@
...
+    reporter : openmmtools.MultiStateReporter
+      Reporter for the MultiStateSampler
+    sampling_method : str
+      The sampling method. Expected values are `repex`, `sams`,
+      and `independent`.
+    result_units : unit.Quantity
+      Units to report results in.
+    forward_reverse_samples : int
+      The number of samples to use in the foward and reverse analysis
+      of the free energies. Default 10.
```
richardjgowers, line 40: this signature is confusing, who is supposed to use this class and when is it designed to be used?  Is this during a simulation or after a simulation?  Is this internal code or using facing?  It seems confused between the two.

Rather than accepting a reporter, can we just request the `.nc` path (which we can trivially create the Reporter from).

Having forward/reverse samples defined here doesn't make sense, surely that's an argument to the related method?

openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -0,0 +1,619 @@
...
+        """
+        if hasattr(self, '_exchange_matrix'):
+            return self._exchange_matrix
+        else:
+            errmsg = ("Exchange matrix was not generated, this is likely "
+                      f"{self.sampling_method} is not repex.")
+            raise ValueError(errmsg)
+
+    @property
+    def _unit_results_dict(self):
```
richardjgowers, line 403: why is this a private method which is required elsewhere?  Why are some keys "private"?  It just feels like we're trying to make this hard to use

IAlibay: The private keys is because @dwhswenson asked for it and I didn't want to cycle on that conversation and avoid merging anytime this side of the year because I somehow introduced something that broke the "public facing" unit result.

Re: private method, we can make it public - I don't really care, I did it because I don't expect anything else but the units to use this.

openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -0,0 +1,619 @@
...
+
+        if sampling_method.lower() not in ['repex', 'sams', 'independent']:
+            wmsg = f"Unknown sampling method {sampling_method}"
+            warnings.warn(wmsg)
+        self.sampling_method = sampling_method.lower()
+
+        # Do a first pass at the analysis
+        self._analyze(forward_reverse_samples)
+
+    def plot(self, filepath: Path, filename_prefix: str):
```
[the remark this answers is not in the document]
IAlibay: Can we shelve this for "once we have all our plotting done"? I suspect a lot of this would go elsewhere eventually.

openfe/protocols/openmm_utils/multistate_analysis.py
IAlibay: > this signature is confusing, who is supposed to use this class and when is it designed to be used? Is this during a simulation or after a simulation? Is this internal code or using facing? It seems confused between the two.

Unfortunately we are trying to do both here, there's both a need to have a class that can generate all this information externally (if you just have a netcdf file & checkpoint), and there's a need to have something that you could use both during and after a simulation within a protocol unit.

I agree that it's not the most amazing API, but I also need this **yesterday**.

> Rather than accepting a reporter, can we just request the .nc path (which we can trivially create the Reporter from).

Because of the very flaky way reporters work, this would require us to both force delete caches and close the reporter twice during a protocol unit, which is a real pain. Until we have a bettter upstream way to do this, passing the reporter through here avoids _some_ mess - indeed if you look at the `_analyze` code you'll realise that stuff that _should_ just be one liners called every time you call an attributte don't happen because even then the whole thing falls apart if you query certain things too often (don't ask me why..).

> Having forward/reverse samples defined here doesn't make sense, surely that's an argument to the related method?

I don't know of any other way to pass that through to `_analyze` unless you want to just always use the default value?

richardjgowers (approved):
I've cycled this a few times doing some maintenance work and it doesn't look like it's causing failures if the pytest fixure explicitly cleans up the Reporter object (try yield, finally close pattern)

PR #545, opened 2023-09-06
fix forwards/backwards analysis for very small files

richardjgowers:
this is generally a better way to portion up the various samples for forwards/backwards anyway


IAlibay:
> @IAlibay is the forwards and backwards analysis not meant to include the 1.0 point?  it looks like the regression test doesn't anticipate this

It's not expected because as long as we get close to a 100% use of the data that's good though - we plot by fraction not by fixed values.

IAlibay (approved):
Slightly changes the behaviour from what alchemlyb and alchemical analysis does but sure seems cleaner.

IAlibay:
> > @IAlibay is the forwards and backwards analysis not meant to include the 1.0 point?  it looks like the regression test doesn't anticipate this
> 
> It's not expected because as long as we get close to a 100% use of the data that's good though - we plot by fraction not by fixed values.

:/ it was meant to get closer to 1 than 90%, I guess that was a bug.

Issue #544, opened 2023-09-06
`generate_radial_network` does not check if `central_ligand` is in `ligands`


Yoshanuikabundi:
There is no check in `generate_radial_network` that the `central_ligand` is not present in `ligands`. Instead, if `ligands` includes `central_ligand`, an edge is created from `central_ligand` to `central_ligand`. This causes the resulting network to include a transformation from the central ligand to itself. This does not appear in the current visualization method `plot_atommapping_network` but could cause spurious calculations to run, I'm not sure. The edge is certainly included in `LigandNetwork.edges` and `LigandNetwork.graph.edges`.

I think this should either raise an exception or silently delete the useless edge, possibly with an optional argument to allow it, and the behavior should be clearly documented in the docstring.

richardjgowers:
Hmm probably silently ignore central if it's in the ligands. Maybe a quick warning or logger message.

mikemhenry:
+1 to delete/ignore it, throw a warning message

PR #546, opened 2023-09-06
don't include central ligand self edge in radial networks

richardjgowers:
fixes issue #544


openfe/setup/ligand_network_planning.py
```diff
@@ -84,6 +85,8 @@ def generate_radial_network(
     edges = []
 
     for ligand in ligands:
+        if ligand == central_ligand:
```
mikemhenry, line 88: No warning? Or is this something that a user would expect we handle in this way? I think it makes sense to skip it this way the more I think about it

richardjgowers:
I documented the behaviour, so arguably its not a warning any more. It does
feel easier to not have to manually separate out the central ligand and
just chuck them in.

openfe/setup/ligand_network_planning.py
IAlibay: Probably would second that.

IAlibay (approved):
I'll approve but a warning message would be good.

hannahbaumann (approved):

PR #550, opened 2023-09-09
[WIP] Fix RTHP results serialization and test results load

IAlibay:


openfe/tests/data/openmm_rfe/vac_results.json.gz
IAlibay, line 1: I think I'm missing something really obvious. I'm trying to load in the results json back into a `RelativeHybridTopologyProtocolResult` but my unit results end up without the `.outputs` attribute.

Current testing looks something like this:

```
 r = json.load(gzip.open('vac_results.json.gz', 'r'), cls=JSON_HANDLER.decoder)
 rhtpr = RelativeHybridTopologyProtocolResult._from_dict(r['protocol_result']['data'])
```

Looks like it hasn't sufficiently deserialized the protocol unit results, is this even something we're meant to be able to do?

Loads fine, has the

richardjgowers:
looks like #551 actually needs fixing in gufe, since the specialised ProtocolResult doesn't override to/from dict

richardjgowers (approved):

openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -125,8 +125,9 @@ def _analyze(self, forward_reverse_samples: int):
         """
         # Do things that get badly cached later
         self._replica_states = self.analyzer.reporter.read_replica_thermodynamic_states()
-        self._equil_iters = self.analyzer.n_equilibration_iterations
-        self._prod_iters = self.analyzer._equilibration_data[2]
+        # float conversions to avoid having to deal with numpy dtype serialization
+        self._equil_iters = float(self.analyzer.n_equilibration_iterations)
```
dwhswenson, line 129: should these be `int` instead of `float`? (Variable name looks to me like an `int`)

IAlibay: I've already raised that issue in openmmtools, currently they return float and it's dumb.

PR #560, opened 2023-09-15
Tests: Test coverage for RFE ProtocolResult

richardjgowers:


openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -1241,10 +1241,101 @@ def test_constraints(tyk2_xml, tyk2_reference_xml):
...
+        pr = openmm_rfe.RelativeHybridTopologyProtocolResult.from_dict(d['protocol_result'])
+
+        assert pr
+
+    def test_get_estimate(self, protocolresult):
+        est = protocolresult.get_estimate()
+
+        assert est
+        assert isinstance(est, unit.Quantity)
+        assert est.is_compatible_with(unit.kilojoule_per_mole)
```
IAlibay, line 1267: Could we assert the value for the estimate & uncertainty? (I'm assuming this should remain static unless something goes terribly wrong)

IAlibay (approved):
Just the one thing, otherwise lgtm!

richardjgowers:
the test errors look like this fix here: https://github.com/OpenFreeEnergy/gufe/pull/236

dwhswenson (approved):
lgtm -- it appears @IAlibay's concern has been addressed?

Only thing I'd suggest changing is something that happened in a previous PR: fixture `transformation_json`; from the name, I would expect that to be the JSON of a `Transformation`, not of the results object. Might rename while we're thinking about it?

PR #561, opened 2023-09-19
Combine installation guides and fix typos

Yoshanuikabundi:
This combines the HPC install guide from the manual into the installation guide, and tidies up some typos in the installation guide.

The HPC guide uses Micromamba, while the rest of the guide recommends Mamba. @dwhswenson mentioned that this is because Micromamba [doesn't yet] support setting environment variables by environment. I'm not sure that we recommend this anywhere, but I do think micromamba simplifies installation for most users. So we should probably pick one and adapt the appropriate half of the installation guide to it.


mikemhenry:
@Yoshanuikabundi can I push directly into this PR or would you like me to make a PR for your `recombinant_install_guide` branch? I can switch to using `mamba` for everything, but maybe I can make an argument that at the start of the HPC section we have a "Note" thing where we explain why we recommend micromamba for HPC deployment? This part of the guide is really only relevant for HPC sysadmins or users who manage their own HPC environments.

dwhswenson:
> This part of the guide is really only relevant for HPC sysadmins or users who manage their own HPC environments.

Gotta disagree, unless by "Users who manage their own HPC environments," you mean "any user who wants to try this out on a cluster where it isn't already installed" -- which is pretty much everyone who we haven't personally walked through the install process. This is absolutely aimed at our first-time users, so I think introducing both mambaforge and micromamba will be a source of confusion.

Part of the reason we're moving this to make it more prominent is that we have gotten questions from partners several times where our answer is a link to this part of the docs. You wrote a clear answer here; we need to make it more findable, and probably best to avoid mixing mamba/micromamba.

mikemhenry:
Oooops I forgot about this! @Yoshanuikabundi Can I push directly onto this branch to update the output and stuff to use `mamba` instead of `micromamba`?

> Gotta disagree, unless by "Users who manage their own HPC environments," you mean "any user who wants to try this out on a cluster where it isn't already installed"

That is what I mean. I am still not convinced that micromamba isn't the right choice, HPC or not, but I can update the docs to be consistent.

dwhswenson:
@mikemhenry : I discussed with @Yoshanuikabundi last night -- yes, please feel free to push directly to this branch (that way we don't forget to make this update (don't want to hold the PR too long waiting for that).

mikemhenry:
Will do!

mikemhenry:
@dwhswenson @Yoshanuikabundi I have made a commit that updates the guide, let me know if you have any questions and feel free to change things as you see fit :smile:

Yoshanuikabundi:
Looks great @mikemhenry, Thanks!

dwhswenson (approved):
Sorry, thought I'd already approved this! LGTM

PR #563, opened 2023-09-20
Switch to a unified Sphinx theme

Yoshanuikabundi:
This PR switches the CSS theme to https://github.com/OpenFreeEnergy/ofe_sphinx_theme. The theme just basically provides some CSS and then falls back to the PyData theme, but it means that we can centralise CSS from this repository and then use those styles everywhere.

I expect it to build on RTD with no problems but I can't wait to see that I'm wrong about that!


mikemhenry:
IDK if this is the right place to raise this (point me in right direction if it isn't) but with dark mode the button to hide search matches needs a darker font IMHO 
image

mikemhenry:
The code blocks look great, that was the only other thing that wasn't working in previous versions of the css (with dark mode)

Yoshanuikabundi:
Thanks for catching that @mikemhenry! Looks like (from searching the PyData theme source) that's the only place that uses `--pst-color-primary-text`, so I've just changed that to be more constrasty. It seems to mean "the text color for objects that are colored the primary color", rather than the more obvious "the primary text color".

richardjgowers:
@Yoshanuikabundi is the pirate-y font on the base repo?  Otherwise looks good.

richardjgowers (approved):

mikemhenry:
Hmmm, mypy passed on the PR but now is failing on main

Yoshanuikabundi:
@richardjgowers @dwhswenson Pirate is a term of endearment right?

I thought it was a cute way to tie in the alchemical theme, but I'm not particularly attached to it no matter how much I like serifs. It's specified here, in the OpenFE theme: https://github.com/OpenFreeEnergy/ofe-sphinx-theme/blob/main/ofe_sphinx_theme/theme/ofe_sphinx_theme/sass/_fonts.scss

Changing it there will affect all projects using the theme from the next time they build the documentation. The commented out EB Garamond is a similar vintage/style font but doesn't look like it's gone through a printing press... which may be better or worse. Commenting out both lines (7 and 8) should fall back to the same font that's used for the body.

Yoshanuikabundi:
@mikemhenry I have no idea how this change would affect typing and it [looks like](https://github.com/OpenFreeEnergy/openfe/actions/workflows/mypy.yaml) mypy is failing in all other PRs from the last two days or so too.

PR #569, opened 2023-09-29
Charge correction via alchemical waters

IAlibay:


IAlibay:
no idea what's going on with codecov.. it's pointing to stuff not being covered that should

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -77,6 +78,38 @@ class Config:
...
+    The minimum distance from the system solutes from which an
+    alchemical water can be chosen. Default 0.8 * unit.nanometer.
+    """
+    alchemical_water_ion_positive_resname = 'NA'
+    """
+    Name of the positive ion used to replace the alchemical water with
+    if ``explicit_charge_correction`` is ``True`` and the absolute charge
+    difference is negative.
+    """
+    alchemical_water_ion_negative_resname = 'CL'
```
richardjgowers, line 101: Can these be pulled automatically from the SolventComponent solvation settings?

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -77,6 +78,38 @@ class Config:
...
+    """
+    Name of the negative ion used to replace the alchemical water if
+    ``explicit_charge_correction`` is ``True`` and the absolute charge
+    difference is positive.
+    """
+    alchemical_water_resname = 'HOH'
```
richardjgowers, line 107: this is probably not a real setting, isn't this just the resname that `add_solvent()` adds?  (i.e. it's know a priori and can be hard coded?)

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -67,9 +68,88 @@ def _get_resname(off_mol) -> str:
...
 
+def _get_alchemical_charge_difference(
+    mapping: LigandAtomMapping,
+    nonbonded_method: str,
+    explicit_charge_correction: bool,
+    positive_ion_resname: str = 'NA',
+    negative_ion_resname: str = 'CL',
+) -> int:
+    """
+    Checks and returns the difference in formal charge between
```
richardjgowers, line 79: these first lines in docstrings are meant to be a one liner.  So `Difference in formal charge between state A and B` gives the same information

openfe/protocols/openmm_rfe/equil_rfe_settings.py
IAlibay: Todo - I need to double check the FF files to see if the resname is inherited or set by modeller.


However! This does bring up an interesting thought, this won't work with 4+ site waters because we don't handle vsite transformations. I need to add in a guard & some docs for it.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
IAlibay: done

IAlibay:
@richardjgowers - so looks like we can assume the residue name _if_ we assume we're picking up auto-generated solvent.

I'll switch the code out for that, however in the longer term we should keep an eye out on if users might want to manually define an ion / water in their PDB that they want to be the alchemical water.

openfe/protocols/openmm_rfe/equil_rfe_settings.py
IAlibay: done

openfe/protocols/openmm_rfe/equil_rfe_settings.py
IAlibay: done on both counts

richardjgowers (approved):

IAlibay:
To update on the current status of testing for this:

@hannahbaumann is leading the validation for this, she is looking into some of the validation tests that were done here: https://pubs.acs.org/doi/abs/10.1021/acs.jctc.8b00825

richardjgowers:
@IAlibay can we merge this as a beta version, then potentially patch it?  I'd rather that than have to keep rebasing merge conflicts around it

IAlibay:
> @IAlibay can we merge this as a beta version, then potentially patch it? I'd rather that than have to keep rebasing merge conflicts around it

I assume that's fine, but would defer to @hannahbaumann here - do you have any thoughts on this?

hannahbaumann:
Sounds good to me! We may want to change the default number of lambda windows for charge changing perturbations before other people are using it, but that can also happen after the merge.

IAlibay:
> Sounds good to me! We may want to change the default number of lambda windows for charge changing perturbations before other people are using it, but that can also happen after the merge.

I've raised #609 so we can discuss this - probably a separate PR would be better.

IAlibay:
@richardjgowers as-is the correction is applied by default, if it's a beta feature would you agree to setting https://github.com/OpenFreeEnergy/openfe/pull/569/files#diff-74b77d39f244b9ced3853d19bef1c2f9b4ba5ba65064af79ef5daa2a188d1bd9R81 to `False`?

IAlibay:
I've made the change for now to see how much of CI is going to break..

PR #577, opened 2023-10-13
rfe protocol: partial charges fix

### LINKED ISSUE #576: simulations fail on failed conformer generation
Some simulations fail if conformer generation failed as part of partial charge assignment.  General traceback looks like:

```python
ValueError: No registered toolkits can provide the capability "assign_partial_charges" for args "()" and kwargs "{'molecule': Molecule with name xx and SMILES xx

'partial_charge_method': 'am1bcc', 'use_conformers': None, 'strict_n_conformers': False, 'normalize_partial_charges': True, '_cls': <class 'openff.toolkit.topology.molecule.Molecule'>}"

Available toolkits are: [ToolkitWrapper around The RDKit version 2023.03.2, ToolkitWrapper around AmberTools version 22.0, ToolkitWrapper around Built-in Toolkit version None]

ToolkitWrapper around The RDKit version 2023.03.2 <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : partial_charge_method 'am1bcc' is not available from RDKitToolkitWrapper. Available charge methods are ['gasteiger', 'mmff94']

 ToolkitWrapper around AmberTools version 22.0 <class 'openff.toolkit.utils.exceptions.ConformerGenerationError'> : RDKit conformer generation failed.

ToolkitWrapper around Built-in Toolkit version None <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : Partial charge method "am1bcc"" is not supported by the Built-in toolkit. Available charge methods are ['zeros', 'formal_charge']
```

I've raised https://github.com/openforcefield/openff-toolkit/issues/1741 to try and fix this upstream.

I think we can probably fix this locally by manually generating partial charges, which will give us the control to use the existing conformer, then supply these recently generated partial charges as "user" partial charges, which should skip conformer generation.

richardjgowers:
fixes #576

calculate partial charges as separate step and use these as "user charges" throughout

this avoids a case where conformer generation can fail and parametrisation fails

this is technically not fully SMIRNOFF spec, but it probably isn't all that bad either


openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -538,26 +538,43 @@ def run(self, *, dry=False, verbose=True,
...
-            if comp == mapping.componentA:
-                molB = mapping.componentB.to_openff()
-                system_generator.create_system(molB.to_topology().to_openmm(),
-                                               molecules=[molB])
+        for mol in chain(off_small_mols['stateA'], off_small_mols['stateB'],
+                         *off_small_mols['both']):
+            # robustly calculate partial charges;
+            try:
+                # try and follow official spec method
+                mol.assign_partial_charges('am1bcc')
```
richardjgowers, line 564: this explicit partial charge step is probably also handy for allowing alternate methods here one day...

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -145,8 +145,8 @@ def get_omm_modeller(protein_comp: Optional[ProteinComponent],
       Protein Component, if it exists.
     solvent_comp : Optional[ProteinCompoinent]
       Solvent Component, if it exists.
-    small_mols : list[SmallMoleculeComponents]
-      List of SmallMoleculeComponents to add.
+    small_mols : list
+      List of OpenFF Molecule to add.
     omm_forcefield : app.ForceField
       ForceField object for system.
     solvent_settings : SolvationSettings
```
richardjgowers, line 152: this is very ugly currently as it's taking a mixture of gufe objects and openff.

The proteincomponent goes straight to openmm rather than having an intermediate openff object representation, so it's not clear there's an easy way to have all objects in the same "type".  system_creation is pretty internal so it's probably not something that is that bad though

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -538,26 +538,43 @@ def run(self, *, dry=False, verbose=True,
...
+        # see openfe issue #576
+        # calculate partial charges manually if not already given
+        # convert to OpenFF here,
+        # and keep the molecule around to maintain the partial charges
+        off_small_mols = {
+            'stateA': mapping.componentA.to_openff(),
+            'stateB': mapping.componentB.to_openff(),
+            'both': [m.to_openff() for m in small_mols
+                     if (m != mapping.componentA and m != mapping.componentB)]
+        }
```
richardjgowers, line 551: from playing with TemplateGenerator, it didn't seem to cache the user charges I passed through (i.e. I couldn't convert to openff, assign partial charges, parametrise here to populate cache, then later retrieve the same partial charges from the SMILES lookup.).  Therefore I've converted the small molecules to openff here, and they remain in this representation throughout. 

I could probably put this dict representation in your previous validate functions

IAlibay: > it didn't seem to cache the user charges I passed through

I'm re-reading this now - I find that super odd, I checked several times that SystemGenerator doesn't double calculate charges & will read from cache (did this by doing repeated independent executions of this code and checking for sqm calls on a big molecule).

I'm confused as to why this would be any different O.o! bug?

IAlibay (changes requested):
Mostly requesting changes on checking if charges already exist & skip calculating them (for consistency we use the same pattern as this - https://github.com/openmm/openmmforcefields/blob/main/openmmforcefields/generators/template_generators.py#L204 ... or change the gufe warnings if we find zeros acceptable.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
richardjgowers: Yeah I didn't mind the structure of AFE with each step being a separate block.  

re: the cache.  It was probably buggy behaviour for me locally, but it wasn't creating the .json file.  I think currently the use of cache is optional on our end too, so we can't rely on it either

richardjgowers:
@IAlibay re: fixing this upstream.  The answer I've got is that this pattern of creating partial charges off an existing conformer is slightly nonstandard, so likely won't get fixed upstream until the spec is clarified around this corner case: https://github.com/openforcefield/openff-toolkit/issues/1741#issuecomment-1762305888

openfe/protocols/openmm_rfe/equil_rfe_methods.py
IAlibay: forgot it was optional 😅

Erm, I'll be honest I kinda feel like we should get this out to users now and bother about cleaning up later - how would you feel about going for option 2?

richardjgowers:
Yeah so option 2 would work great

openfe/protocols/openmm_rfe/equil_rfe_methods.py
richardjgowers: added a check for existing partial charges

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -538,26 +538,46 @@ def run(self, *, dry=False, verbose=True,
...
-                                           molecules=[offmol])
-            if comp == mapping.componentA:
-                molB = mapping.componentB.to_openff()
-                system_generator.create_system(molB.to_topology().to_openmm(),
-                                               molecules=[molB])
+        for mol in chain([off_small_mols['stateA'], off_small_mols['stateB'],
+                         *off_small_mols['both']]):
+            # robustly calculate partial charges;
+            if mol.partial_charges is not None and np.any(mol.partial_charges):
+                # skip if we have existing partial charges
```
IAlibay, line 563:
```suggestion
                # skip if we have existing partial charges unless they are zero (see openmmforcefields)
```

sorry I'm on phone so terrible formatting, but something like this would be great so we remember why

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -132,7 +132,7 @@ def get_system_generator(
 
 def get_omm_modeller(protein_comp: Optional[ProteinComponent],
                      solvent_comp: Optional[SolventComponent],
-                     small_mols: list[SmallMoleculeComponent],
+                     small_mols: list,
```
IAlibay, line 135: looks like CI is unhappy with this somewhere

IAlibay (approved):
I'm going to approve because I think it's mostly all there, just whatever is breaking CI needs fixing, probably a call being missed somewhere

richardjgowers:
@IAlibay the failing tests were more annoying than that, it was that I'd changed a function that AFE was using.  I've gone through and applied this fix also to the AFE Units, in `_get_components` SMC objects are now converted to OpenFF

openfe/protocols/openmm_utils/system_creation.py
IAlibay: Oof ok, this is messier than I thought - I completely missed the fact that we're now keying a dictionary that tracks the residues of the system with a mixture of components & offmols.

That dictionary is then going everywhere in the rest of the protocol which escapes the scope of just this method, that's going to cause someone using our tooling (like Ivan) a massive headache.

IAlibay (commented):
@.@ this is getting really confusing / messy.

@richardjgowers how would you feel about turning `small_mols` into a dictionary of keys smcs and values offmols?

That way we can keep things like comp_resids purely keyed by smcs.

:/ I realise time is of the essence here, I can do a PR into this PR with the necessary changes by end of day if that's reasonable?

Otherwise if you're happy with dealing with this in bugfix, we release 0.14.0 with this, and then we do a 0.14.1 with the necessary change after & tell Ivan to skip 0.14.0?

openfe/protocols/openmm_afe/base.py
```diff
@@ -71,7 +72,7 @@ def __init__(self, *,
                  stateA: ChemicalSystem,
                  stateB: ChemicalSystem,
                  settings: settings.Settings,
-                 alchemical_components: Dict[str, List[str]],
+                 alchemical_components: dict[str, list[Component | OFFMolecule]],
```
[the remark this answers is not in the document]
richardjgowers: Yeah this is probably wrong, it took me a while to track everything down.  Init here probably wants to be gufe-land only.

richardjgowers:
@IAlibay I thought the tooling we were trying to upstream to Tools would be working in terms of OFF Molecules or OMM Topologies.  Ideally we don't include gufe in that.

IAlibay:
> @IAlibay I thought the tooling we were trying to upstream to Tools would be working in terms of OFF Molecules or OMM Topologies.  Ideally we don't include gufe in that.

There's two sets of reusable tooling - stuff we upstream (factories etc..), and stuff we share between protocols but is really specific to those protocols (say tracking residues in a system for alchemical transformations).

The latter is kind of the case here, it's useful to keep track of your gufe components in your system (that way you can keep track of which bits of the system are equal etc..), but I can't see anyone but folks developing an OpenFE protocol to need.

IAlibay:
My understanding is that FEFlow is using openmm_utils for now anyways, so if we change behaviour temporarily we should at least tell them to avoid that version.

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -132,7 +132,7 @@ def get_system_generator(
 
 def get_omm_modeller(protein_comp: Optional[ProteinComponent],
                      solvent_comp: Optional[SolventComponent],
-                     small_mols: list[SmallMoleculeComponent],
+                     small_mols: dict[SmallMoleculeComponent, OFFMol],
```
richardjgowers, line 135: @IAlibay ok it's a dict now.  This is the change to the non-Protocol code here, you're passing through pairs of SMC:OFF, the SMC for resid lookup and OFF which is carrying partial charges

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -538,26 +538,46 @@ def run(self, *, dry=False, verbose=True,
...
+
+            system_generator.create_system(mol.to_topology().to_openmm(),
+                                           molecules=[mol])
 
         # c. get OpenMM Modeller + a dictionary of resids for each component
         stateA_modeller, comp_resids = system_creation.get_omm_modeller(
             protein_comp=protein_comp,
             solvent_comp=solvent_comp,
-            small_mols=small_mols,
+            small_mols=[off_small_mols['stateA']] + off_small_mols['both'],
```
IAlibay, line 580: This expects a dictionary now.

IAlibay (approved):
Thanks @richardjgowers lgtm! There's a couple of lines in the RFE protocol that need fixing but otherwise this seems great.

Issue #579, opened 2023-10-24
cli: plan: yaml format definition


richardjgowers:
Define the yaml format that will act as input to cli planning tools

richardjgowers:
solved in #583

PR #583, opened 2023-10-24
Yaml format for CLI input

richardjgowers:
[skip ci]

addresses #579

draft of yaml format for CI

There's a couple of questions for how we structure things in here, but it's mostly been decided in various discussion


openfecli/commands/input.yaml
```diff
@@ -0,0 +1,43 @@
+# things behind a "method" key have to match up to a known thing in openfe
+
+network:
+  method: plan_radial_network
+  scorer: default_lomap_score
+
+# Question:
+# use settings here, to match Protocol
```
IAlibay, line 8: I like the "settings" nest, since a) allows for further expansion for non-settings keys if they are ever needed, b) it's just visually pleasing to look at.

Not a super strong vote though.

mikemhenry: +1 on settings nesting

dwhswenson, line 8: I honestly don't remember which approach was easier for a JSON Schema. I think the not nested version is a little easier: you define a single subschema with a literal for `method` and the appropriate list of settings keys. With nested settings, I think you have to define the subschema with the `method` literal and point another subschema for the settings. That would make documentation, e.g., as generated by `sphinx-jsonschema`, a little ugly. TBH, it is really just a question of which is easier with whatever tool we'll do to generate the JSON schema (do *not* go down the road of rolling your own).

openfecli/commands/input.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+mappers:
+  - method: LomapAtomMapper
+    settings:
+      timeout: 60
+  - method: LomapAtomMapper
+    settings:
+      timeout: 120
+
+# OR not nesting settings, anything without the "method" key gets put into kwargs
```
dwhswenson, line 18: Either one works from a JSON Schema standpoint. I'd vote to start by requiring it to be a list and add handling for non-list later, just as a matter of do the simple thing first.

openfecli/commands/input.yaml
```diff
@@ -0,0 +1,43 @@
+# things behind a "method" key have to match up to a known thing in openfe
+
+network:
+  method: plan_radial_network
```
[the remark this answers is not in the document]
richardjgowers: I think everything in the file is optional.  It's just to augment the command line `plan_rbfe_network`.

openfecli/commands/input.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - method: LomapAtomMapper
+    timeout: 60
+  - method: LomapAtomMapper
+    timeout: 120
+
+
+# Question:
+# is mapper*s* always a list and mapper always a single item
+# or do we allow mapper to define either a list or single item behind it?
+
```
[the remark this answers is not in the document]
richardjgowers: The way multiple mappers works is already define in the `plan_radial_network` etc functions. see: https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/ligand_network_planning.py#L99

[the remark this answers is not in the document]
richardjgowers: It's more aesthetically, is the spec that we allow `Union[AtomMapper, list[AtomMapper]]` to mapper?  This would match the above signature I guess

dwhswenson (commented):
To be clear: this appears to be a format only for the basic RBFE/RHFE network planners. It looks like this design will not support more complicated alchemical network planners (e.g., something that combines RBFE and ABFE). This isn't necessarily a problem, but I'm just making sure we're conscious that this is solving a subset of the problem space, and we might need to revisit it (but that's an OpenFE 3.0 problem).

My main thought to consider for the future: A user can create a JSON Schema based on their installed plugins (once we have plugins) that represents the allowed input structure. This would be useful because it will allow in-editor [autocomplete and validation](https://github.com/neoclide/coc-yaml) (that's a vim plugin, but many text editors support this). So they get a bunch of usability for free, if we can output a JSON schema. Therefore I'd recommend approaches that are more friendly to JSON schema output.

openfecli/commands/plan_network_schema.py
```diff
@@ -0,0 +1,62 @@
...
+
+"""
+
+from pydantic import BaseModel, ConfigDict
+from typing import Any, Optional
+import yaml
+import warnings
+
+
+class MapperSelection(BaseModel):
```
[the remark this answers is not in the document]
dwhswenson: I was definitely expecting to use pydantic (part of my "mistake I did in OPS CLI: I reinvented pydantic; let's not repeat that mistake").

That said, you'll need a model for, e.g, each AtomMapper, and then a separate model for `MapperSelection`, which is just `mappers: list[LomapMapperModel | KartografMapperModel | ...]`, where `LomapMapperModel` looks like what you have now. (With the nested settings, you may need yet another model for, e.g., `LomapMapperSettingsModel`; this is part of why that makes the scheme messier.)

In the later version, (1) we'd see what we can do to dynamically create the `LomapMapperModel` so devs don't have to repeat typing/docs; (2) we'd dynamically generate the `MapperSelection` model based on what plugins existed.

openfecli/parameters/plan_network_options.py
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        try:
+            cls = mapper_choices[opt.mapper.method]
+        except KeyError:
+            raise KeyError(f"Bad mapper choice: '{opt.mapper.method}'")
+
+        choices['mapper'] = cls(**opt.mapper.settings)
+    if opt.network:
+        network_choices = {
```
richardjgowers, line 99: @dwhswenson avert your eyes!  This is the hacky job I think will initially serve us.  I think having this dict-like thing being eventually much more clever is what you're envisaging, this is as far as I'm taking it.

richardjgowers:
hmm interesting, the failing test is on our pydantic 1 build.  It's the normalisation to lower case

richardjgowers:
Ok this is ready for review.

To try this out, you can create a `settings.yaml` file like below
```yaml
mapper:
  method: LomapAtomMapper
  settings:
    time: 120

network:
  method: generate_minimal_redundant_network
  settings:
    mst_num: 3
```

Then run:

```bash
openfe plan-rbfe-network -p tyk2_protein.pdb -M tyk2_ligands.sdf -s settings.yaml
```

And this should have created a simulation plan with minimal redundant network selected edges

dwhswenson (changes requested):
Needs a number of minor changes even if we want to limit it to this functionality; see individual comments. Big things that are still missing:

* A way to document the available options. This was sort of the big sticking point when we discussed this. It is not solved in any way at all here. An undocumented settings file is not particularly useful to users.
* There's an assumption here that all user-supplied values are YAML/JSON basic types. This is true for the limited subset of functionality included in this PR, but it is going to break down very quickly once we move to, say, adding support for protocol settings. From this PR, it is unclear what the intended direction forward on that is.
* There's no validation of the user input in the `settings` subdicts. This means that something that could easily be a user-understandable `TypeError` will lead to some unclear error at runtime. For example, give bad types to the LOMAP params and you end up with:

    `ArgumentError("Python argument types in\n    rdkit.Chem.rdFMCS.FindMCS(list)\ndid not match C++ signature:\n    FindMCS(boost::python::api::object mols, RDKit::PyMCSParameters {lvalue} parameters)\n    FindMCS(boost::python::api::object mols, bool maximizeBonds=True, double threshold=1.0, unsigned int timeout=3600, bool verbose=False, bool matchValences=False, bool ringMatchesRingOnly=False, bool completeRingsOnly=False, bool matchChiralTag=False, RDKit::AtomComparator atomCompare=rdkit.Chem.rdFMCS.AtomCompare.CompareElements, RDKit::BondComparator bondCompare=rdkit.Chem.rdFMCS.BondCompare.CompareOrder, RDKit::RingComparator ringCompare=rdkit.Chem.rdFMCS.RingCompare.IgnoreRingFusion, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > seedSmarts='')")`

    That will not be a useful error for users.

openfecli/commands/plan_rbfe_network.py
```diff
@@ -78,13 +78,19 @@ def plan_rbfe_network_main(
 @COFACTORS.parameter(
     multiple=True, required=False, default=None, help=COFACTORS.kwargs["help"]
 )
+@YAML_OPTIONS.parameter(
+    multiple=False, required=False, default=None,
+    help=YAML_OPTIONS.kwargs["help"],
+)
```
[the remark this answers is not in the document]
richardjgowers: I'm (we're?) not familiar with the defaults, so being explicit here is useful to know what's happening

openfecli/commands/plan_rbfe_network.py
```diff
@@ -142,27 +148,46 @@ def plan_rbfe_network(
...
+    if yaml_settings is not None:
+        yaml_options = YAML_OPTIONS.get(yaml_settings)
+        mapper_obj = yaml_options.get('mapper', None)
+        ligand_network_planner = yaml_options.get('network', None)
+
+    if mapper_obj is None:
+        mapper_obj = LomapAtomMapper(time=20, threed=True, element_change=False,
+                                     max3d=1)
+    if mapping_scorer is None:
+        mapping_scorer = default_lomap_score
```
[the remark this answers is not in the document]
richardjgowers: this is roughing out space for it in the future

openfecli/commands/plan_rbfe_network.py
```diff
@@ -142,27 +148,46 @@ def plan_rbfe_network(
...
+    if ligand_network_planner is None:
+        ligand_network_planner = generate_minimal_spanning_network
+    if solvent is None:
+        solvent = SolventComponent()
```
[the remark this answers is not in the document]
richardjgowers: roughing out space for future option here

openfecli/commands/plan_rbfe_network.py
```diff
@@ -142,27 +148,46 @@ def plan_rbfe_network(
...
+        mapper_obj = yaml_options.get('mapper', None)
+        ligand_network_planner = yaml_options.get('network', None)
+
+    if mapper_obj is None:
+        mapper_obj = LomapAtomMapper(time=20, threed=True, element_change=False,
+                                     max3d=1)
+    if mapping_scorer is None:
+        mapping_scorer = default_lomap_score
+    if ligand_network_planner is None:
+        ligand_network_planner = generate_minimal_spanning_network
```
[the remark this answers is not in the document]
richardjgowers: yep good call, I centralised this into `YAML_OPTIONS.get()`.  It reduces a lot of duplication, but means that both plan_rbfe/plan_rhfe require the same defaults (which was the case).

richardjgowers:
@dwhswenson your comments on (auto) documentation, non native yaml types and better validation/error messages are not in the MVP, so they're not going to make this PR

openfecli/parameters/plan_network_options.py
```diff
@@ -0,0 +1,179 @@
...
+      will have keys for mapper and network topology choices
+
+    Raises
+    ------
+    ValueError
+      for any malformed inputs
+    """
+    raw = yaml.safe_load(contents)
+
+    if False:
```
mikemhenry, line 73: I think this would be good to have and also helps with typos in keys

openfecli/parameters/plan_network_options.py
```diff
@@ -0,0 +1,179 @@
...
+    if opt and opt.mapper:
+        mapper_choices = {
+            'lomap': LomapAtomMapper,
+            'lomapatommapper': LomapAtomMapper,
+        }
+
+        try:
+            cls = mapper_choices[opt.mapper.method]
+        except KeyError:
+            raise KeyError(f"Bad mapper choice: '{opt.mapper.method}'")
```
mikemhenry, line 135: It would be nice to add 
```suggestion
            raise KeyError(f"Bad mapper choice: '{opt.mapper.method}' Allowed choices are {mapper_choices}")
```

Or something like that

mikemhenry:
If there is a typo/extra key added in `mapper.settings` the error message isn't awful and it looks like we just call lomap so I am not sure if we can improve the error message, but we could catch it and add a hint "check path_to_settings_file for a mistake" or something 
`TypeError: LomapAtomMapper.__init__() got an unexpected keyword argument 'shifts'`

mikemhenry:
`KeyError: "Bad mapper choice: 'kartografatommapper'"`

`KartografAtomMapper` doesn't look like it works, but in this test it looks like it is an option? https://github.com/OpenFreeEnergy/openfe/pull/583/files#diff-486ee2538adfaddbe02f50989f8ac44fd18054d3fce3cd1a8dfc9cfcf3371b9dR24

openfecli/commands/plan_rbfe_network.py
```diff
@@ -142,27 +139,29 @@ def plan_rbfe_network(
         cofactors = []
     write("\t\tCofactors: " + str(cofactors))
 
-    solvent = SolventComponent()
+    yaml_options = YAML_OPTIONS.get(yaml_settings)
```
IAlibay, line 142: Would it make sense for the other components to also be defined in the YAML?

The use case I'm thinking is that you might want to rely on your yaml as "the truth of everything you passed through", so that might include some paths to various files you want to use. That way you can always go back and say "o ok, this was the inputs we passed through".

openfecli/parameters/plan_network_options.py
```diff
@@ -0,0 +1,179 @@
...
+        # convert raw yaml to normalised pydantic model
+        opt = parse_yaml_planner_options(raw)
+    else:
+        opt = None
+
+    # convert normalised inputs to objects
+    if opt and opt.mapper:
+        mapper_choices = {
+            'lomap': LomapAtomMapper,
+            'lomapatommapper': LomapAtomMapper,
```
IAlibay, line 129: What is the advantage of having two choices here? Would it not be safer to just support the one thing?

openfecli/parameters/plan_network_options.py
```diff
@@ -0,0 +1,179 @@
...
+            raw = f.read()
+
+        # convert raw yaml to normalised pydantic model
+        opt = parse_yaml_planner_options(raw)
+    else:
+        opt = None
+
+    # convert normalised inputs to objects
+    if opt and opt.mapper:
+        mapper_choices = {
```
IAlibay, line 127: Not for here, but later - it'd be great if we could define this as a dictionary somewhere else, that way we can just import that (I'm thinking about expanding this to include perses, kartograf, etc..)

openfecli/parameters/plan_network_options.py
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        try:
+            cls = mapper_choices[opt.mapper.method]
+        except KeyError:
+            raise KeyError(f"Bad mapper choice: '{opt.mapper.method}'")
+        mapper_obj = cls(**opt.mapper.settings)
+    else:
+        mapper_obj = LomapAtomMapper(time=20, threed=True, element_change=False,
+                                     max3d=1)
```
IAlibay, line 139: Separate issue: @hannahbaumann did bring up the fact that `max3d=1` is not the default for LomapAtomMapper - we don't know about `element_charge`, but `max3d` really should change to match this (it's real problematic otherwise).

openfecli/parameters/plan_network_options.py
```diff
@@ -0,0 +1,179 @@
...
+                                 'ligand_network_planner', 'solvent'])
+
+
+class MapperSelection(BaseModel):
+    # model_config = ConfigDict(extra='allow', str_to_lower=True)
+    class Config:
+        extra = 'allow'
+        anystr_lower = True
+
+    method: Optional[str] = None
```
IAlibay, line 30: Rather than validating later on, could we just use a validator here for the allowed method names?

IAlibay (commented):
Couple thoughts, nothing blocking.

openfecli/commands/plan_rbfe_network.py
IAlibay: It also helps the folks that are somewhere in the "half programmatic" span of things - where they want to script something, but are only confident in writing out yaml files.

openfecli/parameters/plan_network_options.py
richardjgowers: this should ideally be up to date with our best practices, so sounds like we can patch this

openfecli/commands/plan_rbfe_network.py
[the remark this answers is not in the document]
dwhswenson: If you want to actually be explicit, please list all standard defaults for [click.Option](https://click.palletsprojects.com/en/8.1.x/api/#click.Option) and [click.Parameter](https://click.palletsprojects.com/en/8.1.x/api/#click.Parameter), except for those overridden in `YAML_OPTIONS`. Personally, I think that's way too much information, so I don't think that being explicit is useful here.

In particular, please do not explicitly list anything that is just repeated from `YAML_OPTIONS`, such as `help=YAML_OPTIONS["help"]`. This is not informative to a reader. This got into our codebase because someone copied it from [this early code](https://github.com/OpenFreeEnergy/openfe/blob/2af8dcda2921ccfa131456bb291453ca3c487dbb/openfecli/commands/atommapping.py#L20-L21), and decided they didn't need the modification to the help. But as you can see from that code, the purpose of including it is adjust the default help for the parameter.

mikemhenry (approved):
As an MVP, this is ready, I think we need to start using it raising issues as we find them

openfecli/parameters/plan_network_options.py
richardjgowers: discussed this live, it won't be possible to have a list of possible methods a priori

openfecli/parameters/plan_network_options.py
richardjgowers: see: https://github.com/OpenFreeEnergy/Lomap/pull/45

openfecli/parameters/plan_network_options.py
richardjgowers: also won't be known ahead of time, more likely this is more a EAFP situation

openfecli/parameters/plan_network_options.py
richardjgowers: just aliases to make finding options easier

PR #588, opened 2023-10-26
Add `gather --allow-partial`, improve related error message

dwhswenson:
1. Improves error messages for the case that not all edges have both legs.

2. Adds `--allow-partial` flag to allow results to be collected even if some parts of legs are missing.

The error messages will now look like this:

`This appears to be an RBFE calculation, but we're missing {'solvent'} runs for the edge with ligands {'lig1', 'lig2'}.`
`You can force partial gathering of results, without problematic edges, by using the --allow-partial flag of the gather command. Note that this may cause problems with predicting absolute free energies from the relative free energies.`

`This appears to be an RHFE calculation, but we're missing {'solvent'} runs for the edge with ligands {'lig1', 'lig2'}.`
`You can force partial gathering of results, without problematic edges, by using the --allow-partial flag of the gather command. Note that this may cause problems with predicting absolute free energies from the relative free energies.`

`Unable to determine whether this is an RBFE or an RHFE calculation. Found legs {'solvent'} for ligands {'lig2', 'lig1'}. Those ligands are missing one of: {'complex', 'vacuum'}.`
`You can force partial gathering of results, without problematic edges, by using the --allow-partial flag of the gather command. Note that this may cause problems with predicting absolute free energies from the relative free energies.`

# Developers certificate of origin

### DIFF
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -110,7 +110,7 @@ def _generate_bad_legs_error_message(set_vals, ligpair):
             f"for ligands {ligpair}. Those ligands are missing one "
             f"of: {(expected_rhfe | expected_rbfe) - set_vals}."
         )
-    else:
+    else:  # -no-cov-
         # this should never happen
         msg = (
             "Something went very wrong while determining the type "
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -164,7 +164,7 @@ def _get_ddgs(legs, error_on_missing=True):
 
 
 def _write_ddg(legs, writer, allow_partial):
-    DDGs = _get_ddgs(legs)
+    DDGs = _get_ddgs(legs, error_on_missing=not allow_partial)
     writer.writerow(["ligand_i", "ligand_j", "DDG(i->j) (kcal/mol)",
                       "uncertainty (kcal/mol)"])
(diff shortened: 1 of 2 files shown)


dwhswenson:
@mikemhenry : Could you give this branch a try on your result set?

openfecli/commands/gather.py
```diff
@@ -85,40 +86,85 @@ def legacy_get_type(res_fn):
...
     DDGs = []
     for ligpair, vals in sorted(legs.items()):
+        set_vals = set(vals)
         DDGbind = None
         DDGhyd = None
         bind_unc = None
         hyd_unc = None
 
-        if 'complex' in vals and 'solvent' in vals:
+        if set_vals == {'complex', 'solvent'}:
```
richardjgowers, line 141: this is working a little differently to before.  If I ran all three of vacuum, solvent and complex then this wouldn't now trigger either condition

looking at the original code, because it's an `elif` it also won't trigger *both* branches either

can we change this to intersections with the above cases rather than equality?

dwhswenson: Take a look at [8c643fe](https://github.com/OpenFreeEnergy/openfe/pull/588/commits/8c643fe6a52b2d636446bf136a2a79a10521e3b6). However, the output table will be nonsense if you do that.

richardjgowers (approved):

PR #594, opened 2023-10-31
Fix checkpoint name

IAlibay:


richardjgowers (approved):
I raised an issue about the blind spot in test coverage, but this probably fixes the issue

IAlibay:
I need to check but we could probably add at least a test for the nc & chk files to the dry tests if the files get generated early.

PR #598, opened 2023-10-31
Add charge backends

IAlibay:
Adds various OpenFF-powered options for assigning partial charges.

TODO:
- [x] Add more tests for the various optional backends
- [x] Allow non-RDKit toolkitregistry entries to work over the Protocols
- [x] Add protocol level tests with various charge backends


IAlibay:
Looks like the NAGL stuff requires OpenFF 0.14.4 and above, which we don't pick up yet.

For taking this forward I'd probably just not include the NAGL stuff right now anyways, it's currently there for illustrative purposes / record keeping.

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -235,3 +245,217 @@ def _add_small_mol(comp,
...
+
+    elif method.lower() == 'am1bccelf10':
+        assign_am1bccelf10_charges(
+            offmol, charge_backend, ambertools_elf_generate_n_conformers
+        )
+    elif method.lower() == 'nagl':
+        assign_nagl_am1bcc_charges(offmol, nagl_model)
+    else:
+        errmsg = (f"Unknown charge method: {method} requested. "
+                  "Available methods are: `am1bcc`, `amb1ccelf10`, and `nagl`")
```
hannahbaumann, line 460: typo 'am1bccelf10'

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -235,3 +245,217 @@ def _add_small_mol(comp,
...
+        for conf in offmol_copy.conformers:
+            offmol_copy.assign_partial_charges('am1bcc', use_conformers=[conf])
+            charges += offmol_copy.partial_charges
+
+        charges /= len(offmol_copy.conformers)
+
+        offmol.partial_charges = charges
+
+
+def assign_partial_charges(
```
hannahbaumann, line 404: I first was a bit confused about this function since it has the same name as the openff toolkit function, but maybe that is intentional?

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -235,3 +245,217 @@ def _add_small_mol(comp,
...
+        errmsg = (f"An unknown charge backend was requested {selection} "
+                  "available backend options are: {available_backends.keys()}")
+        raise ValueError(errmsg)
+
+    return toolkitwrapper
+
+
+def assign_am1bcc_charges(
+    offmol: OFFMol,
+    charge_backend: str = 'amber',
```
hannahbaumann, line 280: Would this have to be 'ambertools'?

openfe/protocols/openmm_utils/system_creation.py
IAlibay: This was mostly a "well I guess this method name succintly describes what it does quite well" rather than an intentional name duplication.

I don't really have any attachements tbh.

IAlibay:
DIscussions points for Thursday:

1. Currently we're still limiting things to the RDKit backend - and that's ok, the extra registry bits here I think are useful so that someone (us when we are benchmarking), could import the relevant method and generate charges as needed.
2. The user interface I'd like to see here is the option to choose a charging model in settings, we then just pass this through.
3. The "default" for now I would just keep to "same conformer am1bcc" - until we've really tested the rdkit+ambertools elf10 in production. I'm getting charging errors every so often and I've not worked out why.
4. The efl10 selection with the rdkit backend doesn't actually enforce N=10, because of the rms tolerance criteria, so in practices it's roughly N=1 to N=3. That's reasonable in terms of antechamber call time imho.
5. Benchmarking is happening, albeit very slowly because of resource issues.

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -235,3 +254,234 @@ def _add_small_mol(comp,
...
+        )
+
+        # Then we loop over the selected conformers
+        charges = np.zeros([offmol_copy.n_atoms])
+        skipped_confs = 0
+
+        for conf in offmol_copy.conformers:
+            try:
+                offmol_copy.assign_partial_charges('am1bcc', use_conformers=[conf])
+            except ValueError:
```
IAlibay, line 413: Finding this out now with hif2a - you can sometimes get antechamber failures for some conformations, these throw a ValueError.

Assuming there are N>1 conformers, my approach here would be to discard that conformer and keep going.

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -235,3 +254,234 @@ def _add_small_mol(comp,
...
+      * `oechem`: Use the OpenEye toolkit to generate charges.
+
+    ambertools_generate_n_conformers : int
+      The number of conformers to initially generate for elf10 selection.
+      Note: this is only used by the `ambertools` backend option.
+    """
+    backend = _get_toolkit_wrapper_charge_backend(charge_backend)
+
+    if charge_backend.lower() == 'oechem':
+        offmol.assign_partial_charges(
```
IAlibay, line 380: This might be an issue with the "without oechem" method we are using - more specifically this https://github.com/openforcefield/openff-toolkit/blob/dcd78c97a161b522665d7a4c74ea1eefc5ec2ceb/openff/toolkit/utils/openeye_wrapper.py#L2420 might be a problem because it might attempt to run with rdkit instead of oechem?

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -235,3 +254,234 @@ def _add_small_mol(comp,
...
+
+        charges /= (len(offmol_copy.conformers) - skipped_confs)
+
+        offmol.partial_charges = charges
+
+
+def assign_offmol_partial_charges(
+    offmol,
+    method: str,
+    charge_backend: str = "ambertools",
```
richardjgowers, line 433: Looking at this and other signatures, rather than have `foo: str` maybe we should use `Literal` to explicitly state what the different options are, so `charge_backend: Literal['ambertools'] | Literal['oechem']`.  The benefit here being you don't have to guess if we're saying `oequacpac` or `oechem`

openfe/protocols/openmm_utils/system_creation.py
richardjgowers: I imagine this will replace that hack

IAlibay: unfortunately those are two seprate things - if I remember correctly wasn't the dissallowed backends necessary because of the call to smiles serialisation when creating the db.json file that contains the ligand parameters rather than the charge calculation?

IAlibay:
@richardjgowers @hannahbaumann - this PR was more of a proof of concept than anything. I would be keen to hand it over (if we want to go in this direction).

openfe/protocols/openmm_utils/system_creation.py
ijpulidos: :100: `Literal['ambertools', 'oechem']` should also work, FYI. My 2 cents :).

richardjgowers (commented):
Looking at this again and with how complicated partial charges are, is it worth thinking that the charge backend should be an entire Settings thing unto itself, and treated as a separate step to the FF application? This might also work towards ideas of partial charge gen being a separate (eventually once per node) Unit

openfe/protocols/openmm_afe/base.py
```diff
@@ -272,21 +278,51 @@ def _get_system_generator(
...
-        system_generator = system_creation.get_system_generator(
-            forcefield_settings=settings['forcefield_settings'],
-            integrator_settings=settings['integrator_settings'],
-            thermo_settings=settings['thermo_settings'],
-            cache=ffcache,
-            has_solvent=solvent_comp is not None,
-        )
+        # Block out oechem backend to avoid any issues with
+        # smiles roundtripping between rdkit and oechem
+        with without_oechem_backend():
```
IAlibay, line 283: Here and everywhere else, to avoid completely removing access to oechem as a backend (e.g. for partial charge assignment) we instead block it out from all places where system_generator is used.

I _think_ `get_system_generator` is actually probably safe without, but it also doesn't cost us anything to be a bit extra cautious.

.github/workflows/ci.yaml
```diff
@@ -66,6 +74,16 @@ jobs:
             pydantic=${{ matrix.pydantic-version }}
           init-shell: bash
 
+      - name: "Install OpenEye"
```
IAlibay, line 77: Adding back oechem but only for the one runner (because it's very time consuming)

.gitignore
```diff
@@ -170,3 +170,6 @@ docs/ExampleNotebooks/
 
 # duecredit
 .duecredit.p
+
+# Some charge stuff
+*.model.pt
```
IAlibay, line 175: Either nagl or espaloma is creating these

environment.yml
```diff
@@ -42,4 +43,6 @@ dependencies:
   - sphinx-click
   - pip:
     - sphinx-toolbox
+    - espaloma_charge
```
IAlibay, line 46: espaloma is quite a heavy.. even if it's for testing, I could live with either not testing it or removing the code path completely if we're not happy with it (although it does look promising as a method)

IAlibay: Note: I am not advocating for either nagl or espaloma to be ACTUAL depdencies of OpenFE. I.e. on conda-forge these should not be there! This should be purely optional code that folks can use if they install the relevant things.

openfe/protocols/openmm_md/plain_md_methods.py
```diff
@@ -490,58 +523,59 @@ def run(self, *, dry=False, verbose=True,
...
-            thermo_settings=thermo_settings,
-            cache=ffcache,
-            has_solvent=solvent_comp is not None,
-        )
-
-        smc_components: dict[SmallMoleculeComponent, OFFMolecule]
+        # Note: we block out the oechem backend for all systemgenerator
+        # linked operations to avoid any smiles operations that can
+        # go wrong when doing rdkit->OEchem roundtripping
+        with without_oechem_backend():
```
IAlibay, line 543: I'm trying to be a bit conservative here by moving things around and avoiding using the context manager multiple times in repeat.

Long term this all needs a tidy up into a set of class methods that get called in a row (a bit like afe's base.py), so I'm not too fussed about pretty-ness. If someone cares, please do hack away at this!

openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -0,0 +1,415 @@
...
+    #
+    # Must include:
+    # 1. `confgen_func`: the conformer generation method
+    # 2. `charge_func`: the partial charge assignment method
+    # 2. `backends`: the allowed backends for the method
+    # 3. `max_conf`: maximum number of allowed conformations for the method
+    # 4. `charge_extra_kwargs`: any additional kwargs to be passed to the
+    #    partial charge assignment method beyond the input offmol and
+    #    the toolkitregistry
+    CHARGE_METHODS = {
```
IAlibay, line 341: The use case I'm imagining here is that someone can re-use most of the tooling here and just add more charge methods as they go along.

I.e. you can just redefine this method and add a "random conformer am1bcc" that's rdkit backend only and then the main thing is that it uses a different confgen_func.

Not sure if that's overengineering though.

openfe/tests/protocols/test_openmmutils.py
```diff
@@ -490,3 +497,272 @@ def test_convert_target_error_from_kcal_per_mole_to_kT_zero():
...
+            'nagl', 'rdkit', 'nagl', None,
+            marks=pytest.mark.skipif(
+                not HAS_NAGL or sys.platform.startswith('darwin'),
+                reason='needs NAGL and/or on macos',
+            ),
+        ),
+        pytest.param(
+            'nagl', 'ambertools', 'nagl', None,
+            marks=pytest.mark.skipif(
+                not HAS_NAGL or sys.platform.startswith('darwin')
```
IAlibay, line 672: NAGL has issues with DGL on darwin machines, see: https://github.com/openforcefield/openff-nagl/issues/78

.gitignore
IAlibay: Looks like it's espaloma: https://github.com/choderalab/espaloma_charge/issues/30

openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -0,0 +1,415 @@
...
+# up with a case where the tool is not available, e.g. if OpenEye tk
+# is not installed.
+BACKEND_OPTIONS: dict[str, list[ToolkitWrapper]] = {
+    "ambertools": [RDKitToolkitWrapper, AmberToolsToolkitWrapper],
+    "openeye": [OpenEyeToolkitWrapper],
+    "rdkit": [RDKitToolkitWrapper],
+}
+
+
+def assign_offmol_espaloma_charges(
```
IAlibay, line 68: Espaloma charge hasn't been updated in a while, @mikemhenry do you have any insights on the status? I was under the understanding that a new model was coming out soon?

IAlibay: If the status is ambiguous then we should just remove this for now.

mikemhenry: 1) I will make sure espaloma_charge is updated to use newer model and better fetching
2) I will update espaloma_charge on conda-forge

Maybe you could add a warning that using espaloma isn't well tested until we spruce things up. I think it is better to keep it and make it better than remove it and add it back in when it is working better

IAlibay:
I've handed this over to @hannahbaumann and @mikemhenry (or anyone else that feels like it) to drive code changes necessary.

mikemhenry:
These are the current issues in the PR:
```
 ============================= slowest 10 durations =============================
1237.37s call     openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex_alchemwater_totcharge[benzene_to_aniline_mapping-0-1-False-11-4-1]
1183.54s call     openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex_alchemwater_totcharge[aniline_to_benzene_mapping-0-0-True-14-1-4]
1163.22s call     openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex_alchemwater_totcharge[benzene_to_benzoic_mapping-0-0-True-14-3-1]
1151.49s call     openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex_alchemwater_totcharge[benzene_to_benzoic_mapping-0--1-False-11-3-1]
1145.19s call     openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex_alchemwater_totcharge[aniline_to_benzene_mapping-0--1-False-11-1-4]
1112.68s call     openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex_alchemwater_totcharge[benzoic_to_benzene_mapping-0-1-False-11-1-3]
1100.43s call     openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex_alchemwater_totcharge[benzoic_to_benzene_mapping-0-0-True-14-1-3]
379.40s call     openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_many_molecules_solvent
91.96s call     openfe/tests/protocols/test_openmmutils.py::TestOFFPartialCharge::test_am1bcc_conformer_nochange
44.90s call     openfecli/tests/test_rbfe_tutorial.py::test_run_tyk2
=========================== short test summary info ============================
FAILED openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_ligand_system_cutoff[cutoff0] - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'system_settings'
FAILED openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_ligand_system_cutoff[cutoff1] - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'system_settings'
FAILED openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_ligand_system_cutoff[cutoff2] - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'system_settings'
FAILED openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex[repex] - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'alchemical_sampler_settings'
FAILED openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex[sams] - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'alchemical_sampler_settings'
FAILED openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_dry_run_complex[independent] - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'alchemical_sampler_settings'
FAILED openfe/tests/protocols/test_openmm_rfe_slow.py::test_openmm_run_engine[CPU] - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'alchemical_sampler_settings'
ERROR openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::TestTyk2XmlRegression::test_particles - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'system_settings'
ERROR openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::TestTyk2XmlRegression::test_constraints - AttributeError: 'RelativeHybridTopologyProtocolSettings' object has no attribute 'system_settings'
= 7 failed, 812 passed, 28 skipped, 9 xfailed, 3 xpassed, 1778 warnings, 2 errors, 3 rerun in 8379.61s (2:19:39) =
```

richardjgowers (approved):

PR #641, opened 2023-11-16
Fix docker image building + CI improvements

### LINKED ISSUE #610: docker tests are failing
`test_log_system_probe_unconfigured` seems to be failing.

pinging @mikemhenry

mikemhenry:

Fixes #610

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -83,7 +83,7 @@ jobs:
         run: |
           pytest -n auto -v --cov=openfe --cov=openfecli --cov-report=xml --durations=10
 
-      - name: codecov
+      - name: codecov-pr
         if: ${{ github.repository == 'OpenFreeEnergy/openfe'
                 && github.event_name == 'pull_request' }}
         uses: codecov/codecov-action@v3
@@ -92,3 +92,15 @@ jobs:
           file: coverage.xml
           fail_ci_if_error: False
           verbose: True
+          flags: fast-tests
+
+      - name: codecov-schedule
+        if: ${{ github.repository == 'OpenFreeEnergy/openfe'
+                && github.event_name == 'schedule' }}
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          file: coverage.xml
+          fail_ci_if_error: False
+          verbose: True
+          flags: slow-tests
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -95,9 +95,12 @@ jobs:
           flags: fast-tests
 
       - name: codecov-merge
-        if: ${{ github.repository == 'OpenFreeEnergy/openfe'  # we only want to upload a slow report if
-                && github.event_name != 'schedule'            # 1) it isn't a schedule run
-                && github.event_name != 'pull_request' }}     # 2) it wasn't from a PR (we don't run slow tests on PRs)
+        # we only want to upload a slow report if
+        # 1) it isn't a schedule run
+        # 2) it wasn't from a PR (we don't run slow tests on PRs)
+        if: ${{ github.repository == 'OpenFreeEnergy/openfe'  
+                && github.event_name != 'schedule'            
+                && github.event_name != 'pull_request' }}     
         uses: codecov/codecov-action@v3
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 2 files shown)


mikemhenry:
If I understand https://docs.codecov.com/docs/flags correctly, using flags should help with codcov comparing coverage from when we run slow tests (in the past) to PRs

mikemhenry:
This run tests the docker pipeline: https://github.com/OpenFreeEnergy/openfe/actions/runs/6897862437

.github/workflows/ci.yaml
```diff
@@ -92,3 +92,15 @@ jobs:
           file: coverage.xml
           fail_ci_if_error: False
           verbose: True
+          flags: fast-tests
+
+      - name: codecov-schedule
+        if: ${{ github.repository == 'OpenFreeEnergy/openfe'
+                && github.event_name == 'schedule' }}
```
IAlibay, line 99: schedule is always a bit of a pain because codecov has that annoying thing where it fails if it gets the same commit too many times, would it make sense to limit this to merge instead?

.github/workflows/docker-cron.yaml
```diff
@@ -90,4 +90,4 @@ jobs:
           docker run --rm ${{ steps.fqirp.outputs.FQIRP }} openfe --version
           docker run --rm ${{ steps.fqirp.outputs.FQIRP }} python -c "import gufe; print(gufe.__version__)"
           docker run --rm ${{ steps.fqirp.outputs.FQIRP }} pytest --pyargs gufe -v
-          docker run --rm ${{ steps.fqirp.outputs.FQIRP }} pytest --pyargs openfecli --pyargs openfe -v
+          docker run --rm ${{ steps.fqirp.outputs.FQIRP }} pytest --pyargs openfe --pyargs openfecli -v
```
IAlibay, line 93: So there's some kind of fixture contention right? Do we want to raise a separate issue for this?

IAlibay (approved):
overall lgtm, couple of comments

.github/workflows/docker-cron.yaml
mikemhenry: Yes I need to, I am sure we would get the same error if we ran our test like this on CI (using pyargs and calling`openfe` and `openfecli`)

.github/workflows/ci.yaml
mikemhenry: Ah, so its the merges! I couldn't figure out how (as the yaml existed) codcov got a bunch of the coverage for the slow tests since I noticed we didn't run the slow ones on a schedule, but we run the slow tests whenever it isn't a pull_request. Should I do `not pr & not schedule` to match the logic of when we run slow tests? Like if someone make a commit to main directly, I don't think that would be a merge event

.github/workflows/docker-cron.yaml
mikemhenry: Issue raised here: https://github.com/OpenFreeEnergy/openfe/issues/643
I was able to narrow it down to `test_main_log` test for `openfecli`

.github/workflows/ci.yaml
mikemhenry: @IAlibay Let me know if the updated logic is better now, I went with "not schedule && not pull request" so we can cover any case where a slow test is ran (not pull) but only upload if it isn't a schedule run.

.github/workflows/ci.yaml
```diff
@@ -92,3 +92,16 @@ jobs:
           file: coverage.xml
           fail_ci_if_error: False
           verbose: True
+          flags: fast-tests
+
+      - name: codecov-schedule
+        if: ${{ github.repository == 'OpenFreeEnergy/openfe'  # we only want to upload a slow report if
+                && github.event_name != 'schedule'            # 1) it isn't a schedule run
```
IAlibay, line 99: there's a syntax error on this line apparently

.github/workflows/ci.yaml
```diff
@@ -92,3 +92,16 @@ jobs:
           file: coverage.xml
           fail_ci_if_error: False
           verbose: True
+          flags: fast-tests
+
+      - name: codecov-schedule
```
IAlibay, line 97:
```suggestion
      - name: codecov-merge
```
?

.github/workflows/ci.yaml
mikemhenry: can't weve comments into the expression it seems, fixed!

Issue #260, opened 2023-02-16
Add option to dump system to xml


IAlibay:
Based on call with Joe Bluck.

richardjgowers:
probably TOML not YAML, but yes

https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell

IAlibay:
Actually I was being distracted here, I meant XML 😅

richardjgowers:
duplicate of #234

PR #666, opened 2023-12-18
consequences of gufe #260 change

richardjgowers:
all mappings to Protocol.create/Transformation.__init__ are now lists of mappings, not a dict


richardjgowers:
relies on this change: https://github.com/OpenFreeEnergy/gufe/pull/260

this PR being the downstream consequences of this API change

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -243,6 +247,8 @@ def _validate_alchemical_components(
                 logger.warning(wmsg)
                 warnings.warn(wmsg)  # TODO: remove this once logging is fixed
 
+    return mapping[0]  # type: ignore
```
IAlibay, line 250: The reason why this by default was not returning mapping[0] was so that a future protocol could re-use this method without enforcing the same behaviour. I.e. a requirement that only a single mapping is necessary is specific to this specific Protocol. If we're enforcing that specific protocols' behaviour then maybe we should just make thsi a class method?

IAlibay (commented):
Overall this looks reasonable, although I'm not convinced on the method return change, in fact I think we should go more the other way and remove the "single mapping" error raising. This is especially because I know that Ivan wanted to re-use some of this tooling and maybe it's just overly restrictive for no reason?

openfe/protocols/openmm_rfe/equil_rfe_methods.py
richardjgowers: https://github.com/OpenFreeEnergy/openfe/pull/666/files#diff-1ab9cb2879dcf6337e58cc145548646e494814de8535f1c23b56c13ee9162632R209-R213

currently this function only allows a single mapping, so the function returns the only allowable input.

I'm not sure this code is sufficiently complex/interesting that it can't just be rewritten for new protocols, it's just a bunch of checks specific to this Protocol.  Trying to make it general/useful for many different cases is probably more effort than just rewriting these checks.  The choice of function/staticmethod is not that different?

richardjgowers:
I can change the function return back, but trying to share very simple functions like this feels like [false-sharing](https://en.wikipedia.org/wiki/False_sharing) where the cost of sharing is actually worse than just rewriting some simple code that is specific for each case.

dwhswenson (changes requested):
My main thing is to turn the validate method back into something that only does validation. If, in openfreeenergy/gufe#260, you take my recommendation to change `mapping` -> `mappings`, this PR will obviously need to update for that, too.

> trying to share very simple functions like this feels like [false-sharing](https://en.wikipedia.org/wiki/False_sharing) where the cost of sharing is actually worse than just rewriting some simple code that is specific for each case.

I'm not sure how this has any analogy with false sharing. In that case, you're trying NOT to share a resource and are surprised by the fact that you are, and that leads to performance issues. If I'm missing a connection here, please clarify.

If someone doesn't want to use our validator, they don't need to. If they know it exists and it meets their needs, they use it. If they don't know it exists, then they write their own — that seems like the worst loss of time, and I'm not sure how having something potentially reusable in our makes that worse. It isn't even public API now, so there's no real issue here.

In any case, my argument is that the validator should only validate.

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
```diff
@@ -521,7 +521,7 @@ def _create(
         self,
         stateA: ChemicalSystem,
         stateB: ChemicalSystem,
-        mapping: Optional[Dict[str, gufe.ComponentMapping]] = None,
+        mapping: list[gufe.ComponentMapping],
```
IAlibay, line 524: As discussed on the gufe PR, I would very much like this not to be how we do things going forward - it's very unintuitive from a Pythonic sense.

IAlibay (changes requested):
blocking on upstream decision

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -203,14 +203,13 @@ def _validate_alchemical_components(
     """
     # Check mapping
     # For now we only allow for a single mapping, this will likely change
-    if mapping is None or len(mapping.values()) > 1:
+    if mapping is None or len(mapping) > 1:
```
IAlibay, line 206: What happens if you do `len` on a single ComponentMapping?

IAlibay (commented):
So the current approve in gufe 260 is to do Optional[mapping, list[mapping]], which is convenient from a user stand point.

My original take is that definitely "no mapping" == None (because that's what I'd expect in Python), but I don't have any strong opinions on list[mapping] vs Union[mapping, list[mapping]]. We should however make changes here to account for the single mapping case.

IAlibay:
@dwhswenson - I believe your concerns have been addressed so I'll go ahead with a merge. If they haven't please do raise an issue and we can tackle further changes in a follow-up PR.

PR #673, opened 2023-12-29
OpenMMSolvation: Support for non-cubic boxes and defined box properties

### LINKED ISSUE #585: Solvate with non-cubic boxes
OpenMM now supports non-cubic boxes, we should see if we could pass this as a user input (through SolventComponent?).

IAlibay:
Fixes #585 

In this PR:
- Allows for defining the additional openmm keywords to addSolvent
- New OpenMM solvation settings + docstring


.github/workflows/ci.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
           init-shell: bash
 
       - name: "Install GUFE from main@HEAD"
-        run: python -m pip install --no-deps git+https://github.com/OpenFreeEnergy/gufe@main
+        run: python -m pip install --no-deps git+https://github.com/OpenFreeEnergy/gufe@v0.10.0
```
IAlibay, line 70: TODO - unpin after CI passes

environment.yml
```diff
@@ -26,6 +26,7 @@ dependencies:
   - click
   - typing-extensions
   - lomap2>=2.3.0
+  - openmm>=8.0.0
```
IAlibay, line 29: These solvation settings need OpenMM 8.0. Overall supporting multiple major versions of OpenMM is probably unwise as we'll need to test them all so I would suggest bumping up the minimum version?

Otherwise we can add a version check to the validator.

environment.yml
```diff
@@ -42,4 +43,4 @@ dependencies:
   - sphinx-click
   - pip:
     - sphinx-toolbox
-    - git+https://github.com/OpenFreeEnergy/gufe@main
+    - git+https://github.com/OpenFreeEnergy/gufe@v0.10.0
```
IAlibay, line 46: todo - unpin

openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -29,38 +30,183 @@
...
+    -----
+    * Cannot be defined alongside ``solvent_padding``,
+      ``number_of_solvent_molecules``, or ``box_size``.
+
+    See Also
+    --------
+    :mod:`openff.interchange.components.interchange`
+    :mod:`openff.interchange.components._packmol`
+    """
+    box_size: Optional[ArrayQuantity['nanometer']] = None
```
IAlibay, line 156: This inherits the name from OpenMM, it's not amazingly named.

I.e. if you know what it is in OpenMM this makes sense, but if you don't then it may not...

IAlibay:
I've handed this over to @hannahbaumann to drive forward whilst I'm OOO in case folks feel like pushing this for 1.0.

mattwthompson:
@hannahbaumann Josh Mitchell (@Yoshanuikabundi) wrote `openff.interchange.components._packmol` but I'm the point person in case you run into issues, or have any questions about why something is the way it is in OpenFF's infrastructure

openfe/tests/protocols/test_rfe_tokenization.py
```diff
@@ -39,7 +39,7 @@ def instance(self):
 
 class TestRelativeHybridTopologyProtocol(GufeTokenizableTestsMixin):
     cls = openmm_rfe.RelativeHybridTopologyProtocol
-    key = "RelativeHybridTopologyProtocol-bfb2f60bea0352a15bf6832d14d6e46e"
+    key = "RelativeHybridTopologyProtocol-4929ce992e66c16034cf9a15db34673e"
```
IAlibay, line 42:
```suggestion
    key = "RelativeHybridTopologyProtocol-fbc7c8ac0f58808ad4430a155453932f"
```

IAlibay:
This is low priority, but it would be amazing if we could get it merged sooner than later.

hannahbaumann (commented):
This looks good to me! Just out of curiosity (might be a dumb question): how would you specify the density of molecules in the box?

IAlibay:
> This looks good to me! Just out of curiosity (might be a dumb question): how would you specify the density of molecules in the box?

I don't think it's a dumb question.

Looks like.. you can't? http://docs.openmm.org/development/api-python/generated/openmm.app.modeller.Modeller.html#openmm.app.modeller.Modeller.addSolvent 🙈 

I suspect the answer for setting a given density would be to do it manually, selecting a given box size / vectors and then setting the number of waters added manually?

IAlibay:
> > This looks good to me! Just out of curiosity (might be a dumb question): how would you specify the density of molecules in the box?
> 
> I don't think it's a dumb question.
> 
> Looks like.. you can't? http://docs.openmm.org/development/api-python/generated/openmm.app.modeller.Modeller.html#openmm.app.modeller.Modeller.addSolvent 🙈
> 
> I suspect the answer for setting a given density would be to do it manually, selecting a given box size / vectors and then setting the number of waters added manually?

No... that doesn't work because it's not allowed. Good question indeed.

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -865,6 +870,44 @@ def test_virtual_sites_no_reassign(benzene_system, toluene_system,
...
 
 
+def test_dodecahdron_ligand_box(benzene_system, toluene_system,
+                                benzene_to_toluene_mapping, tmpdir):
+    """
+    Test that a hybrid system with a dodechadron is built properly.
+    """
+    settings = openmm_rfe.RelativeHybridTopologyProtocol.default_settings()
+    settings.solvation_settings.solvent_padding = 1.5 * unit.nanometer
+    settings.solvation_settings.box_shape = 'dodecahedron'
```
mikemhenry, line 880: Should this become our new recommendation? (dodecahedron boxes)

IAlibay:
@hannahbaumann if you have time could you have another look at this? I'd like to get to a point where we can merge this but I definitely want your stamp of approval before we do!

environment.yml
```diff
@@ -27,7 +27,7 @@ dependencies:
   - openfe-analysis>=0.2.0
   - click
   - typing-extensions
-  - openmm !=8.1.0
+  - openmm >=8.0.0,!=8.1.0
```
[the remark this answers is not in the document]
IAlibay: We do not, but we should!

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
IAlibay: Long term yes, but we'll need to do some performance benchmarks.

hannahbaumann (approved):
Lgtm, thanks @IAlibay !

mikemhenry (changes requested):
This could use a changelog entry, otherwise it is good to go!

IAlibay:
> This could use a changelog entry, otherwise it is good to go!

Good call, done!

Issue #678, opened 2024-01-04
Virtual site addition - HTF


IAlibay:
There might be a bug in the HTF where virtual sites are already added before they are handled.

IAlibay:
This is fine, we don't have errand particles.

PR #680, opened 2024-01-05
Expand water virtual site tests

IAlibay:
Adds more checks #678 

TODO:
 - [x] Add test to check if the vs particle parameters are as expected.

### DIFF
--- a/openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
+++ b/openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
@@ -14,7 +14,10 @@ from openff.units import unit
 from importlib import resources
 import xml.etree.ElementTree as ET
 
-from openmm import app, XmlSerializer, MonteCarloBarostat, NonbondedForce
+from openmm import (
+    app, XmlSerializer, MonteCarloBarostat,
+    NonbondedForce, CustomNonbondedForce
+)
 from openmm import unit as omm_unit
 from openmmtools.multistate.multistatesampler import MultiStateSampler
 import pathlib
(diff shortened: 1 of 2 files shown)


IAlibay:
@hannahbaumann this is another good one to review if you want to get familiar with a specific section of the HTF.

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -458,7 +461,58 @@ def test_dry_run_ligand_tip4p(benzene_system, toluene_system,
...
+        for entry in virtual_sites:
+            vs = htf.hybrid_system.getVirtualSite(entry)
+            vs_mass = htf.hybrid_system.getParticleMass(entry)
+            assert ensure_quantity(vs_mass, 'openff').m == pytest.approx(0)
+            vs_weights = [vs.getWeight(ix) for ix in range(vs.getNumParticles())]
+            np.testing.assert_allclose(
+                vs_weights, [0.786646558, 0.106676721, 0.106676721]
+            )
+            c, s, e = nonbond.getParticleParameters(entry)
+            assert ensure_quantity(c, 'openff').m == pytest.approx(-1.04844)
```
richardjgowers, line 507: should we look at contributing some sort of `assert_quantity_equal(val, expected_mag, expected_unit)` to openff units?

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -458,7 +461,58 @@ def test_dry_run_ligand_tip4p(benzene_system, toluene_system,
     with tmpdir.as_cwd():
         sampler = dag_unit.run(dry=True)['debug']['sampler']
         assert isinstance(sampler, MultiStateSampler)
-        assert sampler._factory.hybrid_system
+
+        # Get the htf and check
```
richardjgowers, line 465: is it going to be a pain to make this a fixture and make these three tests individual?

IAlibay: Would have to try, but iirc tmpdir is function scoped, so you can't use it in a module scoped fixture - at least I remember this being a thing in the past. (definitely wouldn't recommend this being function scoped, it's super slow)

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
IAlibay: If Matt wants to? We could also just make our own tooling if it's necessary.

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
richardjgowers: ah yeah, there's [tmp_path_factory](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-factory-fixture) which has longer scope

IAlibay: Done

richardjgowers (approved):

PR #681, opened 2024-01-08
Adding LambdaSettings object for AFE

hannahbaumann:


openfe/protocols/openmm_afe/base.py
```diff
@@ -406,14 +407,16 @@ def _get_lambda_schedule(
...
-        lambdas['lambda_sterics'] = np.concatenate(
-                [np.linspace(1, 1, n_elec), np.linspace(1, 0, n_vdw)[1:]]
-        )
+
+        lambda_elec = settings['lambda_settings'].lambda_elec
+        lambda_vdw = settings['lambda_settings'].lambda_vdw
+
+        # Reverse lambda schedule since in AbsoluteAlchemicalFactory 1
+        # means fully interacting, not stateB
+        lambda_elec = [1-x for x in lambda_elec]
```
IAlibay, line 416: TODO: need to check if upstream expects any iterable or a numpy array

IAlibay: Also python floats might be a bit more succeptible to float truncation, i.e.

```
[1.0,
 0.9,
 0.8,
 0.7,
 0.6,
 0.5,
 0.4,
 0.30000000000000004,
 0.19999999999999996,
 0.09999999999999998,
 0.0]
```

richardjgowers: I'm not sure numpy will have better float handling here, it's just a fundamental fact about floating points, you'd instead need something like this: https://docs.python.org/3/library/decimal.html 

If this is just getting used to scale numbers, then floats are fine, if it needs to look nice then we can use decimal

openfe/protocols/openmm_afe/equil_afe_settings.py
```diff
@@ -40,20 +40,52 @@
...
 
-    @validator('lambda_elec_windows', 'lambda_vdw_windows')
-    def must_be_positive(cls, v):
-        if v <= 0:
-            errmsg = ("Number of lambda steps must be positive ")
-            raise ValueError(errmsg)
+class LambdaSettings(SettingsBaseModel):
+    """Settings for lambda schedule
+    """
+    lambda_elec: list = [0.0, 0.25, 0.5, 0.75, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0,
```
richardjgowers, line 50: `list[float]`?

openfe/protocols/openmm_afe/equil_afe_settings.py
```diff
@@ -40,20 +40,52 @@
...
+    """
+    List of floats of lambda values for the van der Waals.
+    """
+    lambda_restraints: list = None
+    """
+    List of floats of lambda values for the restraints.
+    """
+
+    @validator('lambda_elec', 'lambda_vdw', 'lambda_restraints')
+    def must_be_between_0_and_1(cls, v):
```
richardjgowers, line 67: worth validating that there is actually a 0 and 1.0 value?

openfe/protocols/openmm_afe/equil_afe_settings.py
```diff
@@ -40,20 +40,52 @@
...
+    Zero means state A and 1 means state B.
+    """
+    lambda_vdw: list = [0.0, 0.0, 0.0, 0.0, 0.0, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5,
+                        0.6, 0.65, 0.7, 0.75, 0.8, 0.85, 0.9, 0.95, 1.0]
+    """
+    List of floats of lambda values for the van der Waals.
+    """
+    lambda_restraints: list = None
+    """
+    List of floats of lambda values for the restraints.
```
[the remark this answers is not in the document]
hannahbaumann: Changed the doc string.

openfe/protocols/openmm_afe/equil_afe_settings.py
hannahbaumann: I would say that (esp. for the restraints) it is ok if there's only a 0 value, or only 1. One may e.g. also only want to turn off electrostatic interactions, while keeping the vdw. But we should probably add more validators/warnings here, e.g. not allowing/warning about turning off vdw while electrostatics are still turned on. Thoughts @IAlibay ?

openfe/tests/protocols/test_openmm_afe_solvation_protocol.py
```diff
@@ -33,17 +33,28 @@ def test_create_default_settings():
...
+        lambda_settings.lambda_elec = val['elec']
+        lambda_settings.lambda_vdw = val['vdw']
+        lambda_settings.lambda_restraints = val['restraints']
+
+@pytest.mark.parametrize('val', [
+    {'elec': [0.0, 0.1, 1.0], 'vdw': [0.0, 1.0], 'restraints': [0.0, 1.0]},
+])
+def test_incorrect_number_windows(val, default_settings):
+    errmsg = "Components elec, vdw and restraints must have equal amount " \
+             "of lambda windows"
```
IAlibay, line 52:
```suggestion
    errmsg = ("Components elec, vdw and restraints must have equal amount "
              "of lambda windows")
```

As per our discussion at today's meeting - not a big deal just a nice PEP8 flair thing

IAlibay:
@dwhswenson we're not 100% sure how the settings docs is meant to work - do you have a better idea here? We can add the relevant autopydantic entry under docs, but it looks like `AlchemicalSettings` isn't behaving as expected even when there is such an entry.

openfe/protocols/openmm_afe/equil_afe_settings.py
hannahbaumann: After our discussion I will be adding validators for checking naked charges and monotonical schedules.

dwhswenson:
@IAlibay : could you clarify on the settings docs issue? I don't see anything wrong.

[This PR's docs build for `AlchemicalSettings`](https://openfe--681.org.readthedocs.build/en/681/reference/api/openmm_rfe.html#openfe.protocols.openmm_rfe.equil_rfe_settings.AlchemicalSettings) has no field for `lambda_functions`/`lambda_windows`, whereas they do exist in [current stable docs](https://docs.openfree.energy/en/stable/reference/api/openmm_rfe.html#openfe.protocols.openmm_rfe.equil_rfe_settings.AlchemicalSettings). Isn't that the change you would expect?

You'll see that this PR's build [doesn't have a link for `LambdaSettings`](https://openfe--681.org.readthedocs.build/en/681/reference/api/openmm_rfe.html#openfe.protocols.openmm_rfe.RelativeHybridTopologyProtocolSettings.lambda_settings), but that should just require adding the `.. autopydantic_model::` at `docs/reference/api/openmm_rfe.rst`, etc.

IAlibay:
> @IAlibay : could you clarify on the settings docs issue? I don't see anything wrong.
> 
> [This PR's docs build for `AlchemicalSettings`](https://openfe--681.org.readthedocs.build/en/681/reference/api/openmm_rfe.html#openfe.protocols.openmm_rfe.equil_rfe_settings.AlchemicalSettings) has no field for `lambda_functions`/`lambda_windows`, whereas they do exist in [current stable docs](https://docs.openfree.energy/en/stable/reference/api/openmm_rfe.html#openfe.protocols.openmm_rfe.equil_rfe_settings.AlchemicalSettings). Isn't that the change you would expect?
> 
> You'll see that this PR's build [doesn't have a link for `LambdaSettings`](https://openfe--681.org.readthedocs.build/en/681/reference/api/openmm_rfe.html#openfe.protocols.openmm_rfe.RelativeHybridTopologyProtocolSettings.lambda_settings), but that should just require adding the `.. autopydantic_model::` at `docs/reference/api/openmm_rfe.rst`, etc.

Let's add it to our topics of discussion later today - there's a couple of things I'm not 100% sure but it'll be easier to go through in person.

openfe/protocols/openmm_afe/equil_afe_settings.py
```diff
@@ -40,20 +40,52 @@
...
+                errmsg = "Lambda windows must be between 0 and 1."
+                raise ValueError(errmsg)
+        return v
+
+    @validator('lambda_elec')
+    def must_have_equal_number_elec_vdw_restraints_windows(cls, v, values):
+        if values['lambda_restraints'] is None:
+            if not len(v) == len(values['lambda_vdw']):
+                errmsg = (
+                    "Components elec and vdw must have equal amount of "
```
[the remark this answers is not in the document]
hannahbaumann: Great idea, added this.

IAlibay:
Validators discussed at today's call:
- naked charges
- monotonically increasing values

(for AFE)

openfe/protocols/openmm_afe/equil_afe_settings.py
```diff
@@ -40,20 +41,85 @@
...
+
+    @validator('lambda_elec', 'lambda_vdw', 'lambda_restraints')
+    def must_be_between_0_and_1(cls, v):
+        for window in v:
+            if not 0 <= window <= 1:
+                errmsg = "Lambda windows must be between 0 and 1."
+                raise ValueError(errmsg)
+        return v
+
+    @validator('lambda_elec')
```
hannahbaumann, line 87: There may be a better way to compare different list lengths here?

IAlibay (changes requested):
Regarding our discussion on documentation yesterday, could you add autopydantic entries for LambdaSettings here https://github.com/OpenFreeEnergy/openfe/blob/main/docs/reference/api/openmm_rfe.rst and https://github.com/OpenFreeEnergy/openfe/blob/main/docs/reference/api/openmm_solvation_afe.rst please?

openfe/protocols/openmm_afe/equil_afe_settings.py
hannahbaumann: This validator currently would need to errors when changing the settings from the default protocol (since when changing `lambda_elec` the list of `lambda_vdw` has still the old length and therefore would raise an error. One could move the validator into the protocol instead, or do you have other suggestions @richardjgowers ?

richardjgowers: Yeah there might be a limit to what's possible in the `@validator` and instead you want a protocol function.  The nice thing about these (when possible) is it fails fast & early.

hannahbaumann: I moved the check to the `_get_lambda_schedule` function, or would it be better to have a separate `_validator` function there?

openfe/protocols/openmm_afe/base.py
IAlibay: As per today's call: upstream expectation is a dictionary of list

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
```diff
@@ -404,8 +404,9 @@ def _default_settings(cls):
             solvent_system_settings=SystemSettings(),
             vacuum_system_settings=SystemSettings(nonbonded_method='nocutoff'),
             alchemical_settings=AlchemicalSettings(),
+            lambda_settings=LambdaSettings(),
```
IAlibay, line 407: Would it be reasonable here to change the default to match what we usually run as a default when doing AHFEs, i.e. 14 windows?

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
```diff
@@ -535,6 +588,9 @@ def _create(
         )
         self._validate_alchemical_components(alchem_comps)
 
+        # Validate the lambda schedule
+        self._validate_lambda_schedule(self.settings)
```
IAlibay, line 592: To improve method reproducibility, I would switch this to passing `self.settings.lambda_settings` instead. That way when we make derived classes we can import the staticmethod across to different places and not face typing errors because we're using different protocol settings.

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
```diff
@@ -517,6 +518,58 @@ def _validate_alchemical_components(
...
+        Checks that the lambda schedule is set up correctly.
+
+        Parameters
+        ----------
+        settings : AbsoluteSolvationSettings
+          Settings object.
+
+        Raises
+        ------
+        ValueError
```
IAlibay, line 535: We should also check that there are any non-zero restraints and raise a logger warning if there are (i.e. this protocol doesn't apply restraints, restraints won't be applied).

openfe/protocols/openmm_afe/equil_afe_settings.py
```diff
@@ -40,20 +41,58 @@
...
-    """Number of lambda electrostatic alchemical steps, default 12"""
-    lambda_vdw_windows = 12
-    """Number of lambda vdw alchemical steps, default 12"""
 
-    @validator('lambda_elec_windows', 'lambda_vdw_windows')
-    def must_be_positive(cls, v):
-        if v <= 0:
-            errmsg = ("Number of lambda steps must be positive ")
+class LambdaSettings(SettingsBaseModel):
+    """Settings for lambda schedule
```
IAlibay, line 49: Maybe add more details about what the settings class does?

```
Defines lists of floats to control various aspects of the alchemical transformation.

Notes
--------
* In all cases a lambda value of 0 defines a fully interacting state A and a non-interacting state B, whilst a value of 1 defines a fully interacting state B and a non-interacting state A.
* ``lambda_elec``, `lambda_vdw``, and ``lambda_restraints`` must all be of the same length, defining all the windows of the transformation.
```

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -30,24 +30,33 @@
...
     class Config:
         extra = 'ignore'
         arbitrary_types_allowed = True
 
-    """Settings for the alchemical protocol
-
-    This describes the lambda schedule and the creation of the
-    hybrid system.
+    """Settings for the lambda protocol
+    This describes the lambda schedule.
```
IAlibay, line 39:
```suggestion
    Lambda schedule settings.
    
    Settings controlling the lambda schedule, these include the switching function type, and the number of windows.
```
Or something like that.

IAlibay (changes requested):
Couple of things here; some docs and a couple of validation things. Otherwise this is looking good!

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
hannahbaumann: Ah yes, had changed that in the other PR, but not here, added this now!

IAlibay (approved):
This is amazing, thanks!

PR #689, opened 2024-01-17
Renaming and moving some of the protocol settings

hannahbaumann:


openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -134,62 +134,72 @@ class Config:
...
-    is at or below this value, the simulation will be considered complete. A
-    suggested value of 0.2 * `unit.boltzmann_constant` * `unit.kelvin` has
+    
+    Example:
+    real_time_analysis_interval = 250
+    timestep = 4 fs
+    steps_per_iteration = 250 * 4 fs
+    --> The free energy would be analyzed every 250 ps (250 * 250 * 4 fs)
+    """
+    early_termination_target_error: FloatQuantity = 0.0 * unit.boltzmann_constant * unit.kelvin
```
hannahbaumann, line 167: Is this unit correct or does it need to be changed?

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -57,7 +57,7 @@ class Config:
     This describes the creation of the hybrid system.
     """
 
-    unsampled_endstates = False
+    endstate_dispersion_correction = False
     """
     Whether to have extra unsampled endstate windows for long range
```
hannahbaumann, line 62: Is this doc string clear enough?

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -112,6 +108,18 @@ class RelativeHybridTopologyProtocolSettings(Settings):
     class Config:
         arbitrary_types_allowed = True
 
+    protocol_repeats: int = 3
+    """
+    Number of protocol repeats to run.  Default 3
```
richardjgowers, line 113: since there's always confusion over replicates/repeats etc, maybe define repeats here, something like "The number of completely independent repeats of the entire sampling process.  The mean of the repeats defines the final estimate of FE difference, while the variance between repeats is used as the uncertainty."

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -112,6 +108,18 @@ class RelativeHybridTopologyProtocolSettings(Settings):
...
+    """
+
+    @validator('protocol_repeats')
+    def must_be_positive(cls, v):
+        if v <= 0:
+            errmsg = "protocol_repeats must be a positive value"
```
richardjgowers, line 119: try and always spit back the bad value, e.g "protocol_repeats must be a positive value, got '0'"

openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -134,62 +134,72 @@ class Config:
...
     Default `repex`.
     """
-    online_analysis_interval: Optional[int] = 250
+    steps_per_iteration = 250 * unit.timestep  # todo: IntQuantity
     """
-    MCMC steps (i.e. ``IntegratorSettings.n_steps``) interval at which
+    Number of integration timesteps between each time the MCMC move
+    is applied. Default 250 * unit.timestep.
+    """
+    real_time_analysis_interval: Optional[int] = 250
```
richardjgowers, line 142: This isn't `unit.timestep` as it's a multiple of steps_per_iteration right?  So the default is 250 * 250?

richardjgowers: I think this is answered in the docstring below >.<

openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -345,17 +336,51 @@ class Config:
...
+
+class SimulationSettings(SettingsBaseModel):
+    """
+    Settings for simulation control, including lengths,
+    writing to disk, etc...
+    """
+    class Config:
+        arbitrary_types_allowed = True
+
+    minimization_steps = 5000
```
richardjgowers, line 365: unit.timestep?

IAlibay: Minimization is not a timestep - it's a unitless quantity because it's a pure count of the cycles of your minimizer.

openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -134,62 +134,72 @@ class Config:
     or `independent` (independently sampled lambda windows).
     Default `repex`.
     """
-    online_analysis_interval: Optional[int] = 250
+    steps_per_iteration = 250 * unit.timestep  # todo: IntQuantity
```
hannahbaumann, line 137: For this, would it technically also be `250 * timestep * unit.timestep`?

openfe/protocols/openmm_rfe/equil_rfe_settings.py
hannahbaumann: Changed this!

openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -244,13 +255,8 @@ class Config:
...
     """Size of the simulation timestep. Default 4 * unit.femtosecond."""
-    collision_rate: FloatQuantity['1/picosecond'] = 1.0 / unit.picosecond
+    langevin_collision_rate: FloatQuantity['1/picosecond'] = 1.0 / unit.picosecond
     """Collision frequency. Default 1.0 / unit.pisecond."""
-    n_steps = 250 * unit.timestep  # todo: IntQuantity
-    """
-    Number of integration timesteps between each time the MCMC move
-    is applied. Default 250 * unit.timestep.
-    """
     reassign_velocities = False
```
hannahbaumann, line 266: Potentially move to `AlchemicalSamplerSettings` or change doc string to explain what `move` means here.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -787,8 +791,7 @@ def run(self, *, dry=False, verbose=True,
             softcore_alpha=alchem_settings.softcore_alpha,
             softcore_LJ_v2=alchem_settings.softcore_LJ_v2,
```
hannahbaumann, line 792: This needs to be changed since we decided to create a new setting `softcore_LJ: Literal['gapsys', 'beutler']` (old setting: `softcore_LJ_vs: bool`. @IAlibay : Should we just change it here and keep the `softcore_LJ_v2` in the HTF, or should I also change it there?

IAlibay: For now let's assume the signature of the HTF doesn't change.

openfe/protocols/openmm_rfe/equil_rfe_settings.py
```diff
@@ -107,11 +103,35 @@ class Config:
...
+    @validator('softcore_LJ')
+    def allowed_solvent(cls, v):
+        allowed_models = ['gapsys', 'beutler']
+        if v.lower() not in allowed_models:
+            errmsg = (
+                f"Only {allowed_models} softcore potentials are allowed, "
+                f"got {v.lower()}."
+            )
+            raise ValueError(errmsg)
+        return v
```
richardjgowers, line 116:
```
In [3]: from typing import Literal

In [4]: class Thing(BaseModel):
   ...:     a: Literal['yes', 'no']
   ...: 

In [5]: Thing(a='yes')
Out[5]: Thing(a='yes')

In [6]: Thing(a='chicken')
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[6], line 1
----> 1 Thing(a='chicken')

File ~/miniconda3/envs/openfe_dev/lib/python3.11/site-packages/pydantic/main.py:341, in pydantic.main.BaseModel.__init__()

ValidationError: 1 validation error for Thing
a
  unexpected value; permitted: 'yes', 'no' (type=value_error.const; given=chicken; permitted=('yes', 'no'))
```

it looks like pydantic is smart enough to create this validator automatically when you specify a Literal.

But, it won't (easily) allow (partially) upper case versions of the literals, so 'Gapsys' here would annoyingly fail.

Not to be fixed here, but probably a quality of life thing we can fix in a future issue.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -778,17 +783,22 @@ def run(self, *, dry=False, verbose=True,
         )
 
         # 3. Create the hybrid topology
+        # a. Get softcore potential settings
+        if alchem_settings.softcore_LJ == 'gapsys':
+            softcore_LJ_v2 = True
+        elif alchem_settings.softcore_LJ == 'beutler':
+            softcore_LJ_v2 = False
```
richardjgowers, line 790: this will probably get fixed in the settings validation, but always safe to call `.lower()` on strings when you're doing comparisons, it's free and it'll save you a headache one day

richardjgowers: e.g. `if alchem_settings.softcore_LJ.lower() == 'gapsys'`

hannahbaumann: Thanks! Changed this!

openfe/protocols/openmm_rfe/equil_rfe_settings.py
hannahbaumann: Ok, will remove this validator then!

hannahbaumann:
The last commit (`Larger settings changes, suggestions`) includes a few larger changes. 
- Merge `SystemSettings` to `ForceFieldSettings` (gufe)
- Combine `AlchemicalSamplerSettings` with `SimulationSettings`:
   - Baseclass `SimulationSettings`
   - subclass `MultistateSimulationSettings` (for RFE and HFE)
   - subclass `MDSimulationSettings` (for plain MD)
 - change units to time for better user experience (less math in the head)
    - from `steps_per_iteration: 250 * unit.timestep` to `time_per_iteration: 1 * unit.picosecond`
    - from `real_time_analysis_minimum_iterations = 500` to `real_time_analysis_minimum_time = FloatQuantity['picosecond'] = 500 * unit.picosecond`
    - from `real_time_analysis_interval: Optional[int] = 250` to `real_time_analysis_interval: Optional[FloatQuantity['picosecond']] = 250 * unit.picosecond`
    - in `OutputSettings`: from `checkpoint_interval = 250 * unit.timestep` to `checkpoint_interval = FloatQuantity['picosecond'] = 1 * unit.picosecond`
- change unit to kcal/mol for `early_termination_target_error`; make this Optional

I haven't propagated the changes to the protocols yet, if anyone has objections/better ideas here, please let me know and I can still easily change it at this point.

openfe/protocols/openmm_utils/omm_settings.py
hannahbaumann: Change this to unit kcal/mol

openfe/protocols/openmm_utils/omm_settings.py
richardjgowers: is this for `reassign_velocities`?  seems like an integrator thing

IAlibay: reassign velocities is specific to the behaviour that exists when doing multi state moves, not the timestep integration itself, that makes it a property of the sampler not the integrator

openfe/protocols/openmm_utils/settings_validation.py
```diff
@@ -71,3 +71,95 @@ def get_simsteps(sim_length: unit.Quantity,
         raise ValueError(errmsg)
 
     return sim_steps
+
+
+def convert_steps_per_iteration(
```
IAlibay, line 76: Doesn't this behaviour differ from how get_simsteps behave? I would prefer that we align our divisions/conversions. Especially worth considering the whole attosecond division stuff.

IAlibay (changes requested):
The one thing for now whilst I'm looking at this. Will carve out time to review properly Monday morning.

openfe/protocols/openmm_afe/base.py
```diff
@@ -234,7 +235,7 @@ def _handle_settings(self):
           * solvation_settings : SolvationSettings
           * alchemical_settings : AlchemicalSettings
           * lambda_settings : LambdaSettings
-          * sampler_settings : AlchemicalSamplerSettings
+          * sampler_settings : MultiStateSimulationSettings
```
hannahbaumann, line 238: Is it necessary to have `sampler_settings` here of could the `simulations_settings` just be the `MultiStateSimulationsSettings`?

openfe/protocols/openmm_utils/settings_validation.py
```diff
@@ -71,3 +71,95 @@ def get_simsteps(sim_length: unit.Quantity,
...
+    rta_its = round(int(rta_fs / tpi_fs))
+
+    # convert RTA_minimum_time to iterations
+    rta_min_fs = simulation_settings.real_time_analysis_minimum_time.to(unit.femtosecond).m
+    rta_min_its = round(int(rta_min_fs / tpi_fs))
+
+    return rta_its, rta_min_its
+
+
+def convert_target_error(
```
IAlibay, line 139: Please use a more descriptive name - I have no idea what this does at a first glance.

I would suggest switching this to to reflect the kcal/mol to kT conversion.

openfe/protocols/openmm_utils/settings_validation.py
```diff
@@ -71,3 +71,95 @@ def get_simsteps(sim_length: unit.Quantity,
...
+        thermo_settings,
```
IAlibay, line 140: The aim of openmm_utils is to be as flexible as possible, can we not just pass temperature, and the error as kcal/moil and then say it returns a kT value? 

This will be useful in many other places long term.

openfe/protocols/openmm_utils/settings_validation.py
```diff
@@ -71,3 +71,95 @@ def get_simsteps(sim_length: unit.Quantity,
...
+      suitable for input to real_time_analysis_minimum_iterations
+    """
+    # TODO: Check this is correct
+    tpi_fs = simulation_settings.time_per_iteration.to(unit.femtosecond).m
+
+    # convert real_time_analysis time to interval
+    # rta_its must be number of MCMC iterations
+    # i.e. rta_fs / tpi_fs -> number of iterations
+    rta_fs = simulation_settings.real_time_analysis_interval.to(unit.femtosecond).m
+    rta_its = round(int(rta_fs / tpi_fs))
```
IAlibay, line 130: I'm not sure I agree that we should be brute force converting things.

If the initial float values at the attosecond level don't divide through properly then it should raise a ValueError, forcing an `int` is going to make for a simulation with undefined behaviour (one that isn't even recorded in the output)

openfe/protocols/openmm_utils/system_creation.py
```diff
@@ -87,10 +84,11 @@ def get_system_generator(
         'cutoffnonperiodic': app.CutoffNonPeriodic,
         'cutoffperiodic': app.CutoffPeriodic,
         'ewald': app.Ewald
-    }[system_settings.nonbonded_method.lower()]
+    }[forcefield_settings.nonbonded_method.lower()]
 
-    nonbonded_cutoff = to_openmm(
-        system_settings.nonbonded_cutoff
+    nonbonded_cutoff = ensure_quantity(
```
IAlibay, line 89: why the switch between to ensure_quantity from to_openmm? I'm pretty sure last I looked the code for ensure_quantity just used to_openmm in the background

IAlibay (changes requested):
part review

openfe/protocols/openmm_afe/base.py
```diff
@@ -234,10 +235,11 @@ def _handle_settings(self):
...
```
IAlibay, line 238: Please remove both sampler_settings and system_settings

openfe/protocols/openmm_afe/base.py
```diff
@@ -614,38 +615,46 @@ def _get_ctx_caches(
...
         integrator_settings : IntegratorSettings
+        simulation_settings : MultiStateSimulationSettings
 
         Returns
         -------
         integrator : openmmtools.mcmc.LangevinDynamicsMove
           A configured integrator object.
         """
+        # TODO: Check this is correct
+        tpi_fs = simulation_settings.time_per_iteration.to(unit.femtosecond).m
```
IAlibay, line 637: Shouldn't this be using whatever got added to openmm_utils?

openfe/protocols/openmm_afe/equil_afe_settings.py
```diff
@@ -16,20 +16,24 @@
...
-    SimulationSettings
+    OutputSettings,
 )
 import numpy as np
+from openff.models.types import FloatQuantity
+from openff.units import unit
+from typing import (
+    Literal,
+    Optional,
+)
```
IAlibay, line 36: Not seeing either of these being used in the diff?

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
```diff
@@ -397,13 +396,15 @@ def _default_settings(cls):
           a set of default settings
         """
         return AbsoluteSolvationSettings(
-            forcefield_settings=settings.OpenMMSystemGeneratorFFSettings(),
+            protocol_repeats=3,
+            solvent_forcefield_settings=settings.OpenMMSystemGeneratorFFSettings(),
+            vacuum_forcefield_settings=settings.OpenMMSystemGeneratorFFSettings(
```
IAlibay, line 401: We'll need new validators to ensure that the FF settings haven't diverged, please raise an issue.

openfe/protocols/openmm_utils/settings_validation.py
IAlibay: Going over the MD protocol - I'm probably missing something really obvious but isn't this method exactly the same as using `get_simsteps` but with an MC value of 1 (like we do there)?

IAlibay (changes requested):
Second half of review

devtools/data/gen-serialized-results.py
```diff
@@ -71,18 +82,20 @@ def generate_md_json(smc):
 def generate_ahfe_json(smc):
     settings = AbsoluteSolvationProtocol.default_settings()
     settings.solvent_simulation_settings.equilibration_length = 10 * unit.picosecond
-    settings.solvent_simulation_settings.production_length = 500 * unit.picosecond
+    settings.solvent_simulation_settings.production_length = 10 * unit.picosecond
     settings.vacuum_simulation_settings.equilibration_length = 10 * unit.picosecond
-    settings.vacuum_simulation_settings.production_length = 1000 * unit.picosecond
+    settings.vacuum_simulation_settings.production_length = 10 * unit.picosecond
```
IAlibay, line 87: I've tried this several times below 1 ns and it fails on the regular. It ends up wasting a lot more time re-running everything than just running the full nanosecond.

IAlibay (changes requested):
Looks like the docs build is failing because of missing references to settings?

IAlibay (commented):
Couple of extra comments, none blocking.

Would be nice to think about moving non-unit conversions somewhere eventually - it's going to be necessary going forward to have a good place to do kt->kj->kcal->kt etc...

IAlibay (approved):
lgtm!

PR #691, opened 2024-01-25
Add devtools for json results generation + replace old ones

### LINKED ISSUE #685: Add devtools for creating the test JSON files
We often need to re-generate JSON results files that we use in our tests. We should add some small devtools script that can re-regenerate them so that we don't always end up doing it differently each time.

IAlibay:
Fixes #685

Adds a Python script with quickrun-like serialization behaviour so we can re-generate the results files in a consistent manner.


richardjgowers (changes requested):
the 3 `.gz` files in devtools/data are the same as the test data files right?  So can I remove the 3 devtools/data copies?

IAlibay:
> the 3 `.gz` files in devtools/data are the same as the test data files right? So can I remove the 3 devtools/data copies?

Ah my bad, forgot about those - fixed

richardjgowers (approved):

PR #693, opened 2024-01-29
[DNM] Test settings migrations

IAlibay:
@dwhswenson for today's call - I'm not sure if I'm doing the migration wrong / missing something.


dwhswenson:
@IAlibay : with no changes to this PR, openfreeenergy/gufe#279 fixes most of the issues in the AFE tests here (except a test that is explicitly based on certainly assumptions of how serialization works, which the gufe PR breaks.) I'll work on it more tomorrow. A bit of 1:1 discussion to take a look at where this stands now and how to make this work best for you would be useful, but I also know that would be after a LONG day of meetings.

PR #695, opened 2024-01-30
Solvation settings prep

IAlibay:
Half way point to #673


openfe/protocols/openmm_afe/base.py
```diff
@@ -289,7 +289,7 @@ def _get_modeller(
         solvent_component: Optional[SolventComponent],
         smc_components: dict[SmallMoleculeComponent, OFFMolecule],
         system_generator: SystemGenerator,
-        solvation_settings: SolvationSettings
+        solvation_settings: BaseSolvationSettings
```
IAlibay, line 292: Using BaseSolvation settings here because we can expect derivatives of this base class to have different solvation setttings that aren't OpenMMSolvationSettings

richardjgowers (approved):

PR #703, opened 2024-02-05
Use Interchange for some OpenMM object creation

mattwthompson:
Do not merge


openfe/protocols/openmm_afe/base.py
```diff
@@ -841,19 +902,40 @@ def run(self, dry=False, verbose=True,
...
 
-        # 3. Get system generator
-        system_generator = self._get_system_generator(settings, solv_comp)
-
-        # 4. Get modeller
-        system_modeller, comp_resids = self._get_modeller(
-            prot_comp, solv_comp, smc_comps, system_generator,
-            settings['solvation_settings']
-        )
+        if (
```
IAlibay, line 905: (just in case this moves towards merge in the next couple of weeks): it may be better if this logic went into the getter than in the `run` of base - the idea here is that `run` should be kept pretty lean so it's easy to swap out class methods.

See https://github.com/OpenFreeEnergy/openfe/blob/ahfe-packmol/openfe/protocols/openmm_afe/base.py#L433 for example.

IAlibay (commented):
I was under the impressions that this was going into https://github.com/OpenFreeEnergy/openfe_skunkworks so maybe this isn't meant to aim towards a merge.

Just leaving the one comment in case stuff progresses whilst I'm out - mostly a style thing about moving selection logic to a separate class method to keep `run` pretty much just something where you can subclass and swap out methods.

openfe/protocols/openmm_afe/base.py
mattwthompson: I see the benefit of that design; it didn't come about naturally since the Interchange code path does not always need to call `_get_system_generator`

IAlibay:
I'm closing this as stale / not planned - the likelihood is that any interchange use will be some kind of update to whatever we do with openmmforcefields or pontibus.

PR #719, opened 2024-02-11
Fix checkpointing

### LINKED ISSUE #717: `checkpoint_interval` is improperly being used across protocols
## MD protocol:

https://github.com/OpenFreeEnergy/openfe/blob/61732ae9282b464a1119e98ba4fddc92aff838af/openfe/protocols/openmm_md/plain_md_methods.py#L400C44-L403

Here `output_settings.checkpoint_interval.m`, which is in units of picosecond (see: https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/protocols/openmm_utils/omm_settings.py#L169C5-L169C24) is being used directly.

However, `reportInterval` takes inputs in units of timestep: http://docs.openmm.org/development/api-python/generated/openmm.app.checkpointreporter.CheckpointReporter.html

## RFE protocol:

https://github.com/OpenFreeEnergy/openfe/blob/1917f5ae8e14032d54848c1db5687aa288ce1618/openfe/protocols/openmm_rfe/equil_rfe_methods.py#L832-L842

The checkpoint interval is set as:

checkpoint_interval(femtosecond) / integrator_timestep (femtoseconds) -> rounded and then int

This would assume that the checkpoint interval is in intervals of timestep. However, because this is openmmtools, the checkpoint interval is in frequency of iteration, not timestep.

## AFE protocol

https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/protocols/openmm_afe/base.py#L566

Here we're just setting it to the picosecond value. This should, like the RFE protocol, be in intervals of iteration.

IAlibay:
Fixes #717 

I'm just roughly fixing the checkpointing stuff so I can push that AFE MD equillibration stuff.

Can someone take over here and do some proper tests?

Ideally we need to run a short simulation and check that the number of frames/checkpoints/log entries/etc.. matches the expected number.

- [X] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.


openfe/tests/protocols/test_openmm_afe_slow.py
```diff
@@ -45,18 +45,20 @@ def test_openmm_run_engine(platform,
 
     # Run a really short calculation to check everything is going well
     s = openmm_afe.AbsoluteSolvationProtocol.default_settings()
-    s.alchemsampler_settings.n_repeats = 1
+    s.protocol_repeats = 1
```
IAlibay, line 48: Looks like a lot of these slow tests didn't get fixed when the settings merge got made.

richardjgowers (approved):

PR #727, opened 2024-02-16
detect if API breaks on PR

mikemhenry:


mikemhenry:
See https://mkdocstrings.github.io/griffe/checking/#detected-breakages

We get errors, but that is because we are breaking things, after the 1.0 tag, we can use this check to make sure we update the change log with breakage.

IAlibay (approved):
This is really cool @mikemhenry - I personally like it.

IAlibay (approved):

.github/workflows/change.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "3.12"
+
+    - name: Check for API breaks
+      continue-on-error: true
```
IAlibay, line 25: I can't remember what side effect this has anymore - does it mean that we get a green tick even if it fails? If so, we should make it loud and not a merge requirement imho. That way we can have the red tick and know that it was API breaking?

atravitz (approved):
I like this! @mikemhenry could you just address [this comment](https://github.com/OpenFreeEnergy/openfe/pull/727/files#r1685820906) before merge?

.github/workflows/change.yaml
mikemhenry: Okay counteroffer, how about if the step fails (which indicates API breakage) a comment is added to the PR saying "This PR breaks API, be sure to add a note in the change log" I was thinking it would be annoying to have a red check if API breaks and you knew it was going to. That way the check will always be green, but it tells us in a more loud way then having to read a CI log file every time.

IAlibay: To do that you'll need to trigger a separate workflow iirc. That can be a bit messy / time consuming. Could this be a follow-up PR?

mikemhenry: I don't think it needs to trigger a separate workflow. I just added a bit to do that, it worked in https://github.com/OpenFreeEnergy/openfe/pull/727/commits/192b5e95f7440b59cf955edeff6d22fea718c5d7 now I just need to add an API break to see if it does what we expect

mikemhenry:
cool, bot posted the comment and the check stayed green, and this is what the output looks like on the action
image
We can refine the bot to update the comment/or do more, but it works!

.github/workflows/change.yaml
IAlibay: @mikemhenry is it not that anything that needs secrets (like comment posting) will fail for PRs coming from non-org members unless it's in a `pull_request_target` workflow type?

.github/workflows/change.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
IAlibay, line 15: Just as an FYI - last I tried this, it didn't work when the PR comes from a fork.

mikemhenry:
From https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

>Having said that, mixing pull_request_target with an explicit PR checkout is not always vulnerable. The workflow may, for example:
>* Reformat and commit the code
>* Checkout both base and head repositories and generate a diff
>* Run grep on the checked out source.

>Generally speaking, when the PR contents are treated as passive data, i.e. not in a position of influence over the build/testing process, it is safe. But the repository owners must be extra careful not to trigger any script that may operate on PR controlled contents like in the case of npm install.

I believe now that we don't actually _build_ the code in the PR, we should be okay.

Issue #714, opened 2024-02-10
equilibration_length_nvt doesn't work in gas phase calculations


IAlibay:
Follow up from #713 

The MDSimulationSettings's `equilibration_length_nvt` doesn't make a ton of sense when you do a gase phase calculation. Because the exact simulation isn't known until we know the chemical system, it might just make sense to make this an optional setting?

IAlibay:
Similarly `equil_NVT_structure` should be an optional.

Issue #718, opened 2024-02-11
Inconsistent use of capitalization for certain words (e.g. NVT) in settings


IAlibay:
Follow-up from #714 

There are places where we expect "nvt" and others "NVT".

This has caught me out a few times over the last few days.

Issue #712, opened 2024-02-10
Barostat frequency should be optional


IAlibay:
Given that gas phase calculations don't use barostats, then it might make sense to make the barostat frequency optional?

IAlibay:
I'm reopening this issue because the underlying problem hasn't been addressed.

Langevin integrators are going to be needed for simulation types where barostats don't make a lot of sense - e.g. implicit solvent. The current approach doesn't really make a lot of sense for users in those scenarios, at least as far as I understand things.

May I ask that the decision makers please explain how they envision things working out for non barostat requiring calculations?

richardjgowers:
@IAlibay yeah we looked at this, it's messy to make it Optional so we just ignored this for now and chose to properly document that Barostats are ignored for our vacuum legs, see: https://github.com/OpenFreeEnergy/openfe/pull/729#pullrequestreview-1884643159

If/when we look at implicit solvent we can look at this again but it didn't seem like a blocker for the release

IAlibay:
Just to say - implicit solvent is going to happen near term, as it's already compatible with the AFE protocol. I have experimental validation work that will need this pretty soon (order month).

camattelaer:
hi

I was looking into the OpenFE documentation but could not find any reference whether or not the ABFE protocol is compatible with implicit solvation. I found this issue which apparently addresses the fact that (in beginning 2024) implicit solvation would be required soon, so i just wanted to ask if the current ABFE implementation in OpenFE would be compatible with an openmm context created with implicit solvation?

Kind regards and thanks in advance

Charles-Alexandre

IAlibay:
Hi @camattelaer - at this moment in time, ABFEs do not support implicit solvation. There is a bit of work in this direction, but it's not currently on our priority list. If this is something of interest to you, could you maybe open an issue discussing your use cases?

camattelaer:
Hi @IAlibay 

Thank you for the very quick reply and clarification!

For some context: I came across the following [preprint](https://arxiv.org/abs/2509.21808), they also have a matching [github](https://github.com/LuchkoLab/Impicit-Solvent-DDM) repo. I am aware that their approach is slightly different from typical ABFE thermodynamic cycles, but i wanted to investigate if there are OpenMM compatible libraries for implicit solvent ABFE, before trying to modify this DDM approach (which is only working with AMBER) to OpenMM.

[...]

PR #729, opened 2024-02-16
Change some MD settings

hannahbaumann:

### DIFF
--- a/openfe/protocols/openmm_md/plain_md_methods.py
+++ b/openfe/protocols/openmm_md/plain_md_methods.py
@@ -326,7 +326,7 @@ class PlainMDProtocolUnit(gufe.ProtocolUnit):
         )
         # equilibrate
         # NVT equilibration
-        if equil_steps_nvt is not None or equil_steps_nvt == 0:
+        if equil_steps_nvt is not None or equil_steps_nvt != 0:
             if verbose:
                 logger.info("Running NVT equilibration")
 
--- a/openfe/protocols/openmm_md/plain_md_methods.py
+++ b/openfe/protocols/openmm_md/plain_md_methods.py
@@ -356,7 +356,7 @@ class PlainMDProtocolUnit(gufe.ProtocolUnit):
             mdtraj_top.subset(selection_indices),
         )
         traj.save_pdb(
-            shared_basepath / output_settings.equil_NVT_structure
+            shared_basepath / output_settings.equil_nvt_structure
         )
 
         # NPT equilibration
@@ -388,7 +388,7 @@ class PlainMDProtocolUnit(gufe.ProtocolUnit):
             mdtraj_top.subset(selection_indices),
         )
         traj.save_pdb(
-            shared_basepath / output_settings.equil_NPT_structure
+            shared_basepath / output_settings.equil_npt_structure
         )
 
         # production
@@ -633,8 +633,8 @@ class PlainMDProtocolUnit(gufe.ProtocolUnit):
             return {
                 'system_pdb': shared_basepath / output_settings.preminimized_structure,
                 'minimized_pdb': shared_basepath / output_settings.minimized_structure,
-                'nvt_equil_pdb': shared_basepath / output_settings.equil_NVT_structure,
-                'npt_equil_pdb': shared_basepath / output_settings.equil_NPT_structure,
+                'nvt_equil_pdb': shared_basepath / output_settings.equil_nvt_structure,
+                'npt_equil_pdb': shared_basepath / output_settings.equil_npt_structure,
                 'nc': shared_basepath / output_settings.production_trajectory_filename,
                 'last_checkpoint': shared_basepath / output_settings.checkpoint_storage_filename,
             }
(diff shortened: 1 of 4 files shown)


openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -116,7 +116,7 @@ class Config:
     """
     constraint_tolerance = 1e-06
     """Tolerance for the constraint solver. Default 1e-6."""
-    barostat_frequency = 25 * unit.timestep  # todo: IntQuantity
+    barostat_frequency: Optional[FloatQuantity['timestep']] = 25 * unit.timestep  # todo: IntQuantity
```
hannahbaumann, line 119: Do we have to add a warning that if this is None, the user shouldn't be running a solvent NPT calculations?

openfe/protocols/openmm_md/plain_md_methods.py
```diff
@@ -326,38 +326,39 @@ def _run_MD(simulation: openmm.app.Simulation,
...
-            ).getPositions()))
-
-        traj = mdtraj.Trajectory(
-            positions[selection_indices, :],
-            mdtraj_top.subset(selection_indices),
-        )
-        traj.save_pdb(
-            shared_basepath / output_settings.equil_NVT_structure
-        )
+        if equil_steps_nvt is not None or equil_steps_nvt != 0:
```
richardjgowers, line 329: I think this can just be `if equil_steps_nvt:`.  You've got three possible cases, None, 0 and >0, and that if statement will correctly work on those

openfe/protocols/openmm_md/plain_md_methods.py
```diff
@@ -256,7 +256,7 @@ def _run_MD(simulation: openmm.app.Simulation,
                 temperature: unit.Quantity,
                 barostat_frequency: unit.Quantity,
                 timestep: unit.Quantity,
-                equil_steps_nvt: int,
+                equil_steps_nvt: Optional[int],
```
richardjgowers, line 259: also update the docstring below to reflect this change

openfe/protocols/openmm_md/plain_md_methods.py
```diff
@@ -633,8 +639,8 @@ def run(self, *, dry=False, verbose=True,
             return {
                 'system_pdb': shared_basepath / output_settings.preminimized_structure,
                 'minimized_pdb': shared_basepath / output_settings.minimized_structure,
-                'nvt_equil_pdb': shared_basepath / output_settings.equil_NVT_structure,
-                'npt_equil_pdb': shared_basepath / output_settings.equil_NPT_structure,
+                'nvt_equil_pdb': shared_basepath / output_settings.equil_nvt_structure,
```
richardjgowers, line 642: this file isn't always created any more, so probably need to only add this dict entry if the file is created

openfe/protocols/openmm_utils/omm_settings.py
richardjgowers: There's 4 options here I guess. 
- gas phase & barostat freq set -> warn that this is ignored?
- gas phase & barostat freq None -> ok
- non-gas phase & barostat freq set -> ok
- non-gas phase & barostat freq None -> error out

openfe/protocols/openmm_md/plain_md_methods.py
hannahbaumann: I also made the `equil_npt_structure` optional, so will also change it for that. Is there a way of doing that that is not a series of if statements?

openfe/protocols/openmm_utils/omm_settings.py
hannahbaumann: I guess one could make it non-Optional again and then only do option 1 (gas phase & barostat freq set -> warn that this is ignored)?

openfe/protocols/openmm_md/plain_md_methods.py
hannahbaumann: Changed this!

openfe/protocols/openmm_md/plain_md_methods.py
hannahbaumann: Updated the doc strong (and also in the openmm settings)

openfe/protocols/openmm_md/plain_md_methods.py
richardjgowers: I don't think so

openfe/protocols/openmm_utils/omm_settings.py
richardjgowers: I think that's what we currently have - It's not worth warning if there's nothing that the user can do to avoid the warning.  #712 is asking to make it optional since sometimes it's not relevant.  I guess elsewhere we also have settings that are ignored in certain conditions.

hannahbaumann: @richardjgowers : So you're suggesting leaving it non-Optional?

richardjgowers: Yeah I think it's worse if it's optional, so keep it as-is and document that this setting is ignored for gas-phase simulations

hannahbaumann:
Fixes https://github.com/OpenFreeEnergy/openfe/issues/714 and https://github.com/OpenFreeEnergy/openfe/issues/718

richardjgowers (approved):

PR #737, opened 2024-02-26
Issue 592 alt radial options

### LINKED ISSUE #592: generate_radial_network taking alternate central ligand options
`generate_radial_network` currently takes a `SmallMoleculeComponent` reference as the `central_ligand` argument.  This could be extended to allow either ints (to denote index in the ligand list) and strings (to give the name of the ligand).  This isn't that useful in the Python layer, but makes using this function via the yaml CLI a lot easier.

richardjgowers:
fixes #592

Allows int or str input to `generate_radial_network`.  This is mostly so that the cli can plan radial networks.  Currently this is not possible as the argument to `central_ligand` must be a SMC object, which is impossible to specify via the CLI.

I've checked the corner case of, what if my ligands have  integer names but I want to specify by name (e.g. my ligand is called "123").  If you put `central_ligand: '123'` then the yaml treats this as str not int, so works as intended.

It's a little ugly to have a 3 way option on what this function allows, but I think this is the best way to get the CLI doing radial networks, which is a big win.

Checklist
* [ ] Added a ``news`` entry


openfe/setup/ligand_network_planning.py
```diff
@@ -83,6 +85,21 @@ def generate_radial_network(
     mappers = [_hasten_lomap(m, ligands) if isinstance(m, LomapAtomMapper)
                else m for m in mappers]
 
+    # handle central_ligand arg possibilities
+    # after this, central_ligand is resolved to a SmallMoleculeComponent
+    if isinstance(central_ligand, int):
+        ligands = list(ligands)
+        central_ligand = ligands[central_ligand]
```
hannahbaumann, line 92: Maybe add a ValueError for if index is out of bounds?

hannahbaumann (commented):
generally looks good to me! Only thing I wouldn't know how to do (re. CLI) is how to distinguish between str and int in the settings yaml file, so e.g. if a ligand name is '1', how would I ensure that I put it as str vs. int?

richardjgowers:
@hannahbaumann I added a doc section that explains the int/str case here: https://openfe--737.org.readthedocs.build/en/737/guide/cli/cli_yaml.html#customising-the-network-planner

It's a bit annoying to unit test, but I checked locally and the int/str cases are handled properly by the yaml format.

hannahbaumann (approved):

Issue #759, opened 2024-03-06
userguide: chemical systems


richardjgowers:
need to explain concept of `ChemicalSystem` and how this applies to construction thermodynamic cycles

Issue #743, opened 2024-03-01
UserGuide should explain the alchemical cycles constructed for given protocols


IAlibay:
We have a UserGuide entry that very briefly talks about setting up RFE calculations, but there's nowhere that actually says "this is the type of calculation being done".

Ideally there should be one for:
- [ ] RFE protocol (explaining the hybrid topology approach, it's provenance, etc...)
- [ ] AFE protocol (explaining the partial annhilation strategy)

PR #760, opened 2024-03-06
Docs userguide

richardjgowers:
flattens the models/ section of the userguide into setup and execution.  Creates various stub pages that need to be populated as various issues just raised.


docs/guide/setup/defining_protocols.rst
```diff
@@ -0,0 +1,65 @@
...
+This is done using a :class:`.Protocol`,
+a computational method for estimating the free energy difference between two chemical systems.
+
+Just as there are multiple possible methods for estimating free energy differences,
+there are multiple available Protocols to choose from.
+Included in the ``openfe`` package are the
+:class:`.RelativeHybridTopologyProtocol` and
+the :class:`.AbsoluteSolvationProtocol`.
+This selection is being built upon,
+both by the openfe development team as well as with external academic groups.
```
IAlibay, line 15: Rather than having a list here that will be difficult to find & update, I would suggest instead having a `Protocols` section to the userguide, where there's a list of Protocols with links to their API docs, relevant userguide entries and cookbooks.

docs/guide/setup/defining_protocols.rst
```diff
@@ -0,0 +1,65 @@
+Defining Protocols
```
IAlibay, line 1: I'm not sure this title really matches the contents of what is being presented here.

Possibly could be called "Creating and using Protocols"?

docs/guide/setup/defining_protocols.rst
```diff
@@ -0,0 +1,65 @@
...
+   prot = openmm_rfe.RelativeHybridTopologyProtocol(settings)
+
+Using Protocols
+---------------
+
+Up until this point, the ``Protocol`` has not involved any of the specific chemistry that is of interest.
+This means that a single defined ``Protocol`` can be applied to multiple pairs of ``ChemicalSystem`` objects
+to measure each difference.
+The :class:`.Transformation` object is a handy container for connecting two ``ChemicalSystem`` objects
+and the ``Protocol`` together, and represents a single thermodynamic leg.
```
IAlibay, line 57: This isn't correct - the AFE protocols ingest two end points for a full cycle, whilst the RFE protocol currently only does one leg. I would alter the language here and just avoid talking about this directly. An option here is to describe it as a "pair of states to transform between".

docs/guide/setup/defining_protocols.rst
```diff
@@ -0,0 +1,65 @@
...
+
+   from openfe.protocols import openmm_rfe
+
+   settings = openmm_rfe.RelativeHybridTopologyProtocol.default_settings()
+   prot = openmm_rfe.RelativeHybridTopologyProtocol(settings)
+
+A key point of ``Protocol`` objects is that they cannot be modified once created,
+this is important for later tracking the provenance of data,
+therefore the ``Settings`` objects must be customised before the ``Protocol`` object is created.
+For example to customise the production run length of the RFE Protocol::
```
IAlibay, line 41: A note that we should try to agree on a language standard - either american or british english

docs/guide/setup/defining_protocols.rst
```diff
@@ -0,0 +1,65 @@
...
+
+   from openfe.protocols import openmm_rfe
+
+   settings = openmm_rfe.RelativeHybridTopologyProtocol.default_settings()
+   settings.simulation_settings = '10 ns'
```
IAlibay, line 46: Following this directly will throw you a `ValidationError`, should probably be this?
```suggestion
   settings.simulation_settings.production_length = '10 ns'
```

docs/guide/setup/defining_protocols.rst
IAlibay: Also this skips the PlainMDProtocol

docs/guide/setup/defining_protocols.rst
```diff
@@ -0,0 +1,65 @@
...
+This means that a single defined ``Protocol`` can be applied to multiple pairs of ``ChemicalSystem`` objects
+to measure each difference.
+The :class:`.Transformation` object is a handy container for connecting two ``ChemicalSystem`` objects
+and the ``Protocol`` together, and represents a single thermodynamic leg.
+Often a :class:`.LigandAtomMapping` object is also required to define the correspondence of atoms,
+for further details refer to the :ref:`userguide_mappings` section.
+
+The ``Transformation`` object is then capable of creating computational work via the :func:`.Transformation.create()` method.
+For further details on this, refer to the :ref:`userguide_execution` section.
+Finaly, a ``Protocol`` is responsible for using the data generated in this process to generate an estimate of the
```
IAlibay, line 63: I wouldn't bake things too much on estimates but rather reword this with a "such as getting estimates of the free energy difference". It's subtle but it lets things like the PlainMDProtocol exist (which don't have free energy estimates).

IAlibay (changes requested):
Couple things from a first pass - I'm putting a block on the broken Python code.

docs/guide/setup/defining_protocols.rst
richardjgowers: I've added stub pages for individual Protocols and referenced #743 as the issue that should populate them

IAlibay (approved):
lgtm!

hannahbaumann (approved):
Lgtm, we can just figure out later where to put the thermodynamic cycles.

docs/guide/setup/chemical_systems_and_thermodynamic_cycles.rst
```diff
@@ -0,0 +1,4 @@
+ChemicalSystems and thermodynamic cycles
```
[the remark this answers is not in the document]
richardjgowers: @hannahbaumann this page (#759) should be how `ChemicalSystem` can be used in general to express thermodynamic cycles.  More like a theory section of how ChemicalSystem maps to peoples understanding of that same thermodynamic cycle figure you see everywhere.

  #743 which you're working on, is on a page dedicated to a particular Protocol, and explains the details of the thermodynamic cycle that that Protocol is performing.  So building on the foundation of #759 you don't need to re-explain ChemicalSystem but instead can focus on how it's being used in this instance

PR #766, opened 2024-03-07
Fixing API settings things

### LINKED ISSUE #686: AFE protocol - API settings are not pointing to the right location
Clicking on relevant settings links here https://docs.openfree.energy/en/latest/reference/api/openmm_solvation_afe.html will take you to settings under the RFE protocol. This seems to be linked to the use of `:noindex:` in the autopydantic directives.

Note: `:noindex:` is necessary as you otherwise get an index clash from declaring the same class twice.

### LINKED ISSUE #741: AFE protocol settings not showing up?
Checking here the AFE protocol settings are not showing up: https://docs.openfree.energy/en/latest/reference/api/openmm_solvation_afe.html

### LINKED ISSUE #742: Several settings classes don't have docs entries
For now seeing OpenFFpartialChargeSettings, MultiStateOutputSettings and OpenMMSolvationSettings here: https://docs.openfree.energy/en/latest/reference/api/openmm_rfe.html

Probably a few others (MD output settings?)

### LINKED ISSUE #768: PlainMDProtocolSettings is using default BaseForceFieldSettings instead of OpenMMSystemGeneratorFFSettings
Both the RFE and AFE protocols set their `forcefield_settings` entries to specify the OpenMMSystemGeneratorFFSettings, but the PlainMDProtocolSettings doesn't.

Given that the protocol currently only specifically works with SystemGenerator - does it make sense to set that specifically?

The main thing here is that otherwise docs don't work amazingly well (as in the MD protocol points to the base FF settings which don't include much).

### LINKED ISSUE #769: PlainMDProtocol API isn't shown in docs
Needs relevant doc entry point & docstring changes

IAlibay:
Fixes #768 
Fixes #742 
Fixes #741 
Fixes #686 
Fixes #769 

Aim will be to move everything to the one settings page.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/docs/reference/api/index.rst
+++ b/docs/reference/api/index.rst
@@ -14,3 +14,4 @@ OpenFE API Reference
     openmm_rfe
     openmm_solvation_afe
     openmm_md
+    openmm_protocol_settings
--- a/docs/reference/api/openmm_md.rst
+++ b/docs/reference/api/openmm_md.rst
@@ -3,10 +3,24 @@ OpenMM Molecular Dynamics Protocol
 
 This Protocol implements a standard MD execution for a solvated protein system.
 
+
+Protocol API Specification
+--------------------------
+
+.. module:: openfe.protocols.openmm_md
+
+.. autosummary::
+   :nosignatures:
+   :toctree: generated/
+
+   PlainMDProtocol
+   PlainMDProtocolResult
+
+
 Protocol Settings
 -----------------
 
-.. module:: openfe.protocols.openmm_md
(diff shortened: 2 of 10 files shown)


IAlibay:
Ok this should be good to review.

It isn't a complete / ideal end point, but it provides a significant improvement on what we had before. Now all the Protocol APIs are exposed, the settings are aggregated when they can be, etc...

TLDR; it's still hard to traverse through the information, but at least there's nothing missing.

richardjgowers (approved):

PR #773, opened 2024-03-20
Checkpoint itnterval

### LINKED ISSUE #772: Default checkpoint interval is 250x more frequent than in v0.15
## Background

In 0.15, the value for `checkpoint_interval` had a default value of `250 * timestep`, however in the code this was actually "250 * iterations", which meant `250 * n_steps` or more specifically a default of `250 * 250 * unit.timestep`.

This can be seen here: https://github.com/OpenFreeEnergy/openfe/blob/v0.15.0/openfe/protocols/openmm_rfe/equil_rfe_methods.py#L820

Where the raw value of `checkpoint_interval` is passed to the similarly named argument of `MultiStateReporter`. As per the MultiStateReporter docstring, this is a multiple of iterations.

## Current status

In 1.0rc, the default `checkpoint_interval` is set at `1 ps` and `time_per_iteration` at `1 ps`. This means that we writing a checkpoint every single iteration, exactly 250x more than we were in 0.15.

Since checkpoint files are appended, this is going to lead to very large outputs (@mikemhenry has reported files in excess of 850GB!).

## Solution

Change the default value to `250 ps` instead.

IAlibay:
Fixes #772 

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/news/issue_772.rst
+++ b/news/issue_772.rst
@@ -1,0 +1,25 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Checkpoint interval default value has been set to 250 ps instead of 1 ps.
+  This better matches the previous default for openfe versions < 1.0rc
+  (See issue #772).
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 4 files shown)


news/issue_772.rst
```diff
@@ -0,0 +1,25 @@
...
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Checkpoint interval default value has been set to 250 ps instead of 1 ps.
```
richardjgowers, line 19: I can fix this in the release process but we don't need changelog until we've released 1.0 proper

richardjgowers (approved):

PR #787, opened 2024-04-01
Docs protocol cycles results

richardjgowers:
@hannahbaumann I've started a section on the analysis particular to this protocol. I can finish this section if/when you merge this into your doc branch


docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,34 @@ Each Protocol simulation Unit carries out the following steps:
 
 Note: three different types of multistate sampling (i.e. replica swapping between lambda states) methods can be chosen; HREX, SAMS, and independent (no lambda swaps attempted). By default the HREX approach is selected, this can be altered using ``simulation_settings.sampler_method`` (default: ``repex``).
 
+Analysis
```
hannahbaumann, line 84: Since the analysis part will be very similar for different HREX protocols (AHFE, RBFE), would it make sense to move the analysis to a subsection of 'working with results'? As a user, that's where I would be looking for these details, I think. I feel like we had discussed this somewhere else, but I don't remember your reasons for putting it here.

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,34 @@ Each Protocol simulation Unit carries out the following steps:
...
+~~~~~~~~
+
+As standard, some analysis of the each simulation repeat is performed.
+This analysis is made available through either the dictionary of results in the execution output,
+or through some ready-made plots for quick inspection.
+This analysis can be categorised as relating
+to the energetics of the different lambda states that were sampled,
+or to the analysis of the change in structural conformation over time in each state.
+
+* lambda energetic analysis
```
hannahbaumann, line 94: I think the replica exchange analysis is related to energetics, but maybe a better title here would be something like 'energetic and replica exchange analysis'? (I don't quite like this yet either, but can't come up with something better right now)

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,34 @@ Each Protocol simulation Unit carries out the following steps:
...
+
+  * free energy overlap matrix
```
hannahbaumann, line 96: I think I would say MBAR overlap matrix or potential energy overlap matrix since this is not the overlap of the free energies, but potential energies, i think.

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,34 @@ Each Protocol simulation Unit carries out the following steps:
...
+  * replica exchange overlap matrix (for repex only)
```
hannahbaumann, line 97: Maybe ``replica exchange transition matrix (for repex only)``?

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,34 @@ Each Protocol simulation Unit carries out the following steps:
...
+  * timeseries of replica states
+  * forward and reverse estimates of free energies
```
hannahbaumann, line 99: Since this is the cumulative estimates with the aim of looking at convergence, maybe ``* forward and reverse convergence of free energy differences``

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,34 @@ Each Protocol simulation Unit carries out the following steps:
...
+
+* relating to structural analysis. If a protein was present, these analyses first center and align the system so that
+  the protein is considered the frame of reference.
+
+  * ligand RMSD  This produces a plot called ``xyz.png`` and a results entry ``xyz``
```
hannahbaumann, line 104: Maybe it would be nice to add an example figure for all the plots (also the ones above) here?

hannahbaumann (changes requested):

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -91,19 +91,40 @@ This analysis can be categorised as relating
...
-  * free energy overlap matrix
-  * replica exchange overlap matrix (for repex only)
-  * timeseries of replica states
-  * forward and reverse estimates of free energies
-
-* relating to structural analysis. If a protein was present, these analyses first center and align the system so that
+.. todo: issue 792, consolidate this page into its own analysis page and link both RBFE and AFE pages to it
+
+* Energetic and replica exchange analysis.  These analyses consider the swapping and energetic overlap between the
+  different simulated states to try to establish the convergence and correctness of the estimate of free energy
```
hannahbaumann, line 97:
```suggestion
  different simulated states to help assess the convergence and correctness of the estimate of free energy
```

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -91,19 +91,40 @@ This analysis can be categorised as relating
...
+  difference produced.
+
+  * MBAR overlap matrix.  This plot is used to assess if the different lambda states simulated overlapped energetically.
+    Each matrix element represents the probability of a sample from a given row state being observable in a given column
+    state.
+    This plot should show that the diagonal of the matrix has some "width" so that the two end states are connected.
```
hannahbaumann, line 103: Maybe on could specifically say that the values above the main diagonal (is that what the off-diagonal 1 is called?) should be greater or equal 0.03 (3%) (citing alchemlyb here: "It is recommended that there should be at least 0.03 [[Klimovich2015]] overlap between neighboring states." (https://alchemlyb.readthedocs.io/en/latest/visualisation.html)

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -91,19 +91,40 @@ This analysis can be categorised as relating
...
+  * Replica exchange overlap matrix (for repex type simulations only).  Similar to the MBAR overlap matrix, this shows
+    the probability a given lambda state exchanged with another.  Again, the diagonal of this matrix should be at least
+    tridiagonal wide for the two end states to be connected.
+  * timeseries of replica states.  This plot shows the evolution of time of the different lambda states as they are
+    exchanged between different configurations.
```
hannahbaumann, line 108: I would write it the other way around: This plot shows the time evolution of how the simulation trajectories traverse through lambda space as configurations are exchanged between different lambda states. (or something like that)

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -91,19 +91,40 @@ This analysis can be categorised as relating
...
+* Energetic and replica exchange analysis.  These analyses consider the swapping and energetic overlap between the
+  different simulated states to try to establish the convergence and correctness of the estimate of free energy
+  difference produced.
+
+  * MBAR overlap matrix.  This plot is used to assess if the different lambda states simulated overlapped energetically.
+    Each matrix element represents the probability of a sample from a given row state being observable in a given column
+    state.
+    This plot should show that the diagonal of the matrix has some "width" so that the two end states are connected.
+  * Replica exchange overlap matrix (for repex type simulations only).  Similar to the MBAR overlap matrix, this shows
+    the probability a given lambda state exchanged with another.  Again, the diagonal of this matrix should be at least
```
hannahbaumann, line 105:
```suggestion
    the probability of a given lambda state being exchanged with another.  Again, the diagonal of this matrix should be at least
```

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -91,19 +91,40 @@ This analysis can be categorised as relating
...
+    tridiagonal wide for the two end states to be connected.
+  * timeseries of replica states.  This plot shows the evolution of time of the different lambda states as they are
+    exchanged between different configurations.
+    This plot should show that the states are freely mixing and that there are no cliques forming.
+  * forward and reverse estimates of free energies.  This analysis considers the free energy estimate produced using
```
hannahbaumann, line 110:
```suggestion
  * forward and reverse estimates of free energies.  This analysis calculates the cumulative free energy estimates using
```

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -91,19 +91,40 @@ This analysis can be categorised as relating
...
-  * ligand RMSD  This produces a plot called ``xyz.png`` and a results entry ``xyz``
-  * protein 2D RMSD
-  * ligand COM drift
+  * ligand RMSD.  This produces a plot called ``ligand_RMSD.png`` and a results entry ``ligand_RMSD`` which gives the
+    RMSD of the ligand molecule over time relative to the first frame, for each simulated state.  In correct simulations
+    this metric should quickly converge to a stable value of less than 5 angstrom.
+  * ligand COM drift.  For simulations with a protein present, this metric gives the total distance of the ligand COM
+    from its initial starting (docked) position.  If this metric increases over the course of the simulation it
+    indicates that the ligand drifted from the binding pocket, and the simulation is unreliable.
+    This produces a plot called ``ligand_COM_drift.png`` and a results entry ``ligand_COM_drift``.
```
hannahbaumann, line 124: Is there a rough estimate (in terms of Angstrom) that should raise concerns?

hannahbaumann (changes requested):

docs/guide/protocols/relativehybridtopology.rst
richardjgowers: this is #792 which will get done after your PR which this is adding to gets merged

docs/guide/protocols/relativehybridtopology.rst
richardjgowers: not sure "cumulative" is the best word here, I've tried to reword it

docs/guide/protocols/relativehybridtopology.rst
richardjgowers: I've put 5A?

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,56 @@ Each Protocol simulation Unit carries out the following steps:
...
+    with off-diagonal elements being at least 0.03[2]_.
+  * Replica exchange overlap matrix (for repex type simulations only).  Similar to the MBAR overlap matrix, this shows
+    the probability of a given lambda state being exchanged with another.  Again, the diagonal of this matrix should be
+    at least tridiagonal wide for the two end states to be connected.
+  * Timeseries of replica states.  This plot shows the time evolution of the different system configurations as they are
+    exchanged between different lambda states.
+    This plot should show that the states are freely mixing and that there are no cliques forming.
+  * Forward and reverse convergence of free energy estimates.  Using increasingly larger portions of the total data,
+    this analysis calculates the free energy estimate, and is repeated both forwards and backwards in time.
+    To assess convergence the forward and reverse estimates should agree within error using only a fraction of the total
```
hannahbaumann, line 113:
```suggestion
    In this analysis, forward and backward estimates that agree within error using only a fraction of the total data suggest convergence[2]_.
```

docs/guide/protocols/relativehybridtopology.rst
hannahbaumann: I haven't worked much with COM analysis so I'm not sure what a reasonable, but i agree it should definitely be below 5A.

hannahbaumann (changes requested):

docs/guide/protocols/relativehybridtopology.rst
IAlibay: This is very much an unknown - i.e. it's very binding site dependent and there are lots of different folks in literature claiming lots of different values. Same thing for the RMSD metric (indeed because we are using non-symmetry correct RMSD, a 5A RMSD could easily be much smaller).

I would make it clear in the text that these are example values and that your mileage will vary depending on the system you are studying.

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,56 @@ Each Protocol simulation Unit carries out the following steps:
...
+* Energetic and replica exchange analysis.  These analyses consider the swapping and energetic overlap between the
+  different simulated states to help assess the convergence and correctness of the estimate of free energy
+  difference produced.
+
+  * MBAR overlap matrix.  This plot is used to assess if the different lambda states simulated overlapped energetically.
+    Each matrix element represents the probability of a sample from a given row state being observable in a given column
+    state.
+    This plot should show that the diagonal of the matrix has some "width" so that the two end states are connected,
+    with off-diagonal elements being at least 0.03[2]_.
+  * Replica exchange overlap matrix (for repex type simulations only).  Similar to the MBAR overlap matrix, this shows
```
IAlibay, line 105:
```suggestion
  * Replica exchange probability matrix (for repex type simulations only).  Similar to the MBAR overlap matrix, this shows
```
The original API docs are incorrect here, it's not an overlap matrix, it's a probability of state exchange.

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,56 @@ Each Protocol simulation Unit carries out the following steps:
...
+    this analysis calculates the free energy estimate, and is repeated both forwards and backwards in time.
+    To assess convergence the forward and reverse estimates should agree within error using only a fraction of the total
+    data[2]_.
+
+* Structural analysis. If a protein was present, these analyses first center and align the system so that
+  the protein is considered the frame of reference.
+
+  * Ligand RMSD.  This produces a plot called ``ligand_RMSD.png`` and a results entry ``ligand_RMSD`` which gives the
+    RMSD of the ligand molecule over time relative to the first frame, for each simulated state.  In correct simulations
+    this metric should quickly converge to a stable value of less than 5 angstrom.
```
hannahbaumann, line 121: Maybe something like: Large RMSD values, e.g. greater than 5 angstrom (system dependent), would indicate an unstable ligand binding mode.

hannahbaumann, line 121:
```suggestion
    RMSD of the ligand molecule over time relative to the first frame, for each simulated state.  Large RMSD values, e.g. greater than 5 angstrom (system dependent), would indicate an unstable ligand binding mode.
```

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,56 @@ Each Protocol simulation Unit carries out the following steps:
...
+  * Forward and reverse convergence of free energy estimates.  Using increasingly larger portions of the total data,
+    this analysis calculates the free energy estimate, and is repeated both forwards and backwards in time.
+    To assess convergence the forward and reverse estimates should agree within error using only a fraction of the total
+    data[2]_.
+
+* Structural analysis. If a protein was present, these analyses first center and align the system so that
+  the protein is considered the frame of reference.
+
+  * Ligand RMSD.  This produces a plot called ``ligand_RMSD.png`` and a results entry ``ligand_RMSD`` which gives the
+    RMSD of the ligand molecule over time relative to the first frame, for each simulated state.  In correct simulations
```
hannahbaumann, line 120: Is this the first frame of the production run or the initial input pdb structure (as first frame)?

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -81,6 +81,56 @@ Each Protocol simulation Unit carries out the following steps:
...
+
+* Energetic and replica exchange analysis.  These analyses consider the swapping and energetic overlap between the
+  different simulated states to help assess the convergence and correctness of the estimate of free energy
+  difference produced.
+
+  * MBAR overlap matrix.  This plot is used to assess if the different lambda states simulated overlapped energetically.
+    Each matrix element represents the probability of a sample from a given row state being observable in a given column
+    state.
+    This plot should show that the diagonal of the matrix has some "width" so that the two end states are connected,
+    with off-diagonal elements being at least 0.03[2]_.
```
hannahbaumann, line 104:
```suggestion
    with elements adjacent to the diagonal being at least 0.03[2]_.
```

hannahbaumann (changes requested):

docs/guide/protocols/relativehybridtopology.rst
richardjgowers: clarified this, it's of the production run.  So if it rearranged significantly in equilibration this wouldn't cause a large RMSD value.

docs/guide/protocols/relativehybridtopology.rst
```diff
@@ -87,12 +87,62 @@ Here are some details of how the simulation is carried out which are not detaile
 Getting the free energy estimate
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The free energy differences are obtained from simulation data using the `MBAR estimator <https://www.alchemistry.org/wiki/Multistate_Bennett_Acceptance_Ratio>`_ (multistate Bennett acceptance ratio estimator) as implemented in the `PyMBAR package <https://pymbar.readthedocs.io/en/master/mbar.html>`_.
-In addition to the MBAR estimates of the two legs of the thermodynamic cycle and the overall realtive binding free energy difference,
-the protocol also returns some metrics to help assess convergence of the results.
-The forward and reverse analysis looks at the time convergence of the free energy estimates.
-The MABR overlap matrix checks how well lambda states overlap. Since the accuracy of the MBAR estimator depends on sufficient overlap between lambda states, this is a very important metric.
```
hannahbaumann, line 94: `Since the accuracy of the MBAR estimator depends on sufficient overlap between lambda states, this is a very important metric.`
I think this sentence might be worth adding to the analysis section so the user understands why this is important.

hannahbaumann (commented):
LGTM, just that one comment and potentially the example images for the plots.

hannahbaumann (approved):
I like the table with the explanation and figures, this looks great!

PR #811, opened 2024-04-05
Update plotting.py

### LINKED ISSUE #806: MBAR overlap matrix sometimes wrong
In some cases the MBAR overlap matrix is wrong (not adding up to 1 in the rows/columns, while for other cases it is correct.
mbar_overlap_matrix
mbar_overlap_matrix

IAlibay:
Partially fixes #806 

I realised this morning whilst rewiewing @hannahbaumann's issue that this was missing - I think I just failed to understand why that if branch was added to alchemical analysis originally.

This does not fix the > 1.0 values being returned from PyMBAR, however those might not be fixable either without switching over to PyMBAR 4.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/analysis/plotting.py
+++ b/openfe/analysis/plotting.py
@@ -22,6 +22,11 @@ def plot_lambda_transition_matrix(matrix: npt.NDArray) -> Axes:
     -------
     ax : matplotlib.axes.Axes
       An Axes object to plot.
+
+    Notes
+    -----
+    Borrowed from `alchemlyb <https://github.com/alchemistry/alchemlyb/blob/master/src/alchemlyb/visualisation/mbar_matrix.py>`_
+    which itself borrows from `alchemical-analysis <https://github.com/MobleyLab/alchemical-analysis>`_. 
     """
     num_states = len(matrix)
     fig, ax = plt.subplots(figsize=(num_states / 2, num_states / 2))
@@ -32,7 +37,16 @@ def plot_lambda_transition_matrix(matrix: npt.NDArray) -> Axes:
             ax.axhline(y=i, ls="-", lw=0.5, color="k", alpha=0.25)
         for j in range(num_states):
             val = matrix[i, j]
-            val_str = "{:.2f}".format(val)[1:]
+
+            # Truncate if 0.005 from either 0 or 1
+            # https://github.com/OpenFreeEnergy/openfe/issues/806
+            if matrix[j, i] < 0.005:
+                ii = ""
+            elif matrix[j, i] > 0.995:
+                ii = "1.00"
+            else:
+                ii = "{:.2f}".format(matrix[j, i])[1:]
+
             rel_prob = val / matrix.max()
 
             # shade box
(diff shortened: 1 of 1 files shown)


openfe/analysis/plotting.py
IAlibay, line 1: This needs tests / testing - but I'm not going to do so unless it's prioritised for the release (10 mins is the limit)

IAlibay:
Problematic plots:

mbar_high_float_abnorm

mbar_high_float_prec

IAlibay:
A normal overlap matrix

overlap_normal

IAlibay:
This should be good to go - @hannahbaumann could you check that the plots make sense to you? also any ideas for extra plots we should test.

openfe/analysis/plotting.py
```diff
@@ -32,7 +38,25 @@ def plot_lambda_transition_matrix(matrix: npt.NDArray) -> Axes:
...
+                # This replicates the same behaviour as alchemical-analysis & alchemlyb
+                # i.e. near-zero values will just not be annotated
+                val_str = ""
+            elif matrix[j, i] > 0.995:
+                val_str = "{:.2f}".format(matrix[j, i])[:4]
+                # Throw a warning if the value is greater than float error
+                # This means your overlap matrix is probably wrong
+                if matrix[j, i] > 1.01:
+                    wmsg = (f"Overlap probability exceeds 1.0: {matrix[j, i]} "
+                            "This is inddicates an incorrect overlap matrix "
```
hannahbaumann, line 54:
```suggestion
                            "This indicates an incorrect overlap matrix "
```

hannahbaumann: For the first bad overlap matrix (the one that goes up to 155), the first few rows only sum up to 50 something. Should we also add a warning when rows generally don't add up to ~100 (so either far below or above that) or do you think that such a case would always also have a value in the matrix that is >1.01?

hannahbaumann (changes requested):

openfe/analysis/plotting.py
IAlibay: That's a good question - it might be better to just sum up each row & column and check if any of them exceeds 1.01? That's the same as checking if any individual cell exceeds > 1.0.1

IAlibay: Updated, please let me know what you think @hannahbaumann - wasn't too sure what to call the thing since it takes both a replica exchange probability matrix and an mbar state overlap probability matrix.

openfe/analysis/plotting.py
```diff
@@ -22,8 +23,29 @@ def plot_lambda_transition_matrix(matrix: npt.NDArray) -> Axes:
...
+    which itself borrows from `alchemical-analysis <https://github.com/MobleyLab/alchemical-analysis>`_. 
     """
     num_states = len(matrix)
+
+    # Check if any row or column isn't close to 1.0
+    # Throw a warning if it's the case
+    if (not np.allclose(matrix.sum(axis=0), 1.0) or
+        not np.allclose(matrix.sum(axis=1), 1.0)):
+        wmsg = ("Overlap/probability matrix exceeds a sum of 1.0 in one or "
+                "more column or rows of the matrix. This indicates an "
```
hannahbaumann, line 45:
```suggestion
                "more columns or rows of the matrix. This indicates an "
```

hannahbaumann (commented):
Just the one suggestion but other than that lgtm

openfe/analysis/plotting.py
IAlibay: Thanks!

hannahbaumann (approved):

PR #821, opened 2024-04-10
added min hardware specs to install doc

### LINKED ISSUE #794: [docs] Minimum hardware specs
We should have a section in the docs that tells folks what kind of hardware they _should_ use for optimal performance.

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

Resolves #794


docs/installation.rst
```diff
@@ -577,3 +577,21 @@ openmm.OpenMMException: Error loading CUDA module: CUDA_ERROR_UNSUPPORTED_PTX_VE
   This error likely means that the CUDA version that ``openmm`` was built with is incompatible with the CUDA driver.
   Try re-making the environment while specifying the correct CUDA toolkit version for your hardware and driver.
   See :ref:`installation:mamba_hpc` for more details.
+
+
+Supported Hardware
+------------------
+
+We currently support the following CPU architectures:
```
mikemhenry, line 585: @richardjgowers is this true? I am not to bothered what we list under "can create a conda env" and "we support" I mostly stuck with `x86_64` since that isn't going away, but if an upstream package decided to drop `osx-arm64`, depending on the package, it would be really hard for us to fix that

docs/installation.rst
```diff
@@ -577,3 +577,21 @@ openmm.OpenMMException: Error loading CUDA module: CUDA_ERROR_UNSUPPORTED_PTX_VE
...
+* ``linux-64`` 
+* ``osx-64``
+
+The openfe package is installable on these platforms, but support is done on a best effort basis: 
+
+* ``osx-arm64``
+
+For simulation preparation, any supported platform is suitable.
+For production use, we recomend the ``linux-64`` platform with NVIDIA GPUs for optimal performance.
+When using an OpenMM based protocol on NVIDIA GPUs, we recomend driver version ``525.60.13`` or greater.
```
IAlibay, line 596: As an [enh] suggestion - here I would suggest making it clear that we test on CUDA & CPU (for vacuum transformation) platforms. Since we don't test / benchmark on OpenCL, it's definitely at your own risk.

richardjgowers (approved):

PR #837, opened 2024-04-19
[DNM] lower limit to openff version?

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,7 +18,7 @@ dependencies:
   - pyyaml
   - coverage
   - cinnabar ==0.3.0
-  - openff-toolkit ~=0.13
+  - openff-toolkit ~=0.13.0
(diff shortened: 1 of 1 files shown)


mikemhenry:
`ERROR openfe/tests/protocols/test_openmm_equil_rfe_protocols.py::test_tip4p_check_vsite_parameters - ValueError: No registered small molecule template generators could load force field 'openff-2.1.1'`

we can't use `openff-toolkit =0.13.x`

mikemhenry:
Well at least not with `openff-2.1.1` ff, so since that is the only error, it might work with an older ff version

IAlibay:
I'm really not sure how 0.13 handles things like NAGL - it might be best not to go below 0.14 if we can avoid it.

PR #839, opened 2024-04-19
Save the Rainforest

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,7 @@ on:
   pull_request:
     branches:
       - main
+        #something
   push:
     branches:
       - main
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,8 @@ on:
   pull_request:
     branches:
       - main
-        #something
+    paths-ignore:
+      - "docs/*"
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


mikemhenry:
I don't know how github reports the status of a workflow that doesn't run:

>When using the push and pull_request events, you can configure a workflow to run based on what file paths are changed. Path filters are not evaluated for pushes of tags.

So if we only make a change to docs, I am not sure if we get a green check or not on our CI. It might also just keep it yellow. This isn't a problem, but it might mean we need to override branch protections since required checks won't run on doc changes

mikemhenry:
@richardjgowers this should be good to go, we can test it post merge

.github/workflows/ci.yaml
```diff
@@ -3,6 +3,8 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
```
richardjgowers, line 6: @mikemhenry can we put a comment here about why this entry exists?  i.e. I think this one is something like "if only docs/ touched it doesn't require the unit tests part of the matrix"

IAlibay, line 6:
```suggestion
   # Skip CI if changed files only affect the following folders
   # - docs: documentation changes don't need code validation
   # See here for more details: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths
    paths-ignore:
```

IAlibay: Done in 51ea98ef63768b3d46fb627def6387e9b5c11943

IAlibay (approved):

PR #850, opened 2024-05-10
[DNM] Test python 3.12

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,6 +39,7 @@ jobs:
         python-version:
           - "3.10"
           - "3.11"
+          - "3.12"
         openeye: ["no"]
         ommff-version: ["0.13"]
         include:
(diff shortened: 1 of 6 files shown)


mikemhenry:
okay it looks like because of https://github.com/OpenFreeEnergy/openfe/commit/56bd856136945ea56dae02298642f58a39776e27 we had a numpy<1.24 pin, but that was just in our env and not the conda package, so if we don't have any issues we should have python 3.12 support

mikemhenry:
we just need https://github.com/conda-forge/citeproc-py-feedstock/pull/9 for osx-latest (or at least that is what is currently blocking that)

mikemhenry:
Okay! So I learned we can support python 3.12 and we are maybe a PR away from supporting osx-arm64 in our CI

PR #855, opened 2024-05-31
Test with main on Lomap dev branch

### LINKED ISSUE #856: Fix OpenFE tests to match main on Lomap
The openfe tests depend on some of the old LomapAtomMapper defaults (especially a large max3d and shift=True).

This is because some of the basic test files are unalligned molecules.

To ensure the tests still pass we need to pin these defaults in some of our tests.

IAlibay:
Fixes #856 

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,6 @@ channels:
 dependencies:
   - duecredit<0.10
   - kartograf>=1.0.0
-  - lomap2>=3.0.0
   - numpy
   - networkx
   - rdkit
--- a/environment.yml
+++ b/environment.yml
@@ -26,7 +26,6 @@ dependencies:
   - openfe-analysis>=0.2.0
   - click
   - typing-extensions
-  - lomap2>=2.3.0
   - openmmtools
   - openmmforcefields
   - perses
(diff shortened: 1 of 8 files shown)


environment.yml
```diff
@@ -44,4 +44,5 @@ dependencies:
   - pip:
     - sphinx-toolbox
     - openff-nagl-models>=0.1.2
+    - git+https://github.com/OpenFreeEnergy/Lomap@main
```
IAlibay, line 47: remember to  change back!

environment.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
   - numpy
   - networkx
   - rdkit
+  - packaging
```
IAlibay, line 12: I think this is the healthiest way of doing (at least MDA just puts a direct dependency on packaging)

environment.yml
IAlibay: Last batch of tests all went green so removing this now!

environment.yml
mikemhenry: Sounds good!

mikemhenry (approved):
LGTM! So we have a follow up todo, re-write these tests so we can use the current defaults of lomap (or the defaults we suggest)

PR #866, opened 2024-06-10
Added how to install from conda-lock file

mikemhenry:
Checklist
* [ ] Added a ``news`` entry


mikemhenry:
Eventually we will want to automate the creation of these lock files, but until we do that, this is how I recommend we do it.

1) Use a single file installer to create a conda env
2) Activate the env and then run `env export --from-history > environment.yaml` which will use the same pins defined in the single file constructor.
3) Then run `conda-lock -f environment.yaml-p osx-64 -p linux-64 -p osx-arm64`

That will make a cross platform `conda-lock.yml` file.

IAlibay (approved):
Thanks @mikemhenry ! conda-lock works wonderfully ;)

PR #883, opened 2024-07-04
Use MBAR error as uncertainty with a single protocol repeat in RBFE

frannerin:
When only a single repeat of the RBFE ProtocolUnit is used in a project (which is the approach in e.g., FEP+), the MBAR error estimate should be used as the uncertainty of the free energy estimate. If not, the uncertainty is 0 (applying np.std() to a list of one number returns 0).

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/protocols/openmm_rfe/equil_rfe_methods.py
+++ b/openfe/protocols/openmm_rfe/equil_rfe_methods.py
@@ -288,8 +288,9 @@ class RelativeHybridTopologyProtocolResult(gufe.ProtocolResult):
             vals = [dG.to(u).m for dG in dGs]
             unc = np.std(vals) * u
         else:
+            # use MBAR estimate error directly for a single repeat
             uncs = [pus[0].outputs['unit_estimate_error'] for pus in self.data.values()]
-            assert len(uncs) == 1
+            assert len(uncs) == 1, "Protocol unit estimates and errors number mismatch"
             unc = uncs[0]
         return unc
 
--- a/openfe/protocols/openmm_rfe/equil_rfe_methods.py
+++ b/openfe/protocols/openmm_rfe/equil_rfe_methods.py
@@ -278,15 +278,20 @@ class RelativeHybridTopologyProtocolResult(gufe.ProtocolResult):
 
     def get_uncertainty(self) -> unit.Quantity:
         """The uncertainty/error in the dG value: The std of the estimates of
-        each independent repeat
+        each independent repeat or the MBAR estimate error for a single repeat
         """
-        dGs = [pus[0].outputs['unit_estimate'] for pus in self.data.values()]
-        u = dGs[0].u
-        # convert all values to units of the first value, then take average of magnitude
-        # this would avoid a screwy case where each value was in different units
-        vals = [dG.to(u).m for dG in dGs]
-
-        return np.std(vals) * u
+        if len(self.data.values()) > 1:
+            dGs = [pus[0].outputs['unit_estimate'] for pus in self.data.values()]
+            u = dGs[0].u
+            # convert all values to units of the first value, then take average of magnitude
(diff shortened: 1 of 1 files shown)


IAlibay:
@frannerin - I think this requires more discussion. I don't believe this is the right way to do this, as you end up with a case where the return can mean two vastly different things statistically.

The programmatic approach to getting the mbar errors is via `get_individual_estimates`, although I recognise that it isn't the most intuitive thing nor does it play very well with the CLI. What is your use case?

frannerin:
I'm running the rbfe tutorial and wanted to do everything, up to using the "gather" cli command and plotting with cinnabar, but making only 1 protocol repeat and getting !=0 errors in the plots. After, I will try my own system and ligand set but that's a different beast

IAlibay:
Thanks, I think the solution here might be to fix the CLI gather command to fetch mbar errors instead if there is insufficient data to do a standard deviation (rather than changing the behaviour of the results object itself).

frannerin:
So for gather the column should still be named "uncertainty (kcal/mol)" even if MBAR error est. is used (and e.g. print a warning)? I'm thinking in case there are transformations with different number of protocol repeats that are gathered together, for example if gather is run before all calculations finished, or some other weird extreme cases (or put a check/assert so that all transformations must have the same number of protocolunits)

IAlibay:
> So for gather the column should still be named "uncertainty (kcal/mol)" even if MBAR error est. is used (and e.g. print a warning)? I'm thinking in case there are transformations with different number of protocol repeats that are gathered together, for example if gather is run before all calculations finished, or some other weird extreme cases (or put a check/assert so that all transformations must have the same number of protocolunits)

That's a good question - and where it gets really messy.

This definitely would require a longer chat (would you be interested in joining a call about it at some point?).

A take (and not guaranteed to be the right one - definitely needs more thinking through!):
1. Gather should populate a "stdev (kcal/mol)" and "mbar error (kcal/mol)" column, where the former is just the standard deviation, and the latter is the propagated MBAR error (I _think_ that would be something like `sqrt(err1^2 + err2^2.. + errN^2)/ N`)?
2. If _all_ the samples have a standard deviation, then use the standard deviation when passing things to cinnabar.
3. If not, raise a UserWarning and user the MBAR errors.

Note: we're in a 1.x release, so renaming "uncertainty" might be a tad bit problematic - that might need some consideration (nothing stops us from starting a 2.0 branch though!).

IAlibay:
@frannerin may I please suggest that we have that chat before you spend a lot of time working on this? I would hate to see you spend a lot of efforts on a solution just for us to turn around and claim we want something else.

frannerin:
Hi @IAlibay yes I'm open for a chat, but I'm no expert I have to warn you, and I don't think I could contribute much. The last pushed commits where honestly for me and forgot they would show up here hehe 🐋 What you propose in your previous comment seems reasonable to me

hannahbaumann (changes requested):
Thank you @frannerin , this looks good to me! I think only things left would be to update the tests with the new expected outcomes as well as clean up some of the code that's commented out if that is not required?

frannerin:
Hi @hannahbaumann, first of all sorry for the delay in replying.

@IAlibay raised some very reasonable concerns about this PR in https://github.com/OpenFreeEnergy/openfe/pull/883#issuecomment-2209570108. I would be happy to give a go to the implementation of whatever final decision is taken in terms of column names, error propagation, etc, but I don't think there is one yet? I have to say, I'm already using this code for my own system and it's going ok.

On the other hand, my other PR https://github.com/OpenFreeEnergy/openfe/pull/884 (imo) does solve a bug that does not have as many conceptual implications as this one, and actually the fix that's in that PR is present already in this one, could someone take a look at that one?

I would also be happy to merge the updated main into any or both of these branches to see if everything is still working ok (and cleaning comments etc)

PR #892, opened 2024-07-21
Try speeding up CI

IAlibay:
Checklist
* [ ] Added a ``news`` entry


IAlibay:
Experiment 1: enforcing cpu threads to 1 doesn't seem to have any effect, could even be detrimental.
Experiment 2: removing coverage on long tests shaves off ~ 1/3 of the time.

IAlibay:
Ok current state gets us to ~ 50% less, which at least gets rid of more immediate worries about 5h long runners. Merge will still be a problem though :/

ijpulidos:
I don't know if you are using `feflow`'s HTF, but it might be worth noting that an optimization was performed when creating the HTF object that reduces the time significantly in CI. https://github.com/OpenFreeEnergy/feflow/pull/61/commits/42f2a5901f30d0537853a0f08eef5fd0ee4e9651

IAlibay:
> I don't know if you are using `feflow`'s HTF, but it might be worth noting that an optimization was performed when creating the HTF object that reduces the time significantly in CI. https://github.com/OpenFreeEnergy/feflow/pull/61/commits/42f2a5901f30d0537853a0f08eef5fd0ee4e9651 
> 
> EDIT: I just realized this hasn't been released yet.

We are not and won't be for a while because of dependency order issues. A lot of the cost in CI here is because of coverage though, so we can squash that out for free.

PR #896, opened 2024-07-21
fix typing, fix 2D plotting issues

### LINKED ISSUE #895: `openfe.analysis.plotting.plot_2D_rmsd` does not work for fewer than 5 states
The code calls `chain.from_iterable` which assumed a subplots call with at least 2 rows. This only ever happens if there are at least 5 states.

See: https://github.com/OpenFreeEnergy/openfe/blob/ecc8a50d07a35c42cd9ccb5e6724a85345a9873a/openfe/analysis/plotting.py#L276-L286

### LINKED ISSUE #897: Mypy fails
Looks like there's been an update to matplotlib typing and mypy is now failing.

### LINKED ISSUE #901: 2D RMSD plots leave empty axes when state count is non-divisible by 4
Example:

image

IAlibay:
* Fix #897 
* Fix #895 
* Fix #901 

TODO:
  - Need to check that the switch to flatten calls do what is expected
  - Add some tests

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/analysis/plotting.py
+++ b/openfe/analysis/plotting.py
@@ -305,7 +305,7 @@ def plot_2D_rmsd(data: list[list[float]],
     # except the last one!
     overage = len(axes.flatten()) - len(twod_rmsd_arrs)  # type: ignore
     for i in range(overage, len(axes.flatten())-1):  # type: ignore
-        axes.flatten()[i].set_axis_off()
+        axes.flatten()[i].set_axis_off()  # type: ignore
 
     plt.colorbar(axes.flatten()[0].images[0],  # type: ignore
                  cax=axes.flatten()[-1],  # type: ignore
--- a/openfe/analysis/plotting.py
+++ b/openfe/analysis/plotting.py
@@ -82,24 +82,40 @@ def plot_lambda_transition_matrix(matrix: npt.NDArray) -> Axes:
             )
 
         # anotate axes
-        base_settings = {
+        base_settings: dict[str, Union[str, int]] = {
             'size': 10, 'va': 'center', 'ha': 'center', 'color': 'k',
             'family': 'sans-serif'
         }
         for i in range(num_states):
             ax.annotate(
-                i, xy=(i + 0.5, 1), xytext=(i + 0.5, num_states + 0.5),
+                text=f"{i}",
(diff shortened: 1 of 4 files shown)


openfe/analysis/plotting.py
```diff
@@ -82,24 +82,40 @@ def plot_lambda_transition_matrix(matrix: npt.NDArray) -> Axes:
...
             'size': 10, 'va': 'center', 'ha': 'center', 'color': 'k',
             'family': 'sans-serif'
         }
         for i in range(num_states):
             ax.annotate(
-                i, xy=(i + 0.5, 1), xytext=(i + 0.5, num_states + 0.5),
+                text=f"{i}",
+                xy=(i + 0.5, 1),
+                xytext=(i + 0.5, num_states + 0.5),
+                xycoords='data',
```
IAlibay, line 94: Adding the defaults is a bit much, but thought I might as well be explicit... thoughts?

openfe/analysis/plotting.py
```diff
@@ -278,15 +294,15 @@ def plot_2D_rmsd(data: list[list[float]],
     fig, axes = plt.subplots(nrows, 4)
 
     for i, (arr, ax) in enumerate(
-            zip(twod_rmsd_arrs, chain.from_iterable(axes))):
+            zip(twod_rmsd_arrs, axes.flatten())):   # type: ignore
```
IAlibay, line 297: for some reason the return type for axes can be Axes | ndarray, but in our use case we always have ndarray (because we set ncols to 4).

openfe/analysis/plotting.py
mikemhenry: I like it!

mikemhenry (approved):

IAlibay:
The plots aren't great, but at least they work now:

9 states:
image


4 states:
image

mikemhenry (changes requested):
Looks nicer! I think these changes warrant a change log entry

hannahbaumann (approved):
The plots look good to me!

IAlibay:
Thanks both!

@mikemhenry changelog entry is done.

mikemhenry (approved):

PR #903, opened 2024-07-23
Add conda canary CI job

### LINKED ISSUE #891: Add conda canary
We need to complete #320

IAlibay:
Fixes #891 

Checklist
* [ ] Added a ``news`` entry


.github/workflows/conda_cron.yaml
```diff
@@ -0,0 +1,65 @@
+name: "conda_cron"
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
```
IAlibay, line 9: remove this once CI runs

.github/workflows/conda_cron.yaml
```diff
@@ -0,0 +1,65 @@
...
+        run: |
+          micormamba info
+          micromamba list
+          pip list
+
+      - id: run-tests
+        name: "Run tests"
+        run: |
+          # note: this only runs the fast tests
+          pytest -n auto --pyargs openfe openfecli
```
IAlibay, line 65: I'm relying on a normal failure here rather than raising the issue as we do with gufe. The plan will be to hook up Swenson's cron job checker tool.

IAlibay:
@mikemhenry any ideas why the py3.9 package for macos osx-arm64 wouldn't have bundled the test files :/ seems so odd

mikemhenry:
that is super weird since we are a `no-arch` package so it should be pulling in the same conda-forge packages for all the tests....

.github/workflows/conda_cron.yaml
```diff
@@ -0,0 +1,66 @@
...
+          condarc: |
+            channels:
+              - jaimergp/label/unsupported-cudatoolkit-shim
+              - conda-forge
+          init-shell: bash
+
+      - name: "Install openfe + optional deps"
+        run: |
+          micromamba install openfe pytest -c conda-forge
+          python -m pip install pytest-xdist PyGithub
```
mikemhenry, line 53: what does `PyGithub` do?

IAlibay: o that was just me being silly wanting to do something else

IAlibay:
> that is super weird since we are a `no-arch` package so it should be pulling in the same conda-forge packages for all the tests....

Yeah.. I don't have an osx-arm64 machine to test it out either.

mikemhenry:
We can drop 3.9 from our testing matrix entirely IMHO https://scientific-python.org/specs/spec-0000/

IAlibay:
> We can drop 3.9 from our testing matrix entirely IMHO https://scientific-python.org/specs/spec-0000/

@mikemhenry I agree, but since it's a canary for "what is currently being used", I do like to be a little bit more "what we were intending the release to be applied to". If you have a strong view here I'm happy to change my mind though!

mikemhenry:
Okay I agree it will be nice to have, and if it starts failing due to things outside of our control, we will then know and can mention that it needs to be at least python 3.x, and then drop out the failing python version from the matrix. 

@IAlibay something missing from this is that we want to try and install the latest version of `openfe` and not just what conda decides to pull in since right now it could pull in `0.1` and things would pass.

Is this something we want to figure out with some scripting, or do we hard code the version and add bumping it as part of the release checklist? 

Either way, for testing this PR, I will just hard code it for now

mikemhenry:
image
because we didn't have a `-e` argument in our bash shell, we didn't catch this typo and CI gave us a :heavy_check_mark:

mikemhenry (approved):

PR #909, opened 2024-07-26
run on 4 cores and timeout after 12 hours

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/gpu-runner.yaml
+++ b/.github/workflows/gpu-runner.yaml
@@ -29,6 +29,7 @@ jobs:
         env:
           GH_PAT: ${{ secrets.GH_PAT }}
   self-hosted-test:
+    timeout-minutes: 720  # 12 hours  
     runs-on: self-hosted
     needs:
       - start-aws-runner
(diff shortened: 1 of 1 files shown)


mikemhenry:
testing here https://github.com/OpenFreeEnergy/openfe/actions/runs/10117518516

mikemhenry:
(or did I mess up and make sure the log tests run?)

IAlibay:
> This worked great! I just failed since it expected a GPU, so the nvidia-smi commands failed, `= 2 failed, 855 passed, 40 skipped, 1 xfailed, 2 xpassed, 3594 warnings, 3 rerun in 4059.65s (1:07:39) =` way faster using `-n 4` than `-n auto`

@mikemhenry the code in this PR was 4h not 1h (I think you're looking at my logs 😅).

Main difference is that I used a newer runner and removed coverage options (which we wouldn't use anyways).

mikemhenry:
That is exactly what happened, when we switch to GPUs we probably want the timeout to be 2x or 1.5x what the "normal" run length ends up being so we have a failsafe if things go loopy, but isn't too strict to allow for runtime fluctuations

PR #927, opened 2024-09-03
[WIP] - Konnektor integration

RiesBen:
This PR is integrating Konnektor to openFE which will offer in the future a default parallelization for all Network building and more different network tools.

Checklist
* [ ] replace code with Konnektor implementations
  * [ ] MST
  * [ ] Maximal Network
  * [ ] Star Map
  * [ ] Redundant MST
  
* [ ] Test we are getting the same results
  * [ ] pytest suit

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - duecredit<0.10
   - kartograf>=1.0.0
+  - konnektor>=1.0.0
   - lomap2>=3.0.0
   - numpy
   - networkx
--- a/openfe/setup/ligand_network_planning.py
+++ b/openfe/setup/ligand_network_planning.py
@@ -17,6 +17,10 @@ from openfe.setup.atom_mapping import LigandAtomMapping
 
 from lomap import generate_lomap_network, LomapAtomMapper
 from lomap.dbmol import _find_common_core
+from konnektor.network_planners import (StarNetworkGenerator,
+                                        MaximalNetworkGenerator,
+                                        RedundantMinimalSpanningTreeNetworkGenerator,
+                                        MinimalSpanningTreeNetworkGenerator)
 
 
 def _hasten_lomap(mapper, ligands):
@@ -39,9 +43,10 @@ def _hasten_lomap(mapper, ligands):
 
 def generate_radial_network(
     ligands: Iterable[SmallMoleculeComponent],
-    central_ligand: Union[SmallMoleculeComponent, str, int],
-    mappers: Union[AtomMapper, Iterable[AtomMapper]],
+    mapper: AtomMapper,
+    central_ligand: Union[SmallMoleculeComponent, str, int] = None,
     scorer: Optional[Callable[[LigandAtomMapping], float]] = None,
+    n_processes: int = 1,
 ) -> LigandNetwork:
     """
     Plan a radial network with all ligands connected to a central node.
@@ -80,20 +85,25 @@ def generate_radial_network(
       If no scorer is supplied, the first mapping provided by the iterable
       of mappers will be used.
     """
-    if isinstance(mappers, AtomMapper):
-        mappers = [mappers]
-    mappers = [_hasten_lomap(m, ligands) if isinstance(m, LomapAtomMapper)
-               else m for m in mappers]
+    if isinstance(mapper, LomapAtomMapper):
+        mapper = _hasten_lomap(mapper, ligands)
+    nodes = list(ligands)
 
-    # handle central_ligand arg possibilities
-    # after this, central_ligand is resolved to a SmallMoleculeComponent
+    # Construct network
+    network_planner = StarNetworkGenerator(mapper=mapper,
+        scorer=scorer,
+        progress = progress,
+        n_processes = n_processes,
+    )
+
     if isinstance(central_ligand, int):
         ligands = list(ligands)
         try:
             central_ligand = ligands[central_ligand]
         except IndexError:
-            raise ValueError(f"index '{central_ligand}' out of bounds, there are "
-                             f"{len(ligands)} ligands")
+            raise ValueError(
+                f"index '{central_ligand}' out of bounds, there are "
+                f"{len(ligands)} ligands")
     elif isinstance(central_ligand, str):
         ligands = list(ligands)
         possibles = [l for l in ligands if l.name == central_ligand]
@@ -104,45 +114,17 @@ def generate_radial_network(
             raise ValueError(f"Multiple ligands called '{central_ligand}'")
         central_ligand = possibles[0]
 
-    edges = []
+    network = network_planner.generate_ligand_network(nodes, central_component=central_ligand)
 
-    for ligand in ligands:
-        if ligand == central_ligand:
-            wmsg = (f"The central_ligand {ligand.name} was also found in "
-                    "the list of ligands to arrange around the "
-                    "central_ligand this will be ignored.")
-            warnings.warn(wmsg)
-            continue
-        best_score = 0.0
-        best_mapping = None
-
-        for mapping in itertools.chain.from_iterable(
-            mapper.suggest_mappings(central_ligand, ligand)
-            for mapper in mappers
-        ):
-            if not scorer:
-                best_mapping = mapping
-                break
-
-            score = scorer(mapping)
-            mapping = mapping.with_annotations({"score": score})
-
-            if score > best_score:
-                best_mapping = mapping
-                best_score = score
-
-        if best_mapping is None:
-            raise ValueError(f"No mapping found for {ligand}")
-        edges.append(best_mapping)
-
-    return LigandNetwork(edges)
+    return network
 
 
 def generate_maximal_network(
     ligands: Iterable[SmallMoleculeComponent],
-    mappers: Union[AtomMapper, Iterable[AtomMapper]],
+    mapper: AtomMapper,
     scorer: Optional[Callable[[LigandAtomMapping], float]] = None,
     progress: Union[bool, Callable[[Iterable], Iterable]] = True,
+    n_processes: int = 1,
 ) -> LigandNetwork:
     """
     Plan a network with all possible proposed mappings.
@@ -160,7 +142,7 @@ def generate_maximal_network(
     ----------
     ligands : Iterable[SmallMoleculeComponent]
       the ligands to include in the LigandNetwork
-    mappers : AtomMapper or Iterable[AtomMapper]
+    mapper : AtomMapper or Iterable[AtomMapper]
       the AtomMapper(s) to use to propose mappings.  At least 1 required,
       but many can be given.
     scorer : Scoring function
@@ -169,42 +151,31 @@ def generate_maximal_network(
       progress bar: if False, no progress bar will be shown. If True, use a
       tqdm progress bar that only appears after 1.5 seconds. You can also
       provide a custom progress bar wrapper as a callable.
+    n_processes: int
+        parallelization of network generation.
     """
-    if isinstance(mappers, AtomMapper):
-        mappers = [mappers]
-    mappers = [_hasten_lomap(m, ligands) if isinstance(m, LomapAtomMapper)
-               else m for m in mappers]
-
+    if isinstance(mapper, LomapAtomMapper):
+        mapper = _hasten_lomap(mapper, ligands)
     nodes = list(ligands)
 
-    if progress is True:
-        # default is a tqdm progress bar
-        total = len(nodes) * (len(nodes) - 1) // 2
-        progress = functools.partial(tqdm, total=total, delay=1.5)
-    elif progress is False:
-        def progress(x): return x
-    # otherwise, it should be a user-defined callable
-
-    mapping_generator = itertools.chain.from_iterable(
-        mapper.suggest_mappings(molA, molB)
-        for molA, molB in progress(itertools.combinations(nodes, 2))
-        for mapper in mappers
+    # Construct network
+    network_planner = MaximalNetworkGenerator(mapper=mapper,
+        scorer=scorer,
+        progress = progress,
+        n_processes = n_processes,
     )
-    if scorer:
-        mappings = [mapping.with_annotations({'score': scorer(mapping)})
-                    for mapping in mapping_generator]
-    else:
-        mappings = list(mapping_generator)
 
-    network = LigandNetwork(mappings, nodes=nodes)
+    network = network_planner.generate_ligand_network(nodes)
+
     return network
 
 
 def generate_minimal_spanning_network(
     ligands: Iterable[SmallMoleculeComponent],
-    mappers: Union[AtomMapper, Iterable[AtomMapper]],
+    mapper: AtomMapper,
     scorer: Callable[[LigandAtomMapping], float],
     progress: Union[bool, Callable[[Iterable], Iterable]] = True,
+    n_processes: int = 1,
 ) -> LigandNetwork:
     """
     Plan a network with as few edges as possible with maximum total score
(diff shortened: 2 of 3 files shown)


RiesBen:
This PR should be ready for review to merge after this merge happened:
https://github.com/OpenFreeEnergy/konnektor/pull/84

jameseastwood:
Please discuss with @atravitz and @hannahbaumann  as part of the Konnektor handover plans.

jameseastwood:
Needs follow-up discussion in a meeting planned for Oct. 28

openfe/setup/ligand_network_planning.py
```diff
@@ -17,31 +17,48 @@
...
+        seed=core,
+        shift=mapper.shift,
     )
 
 
 def generate_radial_network(
     ligands: Iterable[SmallMoleculeComponent],
-    central_ligand: Union[SmallMoleculeComponent, str, int],
     mappers: Union[AtomMapper, Iterable[AtomMapper]],
+    central_ligand: Union[SmallMoleculeComponent, str, int] = None,
```
IAlibay, line 58: This is the wrong type hint and changes the default behaviour. The assumed default behaviour is that you are required to pass a value for central_ligand - I would not change this for 1.x.

One option is to add `None` to the Union, but not set a default value.

P.S. docstring needs updating.

openfe/setup/ligand_network_planning.py
```diff
@@ -160,7 +167,7 @@ def generate_maximal_network(
     ----------
     ligands : Iterable[SmallMoleculeComponent]
       the ligands to include in the LigandNetwork
-    mappers : AtomMapper or Iterable[AtomMapper]
+    mapper : AtomMapper or Iterable[AtomMapper]
```
IAlibay, line 170: This docstring rename doesn't match the current signature (note that we can't rename from mappers to mapper without causing an API break).

openfe/setup/ligand_network_planning.py
```diff
@@ -169,34 +176,27 @@ def generate_maximal_network(
       progress bar: if False, no progress bar will be shown. If True, use a
       tqdm progress bar that only appears after 1.5 seconds. You can also
       provide a custom progress bar wrapper as a callable.
+    n_processes: int
+        parallelization of network generation.
```
IAlibay, line 180: Here and everywher eelse, could you:
1. adjust the indentation to two rather than four
2. add more details ("number of threads to parallelize network generation" or something a bit more explicit for our users)

openfe/setup/ligand_network_planning.py
```diff
@@ -205,6 +205,7 @@ def generate_minimal_spanning_network(
     mappers: Union[AtomMapper, Iterable[AtomMapper]],
     scorer: Callable[[LigandAtomMapping], float],
     progress: Union[bool, Callable[[Iterable], Iterable]] = True,
+    n_processes: int = 1,
```
IAlibay, line 208: Needs docstring addition.

openfe/setup/ligand_network_planning.py
```diff
@@ -283,46 +279,35 @@ def generate_minimal_redundant_network(
       Minimum Spanning Tree number: the number of minimum spanning trees to
       generate. If two, the second-best edges are included in the returned
       network. If three, the third-best edges are also included, etc.
+    n_processes: int
```
IAlibay, line 282: needs docstring details

openfe/setup/ligand_network_planning.py
```diff
@@ -283,46 +279,35 @@ def generate_minimal_redundant_network(
...
-    return redund_network
+    return network
 
 
 def generate_network_from_names(
-        ligands: list[SmallMoleculeComponent],
-        mapper: AtomMapper,
-        names: list[tuple[str, str]],
+    ligands: list[SmallMoleculeComponent],
+    mappers: Union[AtomMapper, Iterable[AtomMapper]],
```
IAlibay, line 309: The argument rename is an API break, we should do our best to avoid this in a user facing method in 1.x.

openfe/setup/ligand_network_planning.py
```diff
@@ -350,29 +335,21 @@ def generate_network_from_names(
...
-    return generate_network_from_indices(ligands, mapper, ids)
+    return network
 
 
 def generate_network_from_indices(
-        ligands: list[SmallMoleculeComponent],
-        mapper: AtomMapper,
-        indices: list[tuple[int, int]],
+    ligands: list[SmallMoleculeComponent],
+    mappers: Union[AtomMapper, Iterable[AtomMapper]],
```
IAlibay, line 351: As above about API break.

openfe/setup/ligand_network_planning.py
```diff
@@ -397,31 +374,19 @@ def generate_network_from_indices(
...
+    )
     return network
 
 
 def load_orion_network(
-        ligands: list[SmallMoleculeComponent],
-        mapper: AtomMapper,
-        network_file: Union[str, Path],
+    ligands: list[SmallMoleculeComponent],
+    mappers: Union[AtomMapper, Iterable[AtomMapper]],
```
IAlibay, line 388: As above about API break.

openfe/setup/ligand_network_planning.py
```diff
@@ -444,26 +409,33 @@ def load_orion_network(
...
+
+    return network
 
 
 def load_fepplus_network(
-        ligands: list[SmallMoleculeComponent],
-        mapper: AtomMapper,
-        network_file: Union[str, Path],
+    ligands: list[SmallMoleculeComponent],
+    mappers: Union[AtomMapper, Iterable[AtomMapper]],
```
IAlibay, line 437: As above.

IAlibay (changes requested):
Mostly blocking on API breaks - I agree that the name change may make sense in some cases, but these are very much user facing methods and we cannot break the API for 1.x, especially without some kind of deprecation cycle.

openfe/setup/ligand_network_planning.py
```diff
@@ -82,67 +99,57 @@ def generate_radial_network(
...
-            raise ValueError(f"No ligand called '{central_ligand}' "
-                             f"available: {', '.join(l.name for l in ligands)}")
+            raise ValueError(
+                f"No ligand called '{central_ligand}' "
+                f"available: {', '.join(l.name for l in ligands)}"
+            )
         if len(possibles) > 1:
             raise ValueError(f"Multiple ligands called '{central_ligand}'")
         central_ligand = possibles[0]
+        ligands.remove(central_ligand)
```
atravitz, line 130: I think you can move this outside the `elif` (and remove the call on line 113), since you will be removing the central ligand in either case.

openfe/setup/ligand_network_planning.py
```diff
@@ -82,67 +99,57 @@ def generate_radial_network(
...
     # after this, central_ligand is resolved to a SmallMoleculeComponent
     if isinstance(central_ligand, int):
         ligands = list(ligands)
         try:
             central_ligand = ligands[central_ligand]
+            ligands.remove(central_ligand)
         except IndexError:
-            raise ValueError(f"index '{central_ligand}' out of bounds, there are "
-                             f"{len(ligands)} ligands")
+            raise ValueError(
```
atravitz, line 115: I realize this is true of the original code as well, but why are we replacing an `IndexError` with a `ValueError`? Don't we just want an `IndexError` with a custom error message?

atravitz (changes requested):

openfe/setup/ligand_network_planning.py
RiesBen: and remove line 113

openfe/setup/ligand_network_planning.py
RiesBen: this is a fair question, where I don't know the answer to :(

openfe/setup/ligand_network_planning.py
RiesBen: my suggestion would be:
'
n_processes : int  
  provide a number of cpu processes to modulate the cpu-parallelization of the netwokplanning algorithm
'

openfe/setup/ligand_network_planning.py
RiesBen: see above

openfe/setup/ligand_network_planning.py
RiesBen: I see your reasoning, and I think it makes a lot of sense.  
still maybe for a 2.x you could mark this change, as it harmonizes the function calls. (or you have totally different solution)

openfe/setup/ligand_network_planning.py
atravitz: I believe line 112 would throw an error if the `central _ligand` element is not in the list.

openfe/setup/ligand_network_planning.py
IAlibay: I believe the intent is that it's a `ValueError` because the "incorrect value" (i.e. `central_ligand`) was passed.

That being said, I'm pretty sure we just put a ValueError everywhere because it's easier than really thinking about our error types.

openfe/tests/setup/test_network_planning.py
```diff
@@ -130,21 +130,7 @@ def test_radial_network_index_error(toluene_vs_others, lomap_old_mapper):
             ligands=ligands, central_ligand=2077,
             mappers=lomap_old_mapper, scorer=None,
         )
-
-
-def test_radial_network_self_central(toluene_vs_others, lomap_old_mapper):
```
RiesBen, line 135: This can not happen anymore with the current Konnektor and OpenFE implementation.
I propose removing the test.

RiesBen:
I went further through the current CI problems and hope I caught all problems for this PR (@atravitz):
- [x]  ***unreachable unit tests**: there a error messages missing. _-> We need to check why this error is not appearing. This should be thrown at the Konnektor level._
- [x] **generate maximal network**: the unitest behavior expects each edge direction leading to 56 edges. I assume it is either due to directionality of edges or number of possible mappings? However, Konnektor considers edges as non-directed at the current time point and allows only on possible mapper solution == one edge. Therefore it returns 28 edges. Other approaches might lead to problems with the graph algorithms. _-> I would propose to adjust the edge number to 28. After verifying the behavior reason._ 
- [x] **redundant mst**: generates one edge to less then expected?  (expected 14 (2*(n-1)), got 13) _->  need to look into why this is._
- [x] **explicit networks**: The problem here is image you get a list of 8 ligands and 2 defined edges. Currently only the connected part is returned as a ligand network, so two edges with 3 nodes. That clashes with the current unit tests, expecting 3 edges and 8 nodes. This translates to the expectation of getting back a disconnected network as a default. I'm not sure why here it was allowed to generate disconnected networks. _-> I would propose to throw an error in these cases in order to avoid user confusion._

openfe/setup/ligand_network_planning.py
RiesBen: True, but i think the if/elif branches do not cover the case if `central_ligand` is of type `Component`

openfe/tests/setup/test_network_planning.py
IAlibay: By "this cannot happen" do we mean that Konnektor just gracefully handles things such that the network cannot end up with a self edge?

If so, then my suggestion is to keep the code that throws a warning?

If it's more "if you try to do that Konnektor will error", then we should retain the behaviour that pops the ligand out and throws the warning, with a deprecation warning that this behaviour will go away in 2.0.

openfe/setup/ligand_network_planning.py
```diff
@@ -17,31 +17,46 @@
...
+        seed=core,
+        shift=mapper.shift,
     )
 
 
 def generate_radial_network(
     ligands: Iterable[SmallMoleculeComponent],
-    central_ligand: Union[SmallMoleculeComponent, str, int],
     mappers: Union[AtomMapper, Iterable[AtomMapper]],
+    central_ligand: Union[SmallMoleculeComponent, str, int, None],
```
IAlibay, line 56: I wouldn't change the order unless we really have to - i.e. it's technically API breaking in the case that someone just passes arguments without specifying them?

openfe/setup/ligand_network_planning.py
IAlibay: @atravitz you marked this as resolved but the docstring still doesn't match - was this intentional?

IAlibay (changes requested):
partial review

openfe/setup/ligand_network_planning.py
```diff
@@ -82,67 +103,57 @@ def generate_radial_network(
...
+        mappers=mappers,
+        scorer=scorer,
+        progress=progress,
+        n_processes=n_processes,
+    )
 
-    edges = []
-
-    for ligand in ligands:
-        if ligand == central_ligand:
```
IAlibay, line 110: What happens in the new code if you pass in multiple copies of the same central ligand in the ligand list?

openfe/setup/ligand_network_planning.py
```diff
@@ -160,7 +171,7 @@ def generate_maximal_network(
     ----------
     ligands : Iterable[SmallMoleculeComponent]
       the ligands to include in the LigandNetwork
-    mappers : AtomMapper or Iterable[AtomMapper]
+    mapper : AtomMapper or Iterable[AtomMapper]
```
IAlibay, line 174: revert to mappers?

IAlibay (changes requested):

openfe/setup/ligand_network_planning.py
atravitz: good point, reverted.

openfe/setup/ligand_network_planning.py
atravitz: it throws the `The central component __ is present in the list of components` twice, and removes them both from the ligand list. I think this is fine?

environment.yml
```diff
@@ -7,6 +7,7 @@ dependencies:
   - coverage
   - duecredit<0.10
   - kartograf>=1.0.0
+  # - konnektor
```
IAlibay:
```suggestion
```

openfe/tests/setup/alchemical_network_planner/test_relative_alchemical_network_planner.py
```diff
@@ -33,9 +33,9 @@ def test_rbfe_alchemical_network_planner_call(atom_mapping_basic_test_files, T4_
         solvent=SolventComponent(),
         protein=T4_protein_component,
     )
-
+    # TODO:this network is slightly different when using konnektor,
+    # possibly because of a score tie in _map_scoring,
```
IAlibay: Is this still true? If so, could you add the issue number here?

openfe/tests/setup/test_network_planning.py
```diff
@@ -310,6 +312,7 @@ def test_generate_maximal_network(
     lomap_old_mapper,
     simple_scorer,
 ):
+    """TODO: add test for throwing an error if there is no scorer but multiple mappings"""
```
IAlibay: raise issue?

openfe/tests/setup/test_network_planning.py
```diff
@@ -599,6 +599,7 @@ def test_network_from_indices(self, atom_mapping_basic_test_files, lomap_old_map
 
         requested = [(0, 1), (2, 3)]
 
+        # TODO: konnektor only constructs based on edges, doesn't allow for disconnected networks
```
IAlibay: Is this still a thing?

IAlibay (approved):
Overall lgtm, couple of questions - mostly raise issues on todos if they don't exist

openfe/tests/setup/test_network_planning.py
atravitz: I added this test back in, which checks that a warning is appropriately raised, and that a self-edge is not created.

openfe/tests/setup/test_network_planning.py
atravitz: no it is not, it's been updated in [konnektor v0.2.0](https://konnektor.openfree.energy/en/latest/CHANGELOG.html) - deleting this TODO.

openfe/tests/setup/test_network_planning.py
atravitz: I decided not to implement this in this PR since it would be a breaking change, and the purpose of this PR is to integrate konnektor with the minimal necessary changes.

There's a konnektor issue for making this behavior more strict, and the combination of `extra_mapper=True` and `with_scorer=False` exists here, so I'm just going to delete this comment.

openfe/tests/setup/alchemical_network_planner/test_relative_alchemical_network_planner.py
atravitz: nope, this was addressed here: https://github.com/OpenFreeEnergy/konnektor/pull/159. 

removing this todo!

atravitz (approved):

atravitz:
package install tests happening here: https://github.com/OpenFreeEnergy/openfe/actions/runs/15714814779/job/44281799333

update: and have passed! 🥳

news/konnektor_changes.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* Added optional ``progress`` (whether to show a progress bar) and ``n_processes`` (number of parallel processes to use when generating the network) arguments for network planners.
```
IAlibay: What's the behaviour when we use the CLI?

atravitz: CLI uses the default, which hasn't changed at all (``n_process=1`` and ``progress=True``, except radial network, which didn't have a ``progress`` option before.

I updated so hopefully it's clearer now.

PR #929, opened 2024-09-03
Enable `PersesAtomMapper`

jthorton:
Add the missing abstract methods for the `PersesAtomMapper` and turn on the tests.
Checklist
* [ ] Added a ``news`` entry

## Developers certificate of origin
- [x] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

PR #930, opened 2024-09-12
Perses mapper

jthorton:
Replaces #929 to enable openeye tests

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/setup/atom_mapping/perses_mapper.py
+++ b/openfe/setup/atom_mapping/perses_mapper.py
@@ -6,8 +6,10 @@ The MCS class from Perses shamelessly wrapped and used here to match our API.
 
 """
 
-from openmm import unit
 from openfe.utils import requires_package
+from openff.models.types import FloatQuantity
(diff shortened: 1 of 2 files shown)


IAlibay:
Not sure why codecov isn't picking up the extra test, let's see if bumping the PR helps.

IAlibay (changes requested):
@jthorton could you add a `news` entry please? (since it's technically changing behaviour from omm to off units). Otherwise it looks good to me!

P.S. macos failure is something else, issue has been raised upstream.

jthorton:
thanks @IAlibay I have added it should it show in the docs build as part of the change log?

IAlibay:
> thanks @IAlibay I have added it should it show in the docs build as part of the change log?

I believe it will only show up when we call rever (we probably should fix that at some point?), so it should be good for now!

IAlibay (approved):
lgtm!

PR #937, opened 2024-09-16
Hotfix: structural analysis

### LINKED ISSUE #926: `ProtocolDAGResult`s produced by `RelativeHybridTopologyProtocol` can grow very large in typical usage
A `ProtocolDAGResult` produced by the `RelativeHybridTopologyProtocol` for one of ASAP Discovery's lead optimization programs, performed on `alchemiscale` using `openfe` 1.0.1, measures ~32 MiB when serialized directly to JSON. This is rather large on its own, but results in extremely slow result retrieval for an `AlchemicalNetwork` with ~3000 of these objects, resulting in ~95 GiB of memory usage to load them all at once. Many user machines would not be able to manage this.

We did not see this kind of behavior in `openfe` 0.14.0, indicating that this may be due to the `structural_analysis` outputs that were added in the releases following this. This gives RMSDs for various components over time, for every replica, and at least appears to be the largest new data structure in the `ProtocolDAGResult` object.

There are several options available to address this, none of which are mutually exclusive:
1. Expose a setting for this protocol allowing users to set the frequency with which RMSDs are recorded in the `structural_analysis` outputs, including setting the frequency to `0` (disabling them completely).
2. Make the `structural_analysis` outputs `numpy` arrays instead of lists, then convert to bytes->base64 encoding. Will also help with speeding up [KeyedChain](https://github.com/OpenFreeEnergy/gufe/pull/286) representation generation, which will matter if this protocol uses more than one `ProtocolUnit` with dependencies in the future.
3. Make the `RelativeHybridTopologyProtocolResult` object take only what it needs from the `ProtocolDAGResult` `ProtocolUnitResult`s it aggregates, instead of holding on to those `ProtocolUnitResult`s in memory. This would allow the `ProtocolUnitResult`s to be garbage collected more easily while retaining the `RelativeHybridTopologyProtocolResult` a user likely wants.

A more invasive change would be to move `structural_analysis` outputs to their own file, or if they are already recorded in file outputs, to keep them there. This has the downside that this data would not be readily available to users of `alchemiscale` until result file retrieval is possible with that system.

IAlibay:
Fixes #926 

TODO:
- [x] Move structural analysis writing to npz
- <s>Write results accessor</s>
- [x] Tests
- [x] News
- [x] Fix CI

### DIFF
--- a/openfe/protocols/openmm_rfe/equil_rfe_methods.py
+++ b/openfe/protocols/openmm_rfe/equil_rfe_methods.py
@@ -36,8 +36,6 @@ from openff.units.openmm import to_openmm, from_openmm, ensure_quantity
 from openff.toolkit.topology import Molecule as OFFMolecule
 from openmmtools import multistate
 from typing import Optional
-from openmm import unit as omm_unit
-from openmm.app import PDBFile
 import pathlib
 from typing import Any, Iterable, Union
 import openmmtools
@@ -48,7 +46,7 @@ from rdkit import Chem
 import gufe
 from gufe import (
     settings, ChemicalSystem, LigandAtomMapping, Component, ComponentMapping,
-    SmallMoleculeComponent, ProteinComponent, SolventComponent,
+    SmallMoleculeComponent, SolventComponent,
 )
 
 from .equil_rfe_settings import (
--- a/openfe/protocols/openmm_rfe/equil_rfe_methods.py
+++ b/openfe/protocols/openmm_rfe/equil_rfe_methods.py
@@ -1101,7 +1101,7 @@ class RelativeHybridTopologyProtocolUnit(gufe.ProtocolUnit):
         analysis_out = scratch / 'structural_analysis.json'
 
         ret = subprocess.run(['openfe_analysis', 'RFE_analysis',
-                              str(scratch), str(analysis_out)],
+                              str(shared), str(analysis_out)],
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
         if ret.returncode:
--- a/openfe/protocols/openmm_rfe/equil_rfe_methods.py
+++ b/openfe/protocols/openmm_rfe/equil_rfe_methods.py
@@ -1096,13 +1096,13 @@ class RelativeHybridTopologyProtocolUnit(gufe.ProtocolUnit):
             return {'debug': {'sampler': sampler}}
 
     @staticmethod
-    def analyse(where) -> dict:
+    def structural_analysis(scratch, shared) -> dict:
         # don't put energy analysis in here, it uses the open file reporter
         # whereas structural stuff requires that the file handle is closed
-        analysis_out = where / 'structural_analysis.json'
+        analysis_out = scratch / 'structural_analysis.json'
 
         ret = subprocess.run(['openfe_analysis', 'RFE_analysis',
-                              str(where), str(analysis_out)],
+                              str(scratch), str(analysis_out)],
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
         if ret.returncode:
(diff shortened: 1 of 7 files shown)


openfe/tests/protocols/test_openmm_rfe_slow.py
```diff
@@ -84,7 +90,9 @@ def test_openmm_run_engine(benzene_vacuum_system, platform,
         nc = pur.outputs['nc']
         assert nc == unit_shared / "simulation.nc"
         assert nc.exists()
-        assert (unit_shared / "structural_analysis.json").exists()
+        structural_analysis_file = unit_shared / "structural_analysis.npz"
```
IAlibay, line 93: Coverage is going to fail because we don't run these tests until merge.

Can confirm that this passes locally.

IAlibay:
@hannahbaumann @dotsdl - I _think_ this should work, fixing the giant JSON results file issue whilst keeping the same things around as before (there's deeper issues with structural analysis correctness, but we'll need to deal with this later).

openfe/tests/protocols/test_openmm_rfe_slow.py
hannahbaumann: Would it make sense to also test the failure of the `structural_analysis`?

hannahbaumann (commented):
Thanks, @IAlibay , just the one comment, other than that it looks good to me!

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -1096,13 +1094,14 @@ def run(self, *, dry=False, verbose=True,
...
+    def structural_analysis(scratch, shared) -> dict:
         # don't put energy analysis in here, it uses the open file reporter
         # whereas structural stuff requires that the file handle is closed
-        analysis_out = where / 'structural_analysis.json'
+        # TODO: we should just make openfe_analysis write an npz instead!
+        analysis_out = scratch / 'structural_analysis.json'
 
         ret = subprocess.run(['openfe_analysis', 'RFE_analysis',
-                              str(where), str(analysis_out)],
+                              str(shared), str(analysis_out)],
```
dotsdl, line 1104: Could break these list items into one on each line, put a comment next to each as to what the arguments are.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -1124,21 +1123,43 @@ def analyse(where) -> dict:
...
+            ),
+            ligand_COM_drift=np.asarray(
+                data["ligand_wander"], dtype=np.float32
+            ),
+            protein_2D_RMSD=np.asarray(
+                data["protein_2D_RMSD"], dtype=np.float32
+            ),
+            time_ps=np.asarray(
+                data["time(ps)"], dtype=np.int32
+            ),
```
dotsdl, line 1143: Possible to walk through keys of `data` to avoid hardcodes of these fields?

dotsdl (approved):
LGTM! Some comments to address, but none major or blocking. Thanks @IAlibay!

openfe/protocols/openmm_rfe/equil_rfe_methods.py
IAlibay: Done, thanks!

openfe/protocols/openmm_rfe/equil_rfe_methods.py
IAlibay: "ligand_wander" and "time(ps)" are the main issues here. We should revisit this in a future PR where we clean up structural analysis further.

openfe/tests/protocols/test_openmm_rfe_slow.py
IAlibay: Done, thanks!

PR #943, opened 2024-09-27
fix: avoid nvidia-smi query if the process will get killed

### LINKED ISSUE #854: CalledProcessError: 9
Hi, 

I tried to run openFE tutorial on my laptop and everything worked just fine, but when I tried to run it on our cluster I faced an issue. On a gpu node it gave an error that has been mentioned before (GPU in 'Exclusive_Process' mode (or Prohibited), one context is allowed per device. This may prevent some openmmtools features from working. GPU must be in 'Default' compute mode). While we fix this issue, I wanted to run it without the gpu, but this led to another error:

$ openfe quickrun transformations/easy_rbfe_lig_ejm_31_solvent_lig_ejm_46_solvent.json -o results/easy_rbfe_lig_ejm_31_solvent_lig_ejm_46_solvent_login_node.json -d results/easy_rbfe_lig_ejm_31_solvent_lig_ejm_46_solvent_login_node

Loading file...
Planning simulations for this edge...
Starting the simulations for this edge...
Done with all simulations! Analyzing the results....
Here is the result:
	dG = None ± None

Error: The protocol unit 'lig_ejm_31 to lig_ejm_46 repeat 2 generation 0' failed with the error message:
CalledProcessError: 9

Details provided in output.

The only output is the .json file that is attached.

Cheers,
Raitis

[easy_rbfe_lig_ejm_31_solvent_lig_ejm_46_solvent_no_gpu.json](https://github.com/OpenFreeEnergy/openfe/files/15347656/easy_rbfe_lig_ejm_31_solvent_lig_ejm_46_solvent_no_gpu.json)

PabloNA97:
Avoid GPU query whenever running in a SLURM environment without GPU. If the node doesn't have NVIDIA drivers this query will fail and the process will get killed.

fixes #854

### DIFF
--- a/openfe/utils/system_probe.py
+++ b/openfe/utils/system_probe.py
@@ -273,6 +273,55 @@ def _get_hostname() -> str:
     return socket.gethostname()
 
 
+def _slurm_environment() -> bool:
+    """
+    Check if the current environment is managed by SLURM.
+    """
+    
+    slurm_job_id = os.environ.get("SLURM_JOB_ID")
+    
+    if slurm_job_id:
+        return True
+    else:
+        return False
+
+
+def _check_slurm_gpu_info() -> bool:
+    """
+    Check if the GPU information is available in the SLURM environment.
+
+    Returns
+    -------
+    bool
+        True if the GPU information is available in the SLURM environment, False
+        otherwise.
+
+    Notes
+    -----
+    This function checks if the GPU information is available in the SLURM environment by
+    inspecting the environment variables.
+
+    The function returns True if any of the following environment variables are present:
+    - 'SLURM_JOB_GPUS'
+    - 'SLURM_GPUS'
+    - 'CUDA_VISIBLE_DEVICES'
+    
+    Otherwise, it returns False.
+    """
+    
+    slurm_job_gpus = os.environ.get("SLURM_JOB_GPUS")
+    slurm_gpus = os.environ.get("SLURM_GPUS")
+    cuda_visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES")
+    
+    logging.debug(f"SLURM_JOB_GPUS: {slurm_job_gpus}")
+    logging.debug(f"SLURM_GPUS_PER_NODE: {slurm_gpus}")
+    logging.debug(f"CUDA_VISIBLE_DEVICES: {cuda_visible_devices}")
+    
+    if slurm_job_gpus or slurm_gpus or cuda_visible_devices:
+        return True
+    else:
+        return False
+    
 def _get_gpu_info() -> dict[str, dict[str, str]]:
     """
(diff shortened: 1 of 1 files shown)


openfe/utils/system_probe.py
```diff
@@ -273,6 +273,55 @@ def _get_hostname() -> str:
...
+        otherwise.
+
+    Notes
+    -----
+    This function checks if the GPU information is available in the SLURM environment by
+    inspecting the environment variables.
+
+    The function returns True if any of the following environment variables are present:
+    - 'SLURM_JOB_GPUS'
+    - 'SLURM_GPUS'
```
IAlibay, line 306: I'm afraid these might be very specific to some deployments of SLURM (at least I know in the clusters I've managed in the past we didn't have these env variables set).

IAlibay (changes requested):
Thanks @PabloNA97 - I'm going to put a block on this because we need to discuss this a bit on our end. My initial take is that I'm not sure that this is the solution to the deeper issue of "why are we even running this if there's no GPU".

I'm also not sure I understand why slurm behaves differently here, having a special case like this shouldn't be necessary.

Ideally we should only run this _if_ a user has requested a GPU.

mikemhenry:
I've made a comment on the issue but just to add to this PR, the heuristic of "this is slurm so we don't want to run nvidia-smi" is a bad one since many people are using this software on slrurm systems and *don't* have this issue. I rather we just improve our try/except on running the command, then trying to decide if we should run it.

openfe/utils/system_probe.py
PabloNA97: Thanks for the feedback, appreciate it!  :D
I applied this changes on my end to test it with CPU. But if using CPU for production is a bad idea in itself then it is not worth it to merge in the main branch.

PabloNA97:
> I've made a comment on the issue but just to add to this PR, the heuristic of "this is slurm so we don't want to run nvidia-smi" is a bad one since many people are using this software on slrurm systems and _don't_ have this issue. I rather we just improve our try/except on running the command, then trying to decide if we should run it.

The behaviour of the original code only changes whenever you (1) run on slurm and (2) there is no gpu requested. In that case I thought it didn't make sense to run nvidia-smi. But as @IAlibay explained maybe it doesn't make sense to contemplate this case after all.

openfe/utils/system_probe.py
PabloNA97: Not sure this is mentioned in the openfe docs - maybe it would be worth to include a small comment about it if it's not.

nffaruk:
Didn't realize there was a PR about this already, I posted a possible fix in Issues #1178 that uses an additional exception to handle this instead. If openfe doesn't scale very well with multiple CPU cores, then I agree that it's not so useful, but my team has been considering it for when GPU resources are tied up.

Issue #826, opened 2024-04-15
Request: AMD HIP Platform Support


HiteSit:
Hi all,

I was wondering if is possible to add the support for the AMD HIP platform.
Specifically OpenMM already is able to work on HIP platforms installing a specific hack with Conda:

`mamba install jaimergp/label/unsupported-cudatoolkit-shim::cudatoolkit=11.2.2 && mamba install streamhpc::openmm-hip=8.0.0`

But the problem is that in your code, specifically under openmmutils/utils.py there is an `assert` that allows only `["OpenCL", "CUDA"]`, probably the fix is easy since as said OpenMM is able to see HIP as platform as well as give it the "right" speed.

The fix would have a huge impact since the LUMI HPC (most powerful in europe) only supports HIP platform.

mikemhenry:
@HiteSit Thank you for raising this issue!

"We" (not the OpenFE team but other orgs I am apart of) are working to get ROCm/HIP onto conda-forge so that no hacks will be needed to install openmm.

Can you link the line of code that has this assert? Also do you know how OpenMM reports the HIP platform string-wise? I would be happy to get this working.

HiteSit:
So,
the utils file is under:
`/mambaforge/envs/cheminf_3_11/lib/python3.11/site-packages/openmmtools/utils/utils.py`

```
def platform_supports_precision(platform, precision):
    """Determine whether the specified OpenMM Platform supports the specified minimum precision.

    Parameters
    ----------
    platform : str or openmm.Platform
        The platform or platform name to check
    precision : str
        One of ['single', 'mixed', 'double']

    Returns
    -------
    is_supported : bool
        True if the platform supports the specified precision; False otherwise
    """
    SUPPORTED_PRECISIONS = ['single', 'mixed', 'double']
    assert precision in SUPPORTED_PRECISIONS, f"Precision {precision} must be one of {SUPPORTED_PRECISIONS}"

    if isinstance(platform, str):
        # Get the actual Platform object if the platform_name was specified
        platform = openmm.Platform.getPlatformByName(platform)

    if platform.getName() == 'Reference':
        # Reference is double precision
        return (precision == 'double')

    if platform.getName() == 'CPU':
        return precision in ['mixed']

    if platform.getName() in ['CUDA', 'OpenCL']:
        properties = { 'Precision' : precision }
        system = openmm.System()
        system.addParticle(1.0) # Cannot create Context on a system with no particles
[...]

PR #951, opened 2024-10-08
Support HIP

mikemhenry:
Will use this PR to track what we need to support HIP. Related PRs and Issues:
https://github.com/OpenFreeEnergy/openfe/issues/826
https://github.com/choderalab/openmmtools/pull/753

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/protocols/openmm_rfe/_rfe_utils/compute.py
+++ b/openfe/protocols/openmm_rfe/_rfe_utils/compute.py
@@ -35,6 +35,7 @@ def get_openmm_platform(platform_name=None):
                 'cpu': 'CPU',
                 'opencl': 'OpenCL',
                 'cuda': 'CUDA',
+                'hip': 'HIP',
             }[str(platform_name).lower()]
         except KeyError:
             pass
@@ -43,10 +44,10 @@ def get_openmm_platform(platform_name=None):
         platform = Platform.getPlatformByName(platform_name)
     # Set precision and properties
     name = platform.getName()
-    if name in ['CUDA', 'OpenCL']:
+    if name in ['CUDA', 'OpenCL', 'HIP']:
         platform.setPropertyDefaultValue(
                 'Precision', 'mixed')
-    if name == 'CUDA':
+    if name in ['CUDA', 'HIP']:
         platform.setPropertyDefaultValue(
                 'DeterministicForces', 'true')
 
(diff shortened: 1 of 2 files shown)


mikemhenry:
Will keep adding to this PR as needed, will also expand the news entry when the dust settles

IAlibay:
@mikemhenry do you want to add the allowable platforms list as a validator to the OpenmmEngineSettings class? I feel like we should document what we support somewhere, and it's probably as good a place to go "well you can do this but know that OpenCL and HIP aren't thoroughly validated.

mikemhenry:
@IAlibay that is a good point! I will add that and also add a note somewhere in the docs, maybe we need a page that is like "Using OpenFE in Production" where we list what we have heavily validated and also list tips for using it, like generating all the partial charges before running transformations.

openfe/protocols/openmm_rfe/_rfe_utils/compute.py
```diff
@@ -50,7 +50,7 @@ def get_openmm_platform(platform_name=None):
         platform.setPropertyDefaultValue(
                 'DeterministicForces', 'true')
 
-    if name != 'CUDA':
```
IAlibay, line 53: Do we need to set DeterministicForces or Percision for HIP?

mikemhenry: Good catch, I was able to do that with the OpenMM API, but I need to test on a real GPU to see if it works (as in, setting those properties worked on my openMM HIP build but it could crash as soon as I try and run a MD step on the GPU)

IAlibay: Is there an easy way to test this out? I.e. would spinning up an AWS runner that uses an AMD card work?

mikemhenry:
So we don't currently have any validators:

https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/protocols/openmm_utils/omm_settings.py#L300-L313

A literal would be my preference -- users can deal with errors of using `cuda` instead of `CUDA` since the error happens as soon as they set the setting and not at simulation run time, BUT I do think this would be considered an API change. If I use a validator I can then at least use it to warn users if they choose an option that won't work or choose an option that isn't well validated

IAlibay:
> So we don't currently have any validators:
> 
> https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/protocols/openmm_utils/omm_settings.py#L300-L313
> 
> A literal would be my preference -- users can deal with errors of using `cuda` instead of `CUDA` since the error happens as soon as they set the setting and not at simulation run time, BUT I do think this would be considered an API change. If I use a validator I can then at least use it to warn users if they choose an option that won't work or choose an option that isn't well validated

I personally would prefer it if it was a validator - mainly you get things like "OpenCL"  | "opencl".
Using a literal is too strict, we just need the lowercase to match in the protocol.

jameseastwood:
Addressed in #2127

PR #955, opened 2024-10-11
move slow import inside function call

atravitz:
For cli commands, slow imports should be moved inside functions (as recommended [here](https://github.com/OpenFreeEnergy/openfe/tree/main/openfecli/commands#delay-slow-imports)). 

By moving the `plan_alchemical_network_output` import from the top of the file to inside the functions for rbfe and rhfe  network planning, we avoid importing `openfe` during the command-gathering process.  

The `--help` and `--version` calls now take 0.15 seconds instead of ~4 seconds.

Ideally, we will also speed up `import openfe`, which was the root cause of this slow behavior, in another PR.

Checklist
* [ ] Added a ``news`` entry


IAlibay (approved):
lgtm thanks!

PR #1009, opened 2024-11-08
Move CLI README so that it's easier to find

atravitz:
I think it's good practice to keep READMEs toward the top of the directory structure.


IAlibay (approved):
meant to approve sorry

atravitz:
> I can see why it was there, and I agree with the move.
> 
> Should we open an issue to discuss if we should move it to the sphinx docs?

I don't think it should live in sphinx docs yet, I want to make sure it's more external-user-friendly first.

PR #1022, opened 2024-11-19
add example notebooks bump to task list

atravitz:
adding to the checklist, since we forgot to bump the example notebooks for v1.2.0 release


IAlibay (approved):

IAlibay (commented):
Long term, maybe we should be using a global environment file for this kind of stuff?

Issue #986, opened 2024-10-28
input validation for directory/json names


atravitz:
If a user provides a json file name formatted as `[non-existent directory]/filename.json`, the code runs without error, but the json file(s) are never created. This results in the user needing to rerun costly simulations.

The expected behavior is to throw an error for this up front, or create the directory on the fly

IAlibay:
Yes, I also think it's might be possible to overwite the output JSON (I haven't tested this, this is purely based on what I've been getting as reports from other folks).

mikemhenry:
I think we should modify the behavior slightly to this:

1. We shouldn't overwrite output jsons, and we should do this check at the start and not after a simulation is done and we are writing the results
2. I don't think we need to require the `-d directory` to be empty since even if every transformation used the same directory, the simulations shouldn't end up overwriting each other in the shared-xxx and scratch-xxx dire
3. We should create any directories we need to for the `-d` and `-o` flags if they don't exist. We should create them before we actually execute the dag, so that if we run into any issues (like permissions or something) it fails fast before we waste time running the dag and then try to save the results at the end.

As it stands, the current behavior breaks how things worked in 1.0.1. I ran into this when running some jobs from Hannah using the submission scripts from the industry benchmark since I like to programmatically create the output directories like `transformation_results_$n` where `n` is the replica number.

This was the error I got:
```
Usage: openfe quickrun [OPTIONS] TRANSFORMATION
Try 'openfe quickrun -h' for help.

Error: Invalid value for '-o': Cannot write to results_transformations_tyk2_0/rbfe_ejm_42_ejm_54.json, parent directory does not exist.
done
```

I will say the error message is excellent (thanks @atravitz ) since it clearly explained what the error was and what I needed to do. To me it feels too much like an extra hoop to jump through to add a `mkdir transformation_results_$n` before my quick run command.

atravitz:
thanks @mikemhenry  for flagging that this breaks current behavior! 

1. This is the current behavior, unless I'm misunderstanding you. If a .json output file already exists, an error is thrown up-front.
2. We don't currently require the `-d` directory to empty, so I think this is fine as-is.
3. I agree with creating `-d` if it doesn't exist (the current behavior). However, there's a complication where I believe we don't want to create the `-o` file until it's time to write, since an empty or incomplete file would change current behavior. Both `openfe gather` and users (I think) use the existence of a file as an indicator the simulation is running or complete.

mikemhenry:
[...]

PR #1024, opened 2024-11-21
986 input validation for directory names

atravitz:
addresses #986 

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/news/986_input_validation.rst
+++ b/news/986_input_validation.rst
@@ -4,7 +4,7 @@
 
 **Changed:**
 
-* ``openfe quickrun`` now fails up-front when the user-supplied output file (``-o``) is invalid.
+* ``openfe quickrun`` now fails up-front when the user-supplied output file (``-o``) already exists or has a non-existent parent directory.
 
 **Deprecated:**
 
--- a/news/986_input_validation.rst
+++ b/news/986_input_validation.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* ``openfe quickrun`` now fails up-front when the user-supplied output file (``-o``) is invalid.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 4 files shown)


openfecli/commands/quickrun.py
```diff
@@ -33,10 +33,9 @@ def _format_exception(exception) -> str:
 )
 @click.option(
     'output', '-o', default=None,
-    type=click.Path(dir_okay=False, file_okay=True, writable=True,
```
atravitz, line 36: I removed `writable=True` here, since click ignores any further parameters if the file does not exist (and here we require that the file doesn't exist). 

Similarly, we should be able to just use `file_okay=False` instead of our own custom check in `ensure_file_does_not_exist()`

news/986_input_validation.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* ``openfe quickrun`` now fails up-front when the user-supplied output file (``-o``) is invalid.
```
IAlibay, line 7: Could you maybe add a bit more detail here on what "invalid" means (i.e. already exists and/or the directory to the file doesn't exist)

openfecli/commands/quickrun.py
```diff
@@ -33,10 +33,9 @@ def _format_exception(exception) -> str:
...
-                    path_type=pathlib.Path),
+    type=click.Path(dir_okay=False, file_okay=True, path_type=pathlib.Path),
     help="output file (JSON format) for the final results",
```
IAlibay, line 37: Can you add extra info in the help docstring (or maybe there's a more appropriate place to do it?) to detail the behaviour we expect from this flag?

IAlibay (changes requested):
This looks good to me, just a bit more documentation would be great though!

openfecli/commands/quickrun.py
atravitz: I rewrote this to make it more precise - what additional info do you think is needed here?

IAlibay (approved):
Thanks, lgtm!

openfecli/commands/quickrun.py
IAlibay: I was something like "file must not already exist" or whatever the standard text is for CLI tools - what you have is fine, i.e. now that it says "create"

PR #1036, opened 2024-12-05
include gufe main and latest in ci matrix

atravitz:
CI is currently failing because of changes on gufe `main`.  We should (at least for now) run CI for both gufe `main` and `latest`, since we expect it to pass on `latest` but fail on `main`.


IAlibay (changes requested):
Couple of things:

1. If you do this, then you have to ammend the env yaml file
2. This will use up all our runners when we do a merge with slow tests - i.e. it's too many runners.

IAlibay:
See https://github.com/OpenFreeEnergy/openfe/actions/runs/12186104840/job/33993854139?pr=1036#step:9:617

Even though it's set at latest, the issue you're facing is that you've already pip installed gufe, so mamba attempts to installs, finds that you already have a working version and does nothing.

.github/workflows/ci.yaml
```diff
@@ -96,7 +101,8 @@ jobs:
           micromamba install -c openeye openeye-toolkits
           python -c "import openeye; assert openeye.oechem.OEChemIsLicensed(), 'oechem license check failed!'"
 
-      - name: "Install GUFE from main@HEAD"
+      - name: "Install dev version of GUFE."
```
IAlibay, line 104: This is much better, thanks!

IAlibay (commented):
The main question I have is "how long are we planning to do this".

One key issue I could see is that openfe adapts to match a behaviour from gufe (see for example when we dropped the key stability tests) and then "latest" fails, but that's intentional because we've fixed things to match with upstream.

atravitz:
@mikemhenry I see there's api-breaking changes flagged here - where can I get more details on this?

IAlibay (approved):
this looks good thanks!

.github/workflows/ci.yaml
```diff
@@ -41,22 +41,34 @@ jobs:
...
+            gufe-version: 'a9b5982471eb3644ca8c4423c7001e6309a3f544'
           - os: "macos-latest"
             python-version: "3.12"
             pydantic-version: ">1"
             omff-version: ">0.13"
             openeye: ["no"]
+            gufe-version: 'a9b5982471eb3644ca8c4423c7001e6309a3f544'
+
+          # temporary test against gufe main while we're pinned to a9b5
+          - os: "ubuntu-latest"
```
atravitz, line 67: @IAlibay do you have an opinion about what build to test here?

IAlibay: I'm not sure I fully understand what you mean - this seems fine to me!

atravitz: just what combination of os, openeye, etc. would be most useful, since we're just choosing one here.

IAlibay: This is good, it covers the widest range of tests :)

Issue #987, opened 2024-10-28
gather should support simulation repeats submitted in parallel


atravitz:
When running quickrun in parallel, the directory structure changes so to be `results_0/`, `results_1/`, `results_2/` , and gather no longer works

We need to decide if gather should support this scenario, or if we need to be more strict about user-behavior.

CurtisColwell:
Glad this is already here. I am running into this problem now. I changed the protocol_repeats to 1 to run each repeat separately, but it doesn't add to the currently existing .json output. Can I add the new 'protocol_result' and 'unit_results' or will this not work?

I nice solution would be adding the new repeat to the old output file instead of throwing this error:

**Error: Invalid value for '-o': File 'results/easy_rbfe_mol1_solvent_mol2_solvent.json' already exists.**

PR #1040, opened 2024-12-06
create test data for quickrun submitted in parallel

atravitz:
I re-arranged the data in `openfecli/test/data/results` to mimic the results from a `quickrun` submission where every replicate is separate. I'll use this dataset to test the solution to [this issue](https://github.com/OpenFreeEnergy/openfe/issues/987)

I included the notebook I used to generate this data since I expect we might want to use it again in the future. It's not polished, but I think it's good to have some record of where this test data came from.

### DIFF
--- a/openfecli/tests/commands/test_gather.py
+++ b/openfecli/tests/commands/test_gather.py
@@ -28,13 +28,23 @@ def test_get_column(val, col):
 
 
 @pytest.fixture
-def results_dir(tmpdir):
+def results_dir_serial(tmpdir):
+    """Example output data, with replicates run in serial (3 replicates per results JSON)."""
     with tmpdir.as_cwd():
         with resources.files('openfecli.tests.data') as d:
             t = tarfile.open(d / 'rbfe_results.tar.gz', mode='r')
             t.extractall('.')
-
         yield
+
+@pytest.fixture
+def results_dir_parallel(tmpdir):
+    """Identical data to results_dir_serial(), with replicates run in parallel (1 replicate per results JSON)."""
+    with tmpdir.as_cwd():
+        with resources.files('openfecli.tests.data') as d:
+            t = tarfile.open(d / 'results_parallel.tar.gz', mode='r')
+            t.extractall('.')
+        yield
+
 
 _EXPECTED_DG = b"""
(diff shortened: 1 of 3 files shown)


openfecli/tests/commands/test_gather.py
```diff
@@ -28,14 +28,25 @@ def test_get_column(val, col):
...
+    """Example output data, with replicates run in serial (3 replicates per results JSON)."""
     with tmpdir.as_cwd():
         with resources.files('openfecli.tests.data') as d:
             t = tarfile.open(d / 'rbfe_results.tar.gz', mode='r')
             t.extractall('.')
 
         yield
 
+@pytest.fixture
+def results_dir_parallel(tmpdir):
```
atravitz, line 41: I don't actually use this fixture in this PR, since I just want to add the test data here and not add any tests that would fail.

IAlibay:
@atravitz can you confirm that you loaded these back into ProtocolResults objects and it reads fine?

openfecli/tests/data/restructure_results_data.ipynb
```diff
@@ -0,0 +1,157 @@
...
+    "    assert srckey_to_protocol.keys() == srckey_to_unit_results.keys()\n",
+    "    \n",
+    "    ## write to the new directory\n",
+    "    for n, sk in enumerate(sorted(srckey_to_protocol.keys())):\n",
+    "        rep_dir = new_dir/f\"replicate_{n}\"\n",
+    "        os.makedirs(rep_dir/leg)\n",
+    "    \n",
+    "        # build up the data for this replicate\n",
+    "        replicate_data = {'estimate': srckey_to_estimate[sk],\n",
+    "                          'uncertainty': 0.0,\n",
```
IAlibay, line 123: uncertainty should be an openff.unit.Quantity (compatible with kcal/mol)

IAlibay (changes requested):
Just the one thing, otherwise - if it loads back up properly - then it looks good to me.

jthorton (approved):
LGTM!

atravitz:
> @atravitz can you confirm that you loaded these back into ProtocolResults objects and it reads fine?

Both the `protocol_result` and `unit_results` load correctly

PR #1041, opened 2024-12-06
Add gather docstrings

atravitz:
First pass at adding some docstrings and type hints.

### DIFF
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -4,11 +4,26 @@
 import click
 from openfecli import OFECommandPlugin
 from openfecli.clicktypes import HyphenAwareChoice
+
+import os
 import pathlib
 import warnings
 
 
-def _get_column(val):
+def _get_column(val:float|int)->int:
+    """Determine the index (where the 0th index is the decimal) at which the
+    first non-zero value occurs in a full-precision string representation of a value.
+
+    Parameters
+    ----------
+    val : float|int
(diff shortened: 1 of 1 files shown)


openfecli/commands/gather.py
```diff
@@ -64,7 +121,9 @@ def get_names(result) -> tuple[str, str]:
         return toks[0], toks[2]
 
 
-def get_type(res):
+def get_type(res:dict)->Literal['vacuum','solvent','complex']:
+    """Determine the simulation type based on the component types."""
```
IAlibay, line 125: Maybe here it's worth being a bit explicit, it's determining the simulation type based on the key indices of the components.

The reason for being clearer here is that this is super super flaky / bad practice! We don't enforce key names in the RFE protocol so this won't work if you use the Python API to set up your transformations and then decided to use gather.

I didn't realise this even happened.. we should probably raise an issue about it.

openfecli/commands/gather.py
```diff
@@ -77,7 +136,9 @@ def get_type(res):
         return 'solvent'
 
 
-def legacy_get_type(res_fn):
+def legacy_get_type(res_fn:os.PathLike|str)->Literal['vacuum','solvent','complex']:
+    """TODO: when to deprecate this?"""
```
IAlibay, line 140: I think once get_type gets updated to not rely on the key names then we can drop this.

PR #1043, opened 2024-12-09
[WIP] Restraints API

IAlibay:
Checklist
* [ ] Added a ``news`` entry


openfe/protocols/openmm_utils/restraints/openmm/omm_restraints.py
```diff
@@ -0,0 +1,294 @@
...
+            errmsg = (
+                "host_atoms and guest_atoms must only include a single index "
+                f"each, got {len(host_atoms)} and "
+                f"{len(guest_atoms)} respectively."
+            )
+            raise ValueError(errmsg)
+        super()._verify_geometry(geometry)
+
+
+class BaseRadialllySymmetricRestraintForce(BaseHostGuestRestraints):
```
IAlibay, line 128:
```suggestion
class BaseRadiallySymmetricRestraintForce(BaseHostGuestRestraints):
```

openfe/protocols/openmm_utils/restraints/openmm/omm_restraints.py
```diff
@@ -0,0 +1,294 @@
...
+            thermodynamic_state, volume="system"
+        )
+        dg = corr * thermodynamic_state.kT
+        return from_openmm(dg).to('kilojoule_per_mole')
+
+    def _get_force(self, geometry: DistanceRestraintGeometry):
+        raise NotImplementedError("only implemented in child classes")
+
+
+class HarmonicBondRestraint(BaseRadialllySymmetricRestraintForce, SingleBondMixin):
```
IAlibay, line 165:
```suggestion
class HarmonicBondRestraint(BaseRadiallySymmetricRestraintForce, SingleBondMixin):
```

openfe/protocols/openmm_utils/restraints/openmm/omm_restraints.py
```diff
@@ -0,0 +1,294 @@
...
+        spring_constant = to_openmm(self.settings.spring_constant).value_in_unit_system(omm_unit.md_unit_system)
+        return HarmonicRestraintBondForce(
+            spring_constant=spring_constant,
+            restrained_atom_index1=geometry.host_atoms[0],
+            restrained_atom_index2=geometry.guest_atoms[0],
+            controlling_parameter_name=self.controlling_parameter_name,
+        )
+
+
+class FlatBottomBondRestraint(BaseRadialllySymmetricRestraintForce, SingleBondMixin):
```
IAlibay, line 176:
```suggestion
class FlatBottomBondRestraint(BaseRadiallySymmetricRestraintForce, SingleBondMixin):
```

openfe/protocols/openmm_utils/restraints/openmm/omm_restraints.py
```diff
@@ -0,0 +1,294 @@
...
+        return FlatBottomRestraintBondForce(
+            spring_constant=spring_constant,
+            well_radius=well_radius,
+            restrained_atom_index1=geometry.host_atoms[0],
+            restrained_atom_index2=geometry.guest_atoms[0],
+            controlling_parameter_name=self.controlling_parameter_name,
+        )
+
+
+class CentroidHarmonicRestraint(BaseRadialllySymmetricRestraintForce):
```
IAlibay, line 189:
```suggestion
class CentroidHarmonicRestraint(BaseRadiallySymmetricRestraintForce):
```

openfe/protocols/openmm_utils/restraints/openmm/omm_restraints.py
```diff
@@ -0,0 +1,294 @@
...
+        spring_constant = to_openmm(self.settings.spring_constant).value_in_unit_system(omm_unit.md_unit_system)
+        return HarmonicRestraintForce(
+            spring_constant=spring_constant,
+            restrained_atom_index1=geometry.host_atoms,
+            restrained_atom_index2=geometry.guest_atoms,
+            controlling_parameter_name=self.controlling_parameter_name,
+        )
+
+
+class CentroidFlatBottomRestraint(BaseRadialllySymmetricRestraintForce):
```
IAlibay, line 200:
```suggestion
class CentroidFlatBottomRestraint(BaseRadiallySymmetricRestraintForce):
```

IAlibay (commented):
Some typos

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+    # a list of angles in the branch that are sorted by how close the atoms
+    # are from the central atom
+    for at2 in atom_pool:
+        if at2 in at1_neighbors:
+            at2_neighbors = [
+                at.GetIdx() for at in rdmol.GetAtomWithIdx(at2).GetNeighbors()
+            ]
+            for at3 in atom_pool:
+                if at3 != atom_idx and at3 in at2_neighbors:
+                    angles.append((atom_idx, at2, at3))
```
IAlibay, line 152: From today's discussion: we need to avoid cases where we can cross rings with a non-aromatic bond. Most likely easiest way to do this is to have a special case here that can optionally filter by ring (i.e. only pick angles where you have atoms indices that exist solely within one ring).

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+            at.ix for at in host_ag.atoms[host_restraint_atoms_idxs]
+        ]
+
+        # Set the equilibrium values as those of the final frame
+        u.trajectory[-1]
+        atomgroup = u.atoms[host_anchor + guest_anchor]
+        bond, ang1, ang2, dih1, dih2, dih3 = _get_restraint_distances(
+            atomgroup
+        )
+
```
IAlibay, line 789: TODO: add in checks here so we can warn if it's a bad pick. May push this to a follow-up issue.

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+      Array of host atom indexes
+    """
+    u = mda.Universe(
+        topology,
+        trajectory,
+        format=_get_mda_coord_format(trajectory),
+        topology_format=_get_mda_topology_format(topology),
+    )
+
+    host_ag1 = u.atoms[host_idxs]
```
IAlibay, line 332: Maybe find better words for this (or maybe actually explain what the variable means!)

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+      The force constant for the G1-G0-H0 and G0-H0-H1 angles.
+    temperature : unit.Quantity
+      The target system temperature.
+
+    Returns
+    -------
+    Optional[list[int]]
+      A list of indices for a selected combination of H0, H1, and H2.
+    """
+    h0_eval = EvaluateHostAtoms1(
```
IAlibay, line 591: Maybe add some comments on what is going on here?

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+    def _conclude(self):
+        for i, at in enumerate(self.host_atom_pool):
+            distance_bounds = all(self.results.distances[i] > self.minimum_distance)
+            mean_angle = circmean(self.results.angles[i], high=np.pi, low=0)
+            angle_bounds = check_angle_not_flat(
+                angle=mean_angle * unit.radians,
+                force_constant=self.angle_force_constant,
+                temperature=self.temperature,
+            )
+            angle_variance = check_angular_variance(
```
IAlibay, line 451: Let's try doing this for every frame.

openfe/protocols/restraint_utils/geometry/utils.py
```diff
@@ -0,0 +1,445 @@
...
+      The system temperature in units compatible with Kelvin.
+
+    Returns
+    -------
+    bool
+      False if the angle is less than 10 kT from 0 or pi radians
+
+    Note
+    ----
+    We assume the temperature to be 298.15 Kelvin.
```
IAlibay, line 285: Add acknowledgements to Vytas.

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+            self.results.distances[i][self._frame_index] = distance
+            self.results.angles[i][self._frame_index] = angle
+            self.results.dihedrals[i][self._frame_index] = dihedral
+            self.results.collinear[i][self._frame_index] = collinear
+
+    def _conclude(self):
+        for i, at in enumerate(self.host_atom_pool):
+            distance_bounds = all(self.results.distances[i] > self.minimum_distance)
+            mean_angle = circmean(self.results.angles[i], high=np.pi, low=0)
+            angle_bounds = check_angle_not_flat(
```
IAlibay, line 446: Let's try doing this for every frame.

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+                temperature=self.temperature,
+            )
+            angle_variance = check_angular_variance(
+                self.results.angles[i] * unit.radians,
+                upper_bound=np.pi * unit.radians,
+                lower_bound=0 * unit.radians,
+                width=1.745 * unit.radians,
+            )
+            mean_dihed = circmean(self.results.dihedrals[i], high=np.pi, low=-np.pi)
+            dihed_bounds = check_dihedral_bounds(mean_dihed)
```
IAlibay, line 458: Do this per frame?

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+                [
+                    distance_bounds,
+                    angle_bounds,
+                    angle_variance,
+                    dihed_bounds,
+                    dihed_variance,
+                    not_collinear,
+                ]
+            ):
+                self.results.valid[i] = True
```
IAlibay, line 476: If you do that, then you need to set things to False too!

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+        minimum_distance,
+        angle_force_constant,
+        temperature,
+    )
+    h0_eval.run()
+
+    for i, valid_h0 in enumerate(h0_eval.results.valid):
+        if valid_h0:
+            g1g2h0_atoms = guest_atoms.atoms[1:] + host_atom_pool.atoms[i]
+            h1_eval = EvaluateHostAtoms1(
```
IAlibay, line 603: TODO: fix the order, this is incorrect.

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+                g1g2h0_atoms,
+                host_atom_pool,
+                minimum_distance,
```
IAlibay, line 606: Let's double check this is indeed 0.5 nm (or higher)

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+                angle_force_constant,
+                temperature,
+            )
+            for j, valid_h1 in enumerate(h1_eval.results.valid):
+                g2h0h1_atoms = g1g2h0_atoms.atoms[1:] + host_atom_pool.atoms[j]
```
IAlibay, line 611: Also fix this.

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+            self.results.dihedrals[i][self._frame_index] = dihedral
+            self.results.collinear[i][self._frame_index] = collinear
+
+    def _conclude(self):
+        for i, at in enumerate(self.host_atom_pool):
+            distance1_bounds = all(self.results.distances1[i] > self.minimum_distance)
+            distance2_bounds = all(self.results.distances2[i] > self.minimum_distance)
+            mean_dihed = circmean(self.results.dihedrals[i], high=np.pi, low=-np.pi)
+            dihed_bounds = check_dihedral_bounds(mean_dihed)
+            dihed_variance = check_angular_variance(
```
IAlibay, line 544: Same here could do per frame.

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+            if all(
+                [
+                    distance1_bounds,
+                    distance2_bounds,
+                    dihed_bounds,
+                    dihed_variance,
+                    not_collinear,
+                ]
+            ):
+                self.results.valid[i] = True
```
IAlibay, line 560: Here again, make sure you fill in the False case.

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+                h2_eval = EvaluateHostAtoms2(
+                    g2h0h1_atoms,
+                    host_atom_pool,
+                    minimum_distance,
+                    angle_force_constant,
+                    temperature,
+                )
+
+                if any(h2_eval.ressults.valid):
+                    d1_avgs = np.array([d.mean() for d in h2_eval.results.distances1])
```
IAlibay, line 621: This needs to be filtered by True!

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,870 @@
...
+        # continue if it's empty, otherwise stop
+        if host_anchor is not None:
+            break
+
+    if host_anchor is None:
+        errmsg = "No suitable host atoms could be found"
+        raise ValueError(errmsg)
+
+    # Set the equilibrium values as those of the final frame
+    u.trajectory[-1]
```
IAlibay, line 855: Raise an issue / todo here to experiment on what frame / values to pick - ideally some kind of mean with the frame closest to the mean values.

openfe/protocols/restraint_utils/openmm/omm_restraints.py
```diff
@@ -0,0 +1,703 @@
...
+      ``lambda_restraints``.
+    lambda_restraints : Optional[float]
+      The strength of the restraint. If defined, must be between 0 and 1.
+
+    Acknowledgement
+    ---------------
+    Partially reproduced from Yank.
+    """
+
+    lambda_restraints = GlobalParameterState.GlobalParameter(
```
hannahbaumann, line 72: Does ``1`` here mean that the restraint is fully turned on? Maybe add a comment here?

hannahbaumann (commented):
Thanks @IAlibay , just a few small comments from a first quick pass. I left out the ``boresch.py`` and the ``utils`` script for now as discussed. Will do another pass after the break!

openfe/protocols/restraint_utils/geometry/flatbottom.py
```diff
@@ -0,0 +1,137 @@
...
+      A topology defining the system.
+    trajectory : Union[str, pathlib.Path]
+      A coordinate trajectory for the system.
+    host_atoms : Optional[list[int]]
+      A list of host atoms indices. Either ``host_atoms`` or
+      ``host_selection`` must be defined.
+    guest_atoms : Optional[list[int]]
+      A list of guest atoms indices. Either ``guest_atoms`` or
+      ``guest_selection`` must be defined.
+    host_selection : Optional[str]
```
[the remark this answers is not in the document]
IAlibay: Yeah I totally agree with you - to be fair even a Boresch restraint doesn't need to define host/guest. ComponentA/ComponentB might actually be better since the user supplied indices relate back to a Component.

That being said, I also agree with you that it's probably not critical for a first pass. I'm going to open an issue and we should be able to quickly refactor this out after we're done with the basics of everything else.

openfe/protocols/restraint_utils/geometry/harmonic.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Get a DistanceRestraintGeometry between the central atoms of
+    two molecules.
+
+    Parameters
+    ----------
+    molA_rdmol : Chem.Mol
+      An RDKit Molecule for the first molecule.
+    molB_rdmol : Chem.Mol
+      An RDKit Molecule for the first molecule.
```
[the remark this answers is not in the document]
IAlibay: Thanks!

openfe/protocols/restraint_utils/geometry/harmonic.py
```diff
@@ -0,0 +1,118 @@
...
+    molA_idxs : list[int]
+      The indices of the first molecule in the system. Note we assume these
+      to be sorted in the same order as the input rdmol.
+    molB_idxs : list[int]
+      The indices of the first molecule in the system. Note we assume these
```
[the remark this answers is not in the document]
IAlibay: Thanks!

P.S. for some reason I can't apply suggested changes on VS code.. apologies for not merging your suggestions directly! (I didn't want to push/pull half way through other edits)

openfe/protocols/restraint_utils/openmm/omm_restraints.py
IAlibay: Took me a while to remember, the answer is yes - assuming the underlying Force does it properly ;)

What I mean here is that technically it's up to the underlying for to define how the controlling parameter is applied, so in all the cases below it's "lambda_restraints * E" but nothing really stops it from being "1/lambda_restraints * E" is someone wanted to cause chaos.

That being said, I'm pretty sure we'd do our due dilligence to not let that happen..

This is all to say "Yes I've added the comment and updated the docstring" :)

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,867 @@
...
+    """
+    Filter atoms based on rmsf & rings, defaulting to heavy atoms if
+    there are not enough.
+
+    Parameters
+    ----------
+    rdmol : Chem.Mol
+      The RDKit Molecule to search through
+    rmsf : npt.NDArray
+      A 1-D array of RMSF values for each atom.
```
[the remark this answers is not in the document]
IAlibay: I ?think? I've addressed this, but please do let me know if I've completely missed what you meant 🙈

openfe/protocols/restraint_utils/geometry/boresch.py
IAlibay: Done

openfe/protocols/restraint_utils/geometry/boresch.py
IAlibay: Done

openfe/protocols/restraint_utils/geometry/boresch.py
IAlibay: Done - defaults to False on fill

openfe/protocols/restraint_utils/geometry/boresch.py
IAlibay: Done

openfe/protocols/restraint_utils/geometry/boresch.py
IAlibay: Follow-up issue

openfe/protocols/restraint_utils/geometry/boresch.py
```diff
@@ -0,0 +1,949 @@
...
+            atomgroup.atoms[idx_set[3]].position,
+            box=atomgroup.dimensions,
+        )
+        dihedrals.append(dihed * unit.radians)
+
+    return bond, angles[0], angles[1], dihedrals[0], dihedrals[1], dihedrals[2]
+
+
+def find_boresch_restraint(
+    topology: Union[str, pathlib.Path, openmm.app.Topology],
```
IAlibay, line 767: Switch this to just take in a Universe? It would make it easier to deal with the single PDB case.

openfe/protocols/restraint_utils/geometry/boresch/guest.py
```diff
@@ -0,0 +1,251 @@
...
+    for ring in get_aromatic_rings(rdmol):
+        max_rmsf = rmsf[list(ring)].max()
+        if max_rmsf < rmsf_cutoff:
+            atom_pool.update(ring)
+
+    # if we don't have enough atoms just get all the heavy atoms
+    if len(atom_pool) < 3:
+        ring_atoms_only = False
+        heavy_atoms = get_heavy_atom_idxs(rdmol)
+        atom_pool = set(heavy_atoms[rmsf[heavy_atoms] < rmsf_cutoff])
```
hannahbaumann, line 166: This is currently failing. Something like this could work:
```suggestion
        atom_pool = set(np.array(heavy_atoms)[np.array(rmsf[heavy_atoms] < rmsf_cutoff)])
```

openfe/protocols/restraint_utils/geometry/utils.py
```diff
@@ -0,0 +1,614 @@
...
+        wmsg = (
+            "No bounds found in input Universe, "
+            "will attempt to guess for DSSP assignment"
+        )
+        warnings.warn(wmsg)
+        protein_ag.guess_bonds()
+
+    structures = []  # container for all contiguous secondary structure units
+    structure_residue_counts = {'H': 0, 'E': 0, '-': 0}
+    for frag in protein_ag.fragments:
```
hannahbaumann, line 578: If the structure has a small structure part that is not connected to the rest of the protein (e.g. a loop that is only partially modeled or a small protein that interacts with the main one like in jnk1) the DSSP calculation will fail and likely non suitable atoms (e.g. atom index 0) are returned.
Could one maybe either add a check here that if the `len(frag.residues) < trim_chain_start + trim_chain_end` then it would skip that fragment? Or to implement something that if the DSSP fails that it would at least not pick atoms right at the start of the protein?

openfe/protocols/restraint_utils/geometry/boresch/guest.py
IAlibay: Thanks, there's something else odd in this code `atom_pool` is `set[tuple[int]]` which is nonsensical..

openfe/protocols/restraint_utils/geometry/utils.py
```diff
@@ -0,0 +1,614 @@
...
+
+        # Tag each residue structure by its resindex
+        dssp_results = [
+            (structure, resid) for structure, resid in
+            zip(dssp.results.dssp[0], chain.residues.resindices)
+        ]
+
+        for _, group_iter in groupby(dssp_results, lambda x: x[0]):
+            group = list(group_iter)
+            if len(group) > min_structure_size:
```
hannahbaumann, line 591: Maybe `>=` ?

hannahbaumann: This filter currently leads to failures for galectin and hiv integrase from the industry benchmarking systems. Setting this to `>=` finds atoms successfully.
Potentially loosen the filters to `min_structure_size = 6` and/or `trim_structure_ends=2`?

openfe/protocols/restraint_utils/geometry/utils.py
```diff
@@ -0,0 +1,614 @@
...
+    rmsf.run()
+    return rmsf.results.rmsf * unit.angstrom
+
+
+def stable_secondary_structure_selection(
+    atomgroup: mda.AtomGroup,
+    trim_chain_start: int = 10,
+    trim_chain_end: int = 10,
+    min_structure_size: int = 8,
+    trim_structure_ends: int = 3,
```
hannahbaumann, line 488: Potentially change these to be less strict? (e.g. 6 and 2?)

openfe/protocols/restraint_utils/geometry/utils.py
```diff
@@ -0,0 +1,614 @@
...
+                    group[trim_structure_ends:-trim_structure_ends]
+                )
+                num_residues = len(group) - (2 * trim_structure_ends)
+                structure_residue_counts[group[0][0]] += num_residues
+
+    # If we have fewer alpha-helix residues than beta-sheet residues
+    # then we allow picking from beta-sheets too.
+    allowed_structures = ['H']
+    if structure_residue_counts['H'] < structure_residue_counts['E']:
+        allowed_structures.append('E')
```
hannahbaumann, line 602: Maybe add a check on `allowed_structures` to check whether those have a `structure_residue_counts` that is not zero.

hannahbaumann (dismissed):

openfe/protocols/restraint_utils/geometry/utils.py
hannahbaumann: For the system `mcs_docking_set/hne` (with the defaults 8 and 3) this finds two residues in a beta sheet, but it later is not able to pick suitable atoms from that selection.

openfe/protocols/restraint_utils/geometry/utils.py
IAlibay: Yeah we should skip on < trim_chain_start + trim_chain_end + min_structure_size.

> Implement something that if the DSSP fails that it would at least not pick atoms at the start of the protein

I think I have an idea for this, it'll mean effectively implementing a backup selection method.

openfe/protocols/restraint_utils/geometry/utils.py
IAlibay: I'm not sure I fully understand the need for this here.

Possibly explaining my current view on this:
1. If 'H' is zero and 'E' is non-zero, then you can pick 'E' residues so that's good.
2. If 'H' is nonzero and 'E' is zero, then this is the intended behaviour of only picking from 'H'.
3. If 'H' is zero and 'E' is zero, then you don't have anything to pick from, so the intended behaviour is you have an empty selection.

The only case I could think of here is if we want to alter the approach and set a minimum number limit to 'H' residues where if there are too few residues we should also add 'E' "just in case".

Alterrnatively we could just pick from 'H' and 'E' - I suspect in mosts cases it'll be sufficiently stable.

openfe/protocols/restraint_utils/geometry/utils.py
IAlibay: Will do! We should also make this user-controllable. It'll make for a giant settings object, but this is the kind of thing you'll need "per system" flexibility.

openfe/protocols/restraint_utils/geometry/utils.py
```diff
@@ -612,3 +615,71 @@ def stable_secondary_structure_selection(
...
+    # First let's copy our Universe so we don't overwrite its current state
+    copy_u = atomgroup.universe.copy()
+
+    # Create an AtomGroup that contains all the protein residues in the
+    # input Universe - we will filter by what matches in the atomgroup later
+    copy_protein_ag = copy_u.select_atoms('protein').atoms
+
+    # We need to split by fragments to account for multiple chains
+    # To do this, we need bonds!
+    if not _atomgroup_has_bonds(copy_protein_ag, 'bonds'):
```
hannahbaumann, line 693: It looks like this function is only expecting one argument, maybe leftover from an older version?

openfe/protocols/restraint_utils/geometry/boresch/guest.py
hannahbaumann: This returns an array of `np.int64` which leads to an error in line 55 (`pos = conformer.GetAtomPosition(idx)`) which is expecting an `int.`
Maybe something like
```suggestion
        atom_pool = set(heavy_atoms[rmsf[heavy_atoms] < rmsf_cutoff].tolist())
```

IAlibay: I'll think of something here, but generally it's better not to do these types of list operations, they get very expensive.

hannahbaumann:
Does OpenMM have pbc wrapping issues with restraints, similar to gromacs? I just tested the code on T4-lysozyme and it selected protein atoms quite far from the ligand which would cause problems in Gromacs (if the distance between atoms is close to half the box size). But I'm not sure if OpenMM deals with this maybe better?

IAlibay:
> Does OpenMM have pbc wrapping issues with restraints, similar to gromacs? I just tested the code on T4-lysozyme and it selected protein atoms quite far from the ligand which would cause problems in Gromacs (if the distance between atoms is close to half the box size). But I'm not sure if OpenMM deals with this maybe better?

I'll check re: OpenMM. Regarding "residues picked far away", this looks to me like a consequence of the algorithm design. If we want to pick from nearby residues, we should tighten up the minimum distance criteria and sort the host atoms by their distance from the ligand.

The latter is what I do elsewhere and would be my recommendation, but I didn't want to mess with the implementation too much.

hannahbaumann:
Yes, sorting by distance could make sense!
I had this filter (an additional max_filter which I'm not sure if it fully makes sense...) that makes sure that the atoms (different protein atoms relative to each other and to the ligand atom l1), are not further apart than a little less than half the box size.
https://github.com/MobleyLab/SeparatedTopologies/blob/e7f712379472ea320cc049a78b19405014d8bbce/boresch_restraints.py#L389-L390

IAlibay:
> I had this filter (an additional max_filter which I'm not sure if it fully makes sense...) that makes sure that the atoms (different protein atoms relative to each other and to the ligand atom l1), are not further apart than a little less than half the box size.

Unfortunately filtering by half box isn't as simple when you have a non cubic cell. This is why we instead implemented a "take the minimum combined distance between P3-P1 and P3-P1". Anyways, sorting by distance will yield similar outcomes since we sequentially iterate through the atom list.

openfe/protocols/restraint_utils/openmm/omm_restraints.py
```diff
@@ -0,0 +1,712 @@
...
+        correction : unit.Quantity
+          The standard state correction free energy in units compatible
+          with kilojoule per mole.
+        """
+        self._verify_geometry(geometry)
+        # Note: this is a throw-away force, so we hard code the
+        # controlling parameter name
+        force = self._get_force(geometry, 'lambda_restraints')
+        corr = force.compute_standard_state_correction(
+            thermodynamic_state, volume="system"
```
hannahbaumann, line 278: I'm getting an error here, I think the argument is `max_volume="system"`

openfe/protocols/restraint_utils/geometry/boresch/geometry.py
```diff
@@ -0,0 +1,298 @@
...
+    guest_anchors = find_guest_atom_candidates(
+        universe=universe,
+        rdmol=guest_rdmol,
+        guest_idxs=guest_idxs,
+        rmsf_cutoff=rmsf_cutoff,
+    )
+
+    if len(guest_anchors) == 0:
+        errmsg = "No suitable ligand atoms found for the restraint."
+        raise ValueError(errmsg)
```
jthorton, line 250: Do we need an error here since `find_guest_atom_candidates` will raise the same error if no suitable atoms are found in the ligand?

openfe/protocols/restraint_utils/geometry/boresch/geometry.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+def find_boresch_restraint(
+    universe: mda.Universe,
+    guest_rdmol: Chem.Mol,
+    guest_idxs: list[int],
+    host_idxs: list[int],
+    guest_restraint_atoms_idxs: Optional[list[int]] = None,
+    host_restraint_atoms_idxs: Optional[list[int]] = None,
+    host_selection: str = "all",
```
jthorton, line 134: Do we want the default to be backbone or something which avoids hydrogens?

openfe/protocols/restraint_utils/geometry/boresch/geometry.py
jthorton: Ahh I found an example that can hit this error when we have an atom pool, but all the atoms are collinear in the pool.

openfe/protocols/restraint_utils/geometry/boresch/geometry.py
```diff
@@ -0,0 +1,298 @@
...
+    dssp_filter: bool = False,
```
hannahbaumann, line 136: Should the default be `True`?

openfe/protocols/restraint_utils/geometry/boresch/geometry.py
```diff
@@ -0,0 +1,298 @@
...
+        ]
+
+        # Set the equilibrium values as those of the final frame
+        universe.trajectory[-1]
+        atomgroup = universe.atoms[host_atoms + guest_atoms]
+        bond, ang1, ang2, dih1, dih2, dih3 = _get_restraint_distances(
+            atomgroup
+        )
+
+        # TODO: add checks to warn if this is a badly picked
```
hannahbaumann, line 212: Maybe raise an issue for this if this will be addressed in a later PR?

openfe/protocols/restraint_utils/geometry/utils.py
```diff
@@ -0,0 +1,798 @@
...
+    -------
+    int
+      Index of central atom in Molecule
+
+    Note
+    ----
+    If there are equal likelihood centers, will return
+    the first entry.
+    """
+    # TODO: switch to a manual conversion to avoid an OpenFF dependency
```
hannahbaumann, line 165: Does this need to happen before merging or later on?

openfe/protocols/restraint_utils/settings.py
```diff
@@ -0,0 +1,177 @@
...
+
+    If ``guest_atoms`` and ``host_atoms`` are defined,
+    these indices will be used instead.
+
+    References
+    ----------
+    [1] Baumann, Hannah M., et al. "Broadening the scope of binding free
+        energy calculations using a Separated Topologies approach." (2023).
+    """
+    K_r: FloatQuantity['kilojoule_per_mole / nm ** 2'] = 83.68 * unit.kilojoule_per_mole / unit.nm ** 2
```
hannahbaumann: Should we change this to the default in the protocols (8368.0)?

openfe/tests/protocols/restraints/test_geometry_boresch.py
```diff
@@ -0,0 +1,337 @@
...
+            guest_rdmol=lig.atoms.convert_to("RDKIT"),
+            guest_idxs=[a.ix for a in ligand_atoms],
+            host_idxs=[1, 2, 3],
+            host_selection="backbone",
+        )
+
+
+def test_boresch_no_guest_atoms_found_collinear(eg5_protein_pdb):
+    protein = mda.Universe(eg5_protein_pdb)
+    # generate ethane with a single conformation
```
hannahbaumann:
```suggestion
    # generate carbondioxide with a single conformation
```

openfe/tests/protocols/restraints/test_geometry_boresch_guest.py
```diff
@@ -0,0 +1,122 @@
...
+    # add high rmsf values for all ring atoms
+    for i in ring_atoms:
+        rmsf[i] = 1 * unit.nanometer
+
+    pool_atoms, rings = _get_guest_atom_pool(
+        rdmol=rd_mol,
+        rmsf=rmsf,
+        rmsf_cutoff=0.1 * unit.nanometer
+    )
+    # make sure only rings were found
```
hannahbaumann:
```suggestion
    # make sure no rings were found
```

openfe/tests/protocols/restraints/test_geometry_boresch_guest.py
```diff
@@ -0,0 +1,122 @@
...
+
+def test_get_guest_atom_pool_no_atoms(eg5_ligands):
+    rd_mol = eg5_ligands[0].to_rdkit()
+    rmsf = np.ones(rd_mol.GetNumAtoms()) * unit.nanometer
+    pool_atoms, rings = _get_guest_atom_pool(
+        rdmol=rd_mol,
+        rmsf=rmsf,
+        rmsf_cutoff=0.1 * unit.nanometer
+    )
+    # make sure only rings were found
```
hannahbaumann:
```suggestion
    # make sure no rings were found
```

openfe/protocols/restraint_utils/settings.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    Boresch-like restraint search parameter.
+    Whether or not to try to do a DSSP filter on the host atoms.
+    """
+    rmsf_cutoff: Optional[FloatQuantity['nanometer']] = 0.1 * unit.nanometer
+    """
+    Boresch-like restraint search parameter.
+    The cutoff value for filtering atoms by their root mean square fluctuation. Atoms with values above this cutoff will be disregarded.
+    """
+    host_min_distance: Optional[FloatQuantity['nanometer']] = 1 * unit.nanometer
```
hannahbaumann:
```suggestion
    host_min_distance: Optional[FloatQuantity['nanometer']] = 1.0 * unit.nanometer
```
For Protocol json roundtrip

openfe/protocols/restraint_utils/settings.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    Boresch-like restraint search parameter.
+    The minimum distance between any host atom and the guest G0 atom. Must be in units compatible with nanometer.
+    """
+    host_max_distance: Optional[FloatQuantity['nanometer']] = 3 * unit.nanometer
```
hannahbaumann:
```suggestion
    host_max_distance: Optional[FloatQuantity['nanometer']] = 3.0 * unit.nanometer
```
For Protocol json roundtrip

openfe/protocols/restraint_utils/geometry/boresch/geometry.py
IAlibay: I'm going to raise an issue where we just remove all the defaults here. Since it's not user facing, it'll be easier to let settings decide what the defaults should be.

IAlibay: See #1415

openfe/protocols/restraint_utils/geometry/boresch/geometry.py
IAlibay: See #1415

openfe/protocols/restraint_utils/geometry/boresch/geometry.py
IAlibay: Issue #1416

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -703,7 +703,7 @@ def run(self, *, dry=False, verbose=True,
         # Extract relevant settings
         protocol_settings: RelativeHybridTopologyProtocolSettings = self._inputs['protocol'].settings
         stateA = self._inputs['stateA']
-        stateB = self._inputs['stateB']
+        stateB = self._inputs['stateB']  # TODO: open an issue about this not being used.
```
IAlibay:
```suggestion
        stateB = self._inputs['stateB']
```

IAlibay: This has been done.

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -0,0 +1,445 @@
...
+    # 2. Search of atoms within the min/max cutoff
+    atom_finder = FindHostAtoms(
+        host_atoms=filtered_host_ag,
+        guest_atoms=universe.atoms[l1_idx],
+        min_search_distance=min_distance,
+        max_search_distance=max_distance,
+    )
+    atom_finder.run()
+
+    if not atom_finder.results.host_idxs.any():
```
IAlibay, line 121: Issue #1418

openfe/protocols/restraint_utils/geometry/utils.py
IAlibay: This isn't important at this point. It's more something to do if we ever move this to a separate repo.

openfe/protocols/restraint_utils/geometry/utils.py
```diff
@@ -0,0 +1,785 @@
...
+        errmsg = "atoms is not a list of index integers"
+        raise ValueError(errmsg)
+
+    result = False
+    for i in range(len(atoms) - 2):
+        v1 = positions[atoms[i + 1], :] - positions[atoms[i], :]
+        v2 = positions[atoms[i + 2], :] - positions[atoms[i + 1], :]
+        if dimensions is not None:
+            v1 = minimize_vectors(v1, box=dimensions)
+            v2 = minimize_vectors(v2, box=dimensions)
```
IAlibay, line 242: Needs tests see #1419

IAlibay:
Dismissing @hannahbaumann's review as most things have either been addressed or will be addressed in a future PR.

PR #1044, opened 2024-12-09
Add gather support for parallel dir structure

atravitz:
Checklist
* [x] Added a ``news`` entry


openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -256,6 +256,15 @@ def __init__(self, **data):
         if any(len(pur_list) > 2 for pur_list in self.data.values()):
             raise NotImplementedError("Can't stitch together results yet")
 
+    @staticmethod
+    def compute_mean_estimate(dGs:list[dict]):
```
IAlibay, line 260:
```suggestion
    def compute_mean_estimate(dGs:list[unit.Quantity]):
```
?

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -267,24 +276,25 @@ def get_estimate(self) -> unit.Quantity:
         """
         # TODO: Check this holds up completely for SAMS.
         dGs = [pus[0].outputs['unit_estimate'] for pus in self.data.values()]
+        return self.compute_mean_estimate(dGs)
+
+    @staticmethod
+    def compute_uncertainty(dGs:list):
```
IAlibay, line 282: This should have the same signature as above.

Also we should possible rename this to "compute_stdev" - just because "uncertainty" is highly undefined.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -256,6 +256,15 @@ def __init__(self, **data):
...
             raise NotImplementedError("Can't stitch together results yet")
 
+    @staticmethod
+    def compute_mean_estimate(dGs:list[dict]):
+        u = dGs[0].u
+        # convert all values to units of the first value, then take average of magnitude
+        # this would avoid a screwy case where each value was in different units
+        vals = [dG.to(u).m for dG in dGs]
+
+        return np.average(vals) * u
```
IAlibay, line 266: Given that the two staticmethods are effectively the same thing except from this call, how about passing through a callable (i.e. the relevant numpy function) and just standardising it to a single method?

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -256,6 +256,15 @@ def __init__(self, **data):
         if any(len(pur_list) > 2 for pur_list in self.data.values()):
             raise NotImplementedError("Can't stitch together results yet")
 
+    @staticmethod
+    def compute_mean_estimate(dGs:list[dict]):
+        u = dGs[0].u
+        # convert all values to units of the first value, then take average of magnitude
+        # this would avoid a screwy case where each value was in different units
```
IAlibay, line 263: Might just be my sensibilities, and I know it's a historical comment, but I'm not super fond of this comment - would be happy to see it reworded.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
IAlibay: I'm mostly suggesting this because it is kind of a temporary solution to a temporary CLI state.

IAlibay (commented):
Couple comments on the protocol side of things, not read the rest yet.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
atravitz: There's a lot I don't like my fix here, but I think it's the most clear in the short-term. If we pass through a callable, that means we would need to specify _which_ callable on the `gather.py` side, and I want to keep that all within the protocol.

IAlibay: Yeah that's fair. I'll say for now that I don't really mind that much.

However I would, order ~ 2-3 months, would love it if we didn't depend on shims in the Protocol itself. It's fine for now, but when we (I'll say "I" because it's not so enthralling work and I'll probably spend my time on it after ABFEs) do the refactor it would be great to not be in a situation where we could go without the staticmethods. At least it's the first thing I'd likely axe in a feflow move refactor.

atravitz: I agree that this can be handled way better with a bit more refactoring, and that we unfortunately need a shorter-term fix than that. Merging for now and opened and issue here.

documenting here to make sure it gets addressed in the medium-term: https://github.com/OpenFreeEnergy/openfe/issues/1056

news/support_gather_parallel.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* ``openfe gather`` now supports replicates that have been submitted in parallel across separate directories.
```
atravitz, line 3: @IAlibay will this be clear to users, or is there some better language I could be using?

jthorton: I think we could make this very clear by having something similar to [this](https://github.com/OpenFreeEnergy/gufe/blob/339143f86ffe234e474ac3313a9fb3ddb90fca98/gufe/chemicalsystem.py#L14) from the industry benchmarking but in the main CLI docs and point to that, there we could add more details on the expected file structure?

hannahbaumann (commented):
Thanks @atravitz , this looks great. Only thing I'm not so sure about yet is the input file path for gather.
I wanted to test this on one of my old result files, however they were in a structure such as this:
`all_results` with following subfolders:
`results_0`, `results_1`, `results_2`, `results_other_setting_0`, `results_other_setting_1`, `results_other_setting_2`.
If I only want to analyze the results from the first three folders, I'd have to move the other folders in a different directory.
This could be fine, but alternatively, gather could also take a list of Paths. I'm not sure yet what is better, but just wanted to raise this use case.

IAlibay:
> gather could also take a list of Paths.

Just wanted to chime in here that having gather take in a list of results JSON and/or directories would be great (the former more than the latter)! That being said I do think this is maybe something to be dealt with elsewhere.

news/support_gather_parallel.rst
atravitz: I agree- opened an issue [here](https://github.com/OpenFreeEnergy/openfe/issues/1063) since this is outside of the scope of this PR.

atravitz:
I agree with both of you @IAlibay @hannahbaumann  - how do you feel about me addressing this in a follow-up PR?

hannahbaumann (approved):
Thanks @atravitz , lgtm!

PR #1046, opened 2024-12-12
Make MST network test deterministic 

### LINKED ISSUE #1037: Fix `test_network_planning` regression tests
Follow-up from #1033 

The following tests are broken:
  - openfe.tests.setup.test_network_planning.test_minimal_spanning_network_regression
  - openfe.tests.setup.test_network_planning.test_minimal_redundant_network_edges

The cause for these failures is that we are using a scorer that scores based on the length of a mapping: https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/tests/setup/test_network_planning.py#L278-L279
This is prone to clashes.

The team has agreed that we should change this scorer to be deterministic, probably a dictionary that gives the edge we want the right score so we can get the right network.

jthorton:
Fixes #1037 by making the edge scores deterministic, this should ensure we always get the same networks regardless of edge ordering.
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/tests/setup/test_network_planning.py
+++ b/openfe/tests/setup/test_network_planning.py
@@ -273,10 +273,21 @@ def test_minimal_spanning_network_mappers(
 def minimal_spanning_network(toluene_vs_others, lomap_old_mapper):
     toluene, others = toluene_vs_others
 
-    mappers = [BadMapper(), lomap_old_mapper]
+    mappers = [lomap_old_mapper]
 
     def scorer(mapping):
-        return len(mapping.componentA_to_componentB)
+        """Scores are designed to give the same mst everytime"""
+        scores = {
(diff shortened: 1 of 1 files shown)


IAlibay (approved):
lgtm!

IAlibay (changes requested):
Actually backing out a litlte bit - isn't this also likely to have clashes because they all have the same score? I.e. it's still sort dependent?

jthorton:
There should only be a single possible MST with these new edge scores regardless of the graph's ordering though where as the current failures are due to multiple possible MST networks and the returned one is selected based on the ordering of the edges. We also seem to get the same network on main and on the last working commit on gufe so hopefully that shows its not order dependent.

PR #1048, opened 2024-12-12
Test cofactor networks generated with the CLI

jthorton:
Extend the CLI network tests with cofactors to explicitly check that they are included in the transformations. 
Checklist
* [ ] Added a ``news`` entry


jthorton:
Python 3.11/12 tests are failing as the name of the cofactor component is not what we expect, see [here](https://github.com/OpenFreeEnergy/openfe/actions/runs/12300258258/job/34327983233?pr=1048#step:10:2448). 
> AssertionError: assert 'cofactor1' in {'RFEComponentLabels.COFACTOR1': SmallMoleculeComponent(name=3L9H), 'ligand': SmallMoleculeComponent(name=lig_CHEMBL1078691), 'protein': ProteinComponent(name=), 'solvent': SolventComponent(name=O, Na+, Cl-)}

openfe/setup/chemicalsystem_generator/easy_chemicalsystem_generator.py
```diff
@@ -112,7 +112,7 @@ def __call__(
                 RFEComponentLabels.PROTEIN: self.protein,
             }
             for i, c in enumerate(self.cofactors):
-                components.update({f'{RFEComponentLabels.COFACTOR}{i+1}': c})
+                components.update({f'{RFEComponentLabels.COFACTOR.value}{i+1}': c})
```
atravitz, line 115: what is the reason for using the value here instead of using the `RFEComponentLabels` object itself?

jthorton: For some reason using it like this gives an unexpected label, the cofactor is called `RFEComponentLabels.COFACTOR1` whereas the ligand and protein components are labelled `ligand` and `protein` as expected. See the [error](https://github.com/OpenFreeEnergy/openfe/actions/runs/12300258258/job/34327983233?pr=1048#step:10:2448) for how this looks, I guess its not super important to change as I don't think these labels are used for anything but it doesn't look nice and its not easy to pull out the cofactor with the inconsistent naming.

atravitz: within `abstract_chemicalsystem_generator.py`, if you make the class inherit from `StrEnum` instead of both `str` and `Enum` (i.e.`RFEComponentLabels(StrEnum)`), it'll solve this! 

It looks like `StrEnum` has some additional handling for string representation, but I don't quite understand the details.

atravitz (changes requested):
Just one change to make the enum handling cleaner.

openfe/setup/chemicalsystem_generator/easy_chemicalsystem_generator.py
jthorton: Ahhh thanks, though I think that the component keys should all be strings going by the [type hint](https://github.com/OpenFreeEnergy/gufe/blob/339143f86ffe234e474ac3313a9fb3ddb90fca98/gufe/chemicalsystem.py#L14) in gufe so maybe they should all be using `.value`?

atravitz: this is a nice explainer: https://tsak.dev/posts/python-enum/

tldr: 3.11 introduces some enum behavior changes, but using `.value` gives consistent behavior across versions, so please go ahead and use `.value` in all instances.

jthorton: Thanks thats super helpful!

atravitz (approved):

Issue #852, opened 2024-05-14
raise a warning  if a user supplies an unsupported key in a settings yaml


mikemhenry:
see https://github.com/OpenFreeEnergy/openfe/issues/851#issuecomment-2109129135

PR #1051, opened 2024-12-12
add warning for unsupported settings yaml parameters

atravitz:
addresses !https://github.com/OpenFreeEnergy/openfe/issues/852


mikemhenry (approved):
LGTM -- eventually we should use a schema to verify yaml but this is a great fix for now

PR #1052, opened 2024-12-12
Simplify settings help message

atravitz:
This is just my opinion so I'm open to a veto, but I think that CLI help messages should be concise where possible.


jthorton:
I agree it should give a brief example and then point to the docs which lists all the choices!

IAlibay:
I've not read the change but I agree to simplifying too!

openfecli/parameters/plan_network_options.py
```diff
@@ -180,21 +180,14 @@ def load_yaml_planner_options(path: Optional[str], context) -> PlanNetworkOption
...
-Finally, a `settings:` key can be given to customise the algorithm,
-with allowable options corresponding to the keyword arguments of the Python API for these algorithms.
-
-For example, this is a valid settings yaml file to specify that
-the Lomap atom mapper should be used forbidding element changes,
-while the generate_minimal_redundant_network function used to plan the network
+Path to a YAML file specifying the atom mapper (`mapper:`) and/or network planning algorithm (`network:`) to use.
+
+Supported atom mapper choices include `LomapAtomMapper` or `KartografAtomMapper`.
+Supported network planning algorithms include `generate_minimal_spanning_tree` or `generate_minimal_redundant_network`.
```
hannahbaumann, line 186: Maybe also include `generate_radial_network` as that one might be quite common?

hannahbaumann (commented):
Lgtm @atravitz , just left that one comment!

openfecli/parameters/plan_network_options.py
atravitz: I agree. We also just need more thorough documentation for `plan_network_options` in general.

hannahbaumann (approved):
Thanks @atravitz , lgtm!

PR #1053, opened 2024-12-13
CLI adaptive settings for charge change transformations

### LINKED ISSUE #1000: Support net charge params on CLI
The default parameters are not applicable for net charges, so we need to consider how to support this.

jthorton:
Fixes #1000 by adding adaptive settings based on if the transformation involves a net charge change, for the default RBFE protocol only. 
Checklist
* [x] Added a ``news`` entry


openfecli/tests/commands/test_plan_rbfe_network.py
```diff
@@ -149,6 +152,43 @@ def test_plan_rbfe_network_cofactors(eg5_files):
...
+    args = [
+        '-p', cdk8_files[0],
+        '-M', cdk8_files[1],
+    ]
+
+    with runner.isolated_filesystem():
+        with pytest.warns(UserWarning, match="Charge changing transformation between ligands lig_40 and lig_41"):
+            result = runner.invoke(plan_rbfe_network, args)
+
+            print(result.output)
```
atravitz, line 179: I don't think we need this print statement here.

openfe/utils/ligand_utils.py
```diff
@@ -0,0 +1,23 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from gufe import LigandAtomMapping
+
+def get_alchemical_charge_difference(mapping: "LigandAtomMapping") -> int:
```
atravitz, line 6:
```suggestion
def get_alchemical_charge_difference(mapping: LigandAtomMapping) -> int:
```

openfe/utils/ligand_utils.py
```diff
@@ -0,0 +1,23 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from gufe import LigandAtomMapping
```
atravitz, line 4: just curious, what is the reason for avoiding importing this during runtime?

openfe/setup/alchemical_network_planner/relative_alchemical_network_planner.py
```diff
@@ -315,6 +315,50 @@ def __init__(
...
+        """
+        transformation_name = self.name + "_" + stateA.name + "_" + stateB.name
+
+        # Todo: Another dirty hack! - START
+        protocol_settings = transformation_protocol.settings.unfrozen_copy()
+        if "vacuum" in transformation_name:
+            protocol_settings.forcefield_settings.nonbonded_method = "nocutoff"
+        elif get_alchemical_charge_difference(ligand_mapping_edge) != 0:
+            from openff.units import unit
+            import warnings
```
atravitz, line 336: I'd prefer these get imported at the top unless there's a specific benefit to importing them here.

openfe/setup/alchemical_network_planner/relative_alchemical_network_planner.py
```diff
@@ -315,6 +315,50 @@ def __init__(
...
+        stateA: ChemicalSystem,
+        stateB: ChemicalSystem,
+        transformation_protocol: Protocol,
+    ) -> Transformation:
+        """
+        Overwrite the default method to handle net charge change transformations with our default protocol.
+        """
+        transformation_name = self.name + "_" + stateA.name + "_" + stateB.name
+
+        # Todo: Another dirty hack! - START
```
atravitz, line 330: what needs to be done to address this? can you open an issue?

atravitz (changes requested):
Just a few questions/comments

openfecli/tests/commands/test_plan_rbfe_network.py
jthorton: Good catch, this is what happens when I copy and paste a test!

openfe/setup/alchemical_network_planner/relative_alchemical_network_planner.py
jthorton: Ah this was in the other transformation function I [copied](https://github.com/OpenFreeEnergy/openfe/blob/d2f47194fb955c15861809d89cd456ad383202a1/openfe/setup/alchemical_network_planner/relative_alchemical_network_planner.py#L213-L216) I'll remove it.

openfe/utils/ligand_utils.py
jthorton: As I only use it for a type hint, I thought it might be best to skip the import. I'm happy to change it though!

openfe/setup/alchemical_network_planner/relative_alchemical_network_planner.py
```diff
@@ -315,6 +317,47 @@ def __init__(
...
+    ) -> Transformation:
+        """
+        Overwrite the default method to handle net charge change transformations with our default protocol.
+        """
+        transformation_name = self.name + "_" + stateA.name + "_" + stateB.name
+
+        protocol_settings = transformation_protocol.settings.unfrozen_copy()
+        if "vacuum" in transformation_name:
+            protocol_settings.forcefield_settings.nonbonded_method = "nocutoff"
+        elif get_alchemical_charge_difference(ligand_mapping_edge) != 0:
```
hannahbaumann, line 335: Maybe close to zero would be better? E.g. 1e-3?

openfecli/tests/commands/test_plan_rbfe_network.py
```diff
@@ -149,6 +152,44 @@ def test_plan_rbfe_network_cofactors(eg5_files):
...
+def cdk8_files():
+    with resources.files("openfe.tests.data") as p:
+        if not (cdk8_dir := p.joinpath("cdk8")).exists():
+            shutil.unpack_archive(cdk8_dir.with_suffix(".zip"), p)
+        pdb_path = str(cdk8_dir.joinpath("cdk8_protein.pdb"))
+        lig_path = str(cdk8_dir.joinpath("cdk8_ligands.sdf"))
+
+        yield pdb_path, lig_path
+
+def test_plan_rbfe_network_charge_changes(cdk8_files):
```
hannahbaumann, line 165: I wonder if it would make sense to have a test with three ligands and two edges, one with charge change and one without to test the "mixed" networks.

openfecli/tests/commands/test_plan_rbfe_network.py
hannahbaumann: I still see the print statement, just putting it here in case you forgot to remove it, if not, please ignore this comment!

hannahbaumann (changes requested):
Thanks @jthorton , lgtm, just the few comments!

openfe/setup/alchemical_network_planner/relative_alchemical_network_planner.py
jthorton: I think this should be fine as the `rdkit.Chem.rdmolops.GetFormalCharge` returns an int the difference between them should also be an int, it also seems to work in the test, I might have missed something though.

openfecli/tests/commands/test_plan_rbfe_network.py
jthorton: Good idea I was hoping the other tests would catch any breaks in the settings but they don't seem to be checking for it.

openfe/setup/alchemical_network_planner/relative_alchemical_network_planner.py
hannahbaumann: Thanks for clarifying, sounds good!

openfecli/tests/commands/test_plan_rbfe_network.py
jthorton: done in [983dce8](https://github.com/OpenFreeEnergy/openfe/pull/1053/commits/983dce809a8489d7aa7ad5a9ee76aab596601331)

hannahbaumann (approved):
Thanks @jthorton , lgtm!

PR #1058, opened 2024-12-17
Make MD output settings optional

hannahbaumann:
Checklist
* [ ] Added a ``news`` entry


jthorton (approved):
LGTM! Would it make sense to make the whole output settings class optional as well rather than having to set all of these options to `None` individually when you want to turn off all output?

IAlibay:
> LGTM! Would it make sense to make the whole output settings class optional as well rather than having to set all of these options to `None` individually when you want to turn off all output?

Do you mean at the ProtocolSettings level?

PR #1077, opened 2025-01-14
Use MBAR bootstrap error 

### LINKED ISSUE #1012: Switch to bootstrapping for MBAR errors.

jthorton:
Fixes #1012 by using the bootstrap error from pymbar3/4. 

Would this be a good time to switch to only supporting pymbar4 so we only have to maintain a single interface for MBAR?

Note:
- the full pymbar4 package brings in JAX
- I found that 1000 iterations of bootstrapping only takes around 1 min for the default protocol (using jax)
- For the extended charge changing protocol this can take up to 15 mins (using jax)
- The variability in the dDGs between test runs was larger which meant I had to relax the relative tolerance on the tests

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/news/pymbar_bootstrap.rst
+++ b/news/pymbar_bootstrap.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* The MBAR bootstrap (1000 iterations) error is used to estimate protocol uncertainty and pymbar3 is no longer supported `PR#1077 <https://github.com/OpenFreeEnergy/openfe/pull/1077>`_
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/pymbar_bootstrap.rst
+++ b/news/pymbar_bootstrap.rst
@@ -4,7 +4,7 @@
 
 **Changed:**
 
-* The MBAR bootstrap (1000 iterations) error is used to estimate protocol uncertainty and pymbar3 is no longer supported `PR#1077 <https://github.com/OpenFreeEnergy/openfe/pull/1077>`_
+* The MBAR bootstrap (1000 iterations) error is used to estimate protocol uncertainty instead of the statistical uncertainty (one standard deviation) and pymbar3 is no longer supported `PR#1077 <https://github.com/OpenFreeEnergy/openfe/pull/1077>`_
 
 **Deprecated:**
 
--- a/openfe/protocols/openmm_utils/multistate_analysis.py
+++ b/openfe/protocols/openmm_utils/multistate_analysis.py
@@ -350,15 +350,8 @@ class MultistateEquilFEAnalysis:
             * ``matrix``: Estimated overlap matrix of observing a sample from
               state i in state j
         """
-        try:
-            # pymbar 3
-            overlap_matrix = self.analyzer.mbar.computeOverlap()
-            # convert matrix to np array
-            overlap_matrix['matrix'] = np.array(overlap_matrix['matrix'])
-        except AttributeError:
-            overlap_matrix = self.analyzer.mbar.compute_overlap()
+        return self.analyzer.mbar.compute_overlap()
 
-        return overlap_matrix
 
     def get_exchanges(self) -> dict[str, npt.NDArray]:
         """
--- a/openfe/protocols/openmm_utils/multistate_analysis.py
+++ b/openfe/protocols/openmm_utils/multistate_analysis.py
@@ -69,7 +69,7 @@ class MultistateEquilFEAnalysis:
     result_units : unit.Quantity
       Units to report results in.
     forward_reverse_samples : int
-      The number of samples to use in the foward and reverse analysis
+      The number of samples to use in the forward and reverse analysis
       of the free energies. Default 10.
     """
     def __init__(self, reporter: multistate.MultiStateReporter,
@@ -199,15 +199,13 @@ class MultistateEquilFEAnalysis:
         Parameters
         ----------
         analyzer : multistate.MultiStateSamplerAnalyzer
-          MultiStateSamplerAnalyzer to extract free eneriges from.
+          MultiStateSamplerAnalyzer to extract free energies from.
         u_ln : npt.NDArray
           A n_states x (n_sampled_states * n_iterations)
           array of energies (in kT).
         N_l : npt.NDArray
           An array containing the total number of samples drawn from each
           state.
-        unit_type : unit.Quantity
-          What units to return the free energies in.
 
         Returns
         -------
@@ -364,7 +362,7 @@ class MultistateEquilFEAnalysis:
           A dictionary containing the following:
             * ``eigenvalues``: The sorted (descending) eigenvalues of the
               lambda state transition matrix
-            * ``matrix``: The transition matrix estimate of a replica switchin
+            * ``matrix``: The transition matrix estimate of a replica switching
               from state i to state j.
         """
         # Get replica mixing statistics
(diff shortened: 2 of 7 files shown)


IAlibay:
> On the other hand, adding JAX is not too noticeable compared to the cudatoolkit?

Yeah - I also suspect we're picking up a ton of dependencies elsewhere.

Long term maybe we should look into an `openfe-base` version that has the very minimal set of dependencies for everything.

I'll let @atravitz weigh in, but generally I'm ok / would very much like it if we pushed for pymbar4 w/ JAX.

openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -220,13 +221,14 @@ def _get_free_energy(
...
-
         try:
             # pymbar 3
-            DF_ij, dDF_ij = mbar.getFreeEnergyDifferences()
+            mbar = MBAR(u_ln, N_l, nbootstraps=1000)
+            DF_ij, dDF_ij = mbar.getFreeEnergyDifferences(compute_uncertainty=True, uncertainty_method="bootstrap")
         except AttributeError:
-            r = mbar.compute_free_energy_differences()
+            # pymbar 4
+            mbar = MBAR(u_ln, N_l, solver_protocol="robust", n_bootstraps=1000, bootstrap_solver_protocol="robust")
```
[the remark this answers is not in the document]
jthorton: Yeah running the bootstrapping on repeat is expensive! One thought on the forward and backward estimates should we be subsampling using g_t calculated for this subset of data? In the industry benchmarking I calculated it 3 ways no subsampling, subsample based on the % of data and subsample using the g_t calculated for the full set of data.

IAlibay:
Completely forgot to ask @jthorton - could you have a look through our docs and see if there's anywhere we can make it clear that this is now the bootstrap error? I know some folks got confused by it all.

jthorton:
Currently blocked by perses=0.10.3 which pins to pymbar3.

IAlibay (approved):
@jthorton I'm going to approve with the expectation that we can merge once tests pass once there's a new release of perses.

jthorton:
We might need to delay the import of pymbar, while testing out the CLI for the partial charge generation I see that it prints a lot of info to the terminal, this gets even worse when using multiprocessing. 
```
openfe charge-molecules -M malt1_ligands.sdf -o charged_ligands.sdf -w 6
Warning on use of the timeseries module: If the inherent timescales of the system are long compared to those being analyzed, this statistical inefficiency may be an underestimate.  The estimate presumes the use of many statistically independent samples.  Tests should be performed to assess whether this condition is satisfied.   Be cautious in the interpretation of the data.

****** PyMBAR will use 64-bit JAX! *******
* JAX is currently set to 32-bit bitsize *
* which is its default.                  *
*                                        *
* PyMBAR requires 64-bit mode and WILL   *
* enable JAX's 64-bit mode when called.  *
*                                        *
* This MAY cause problems with other     *
* Uses of JAX in the same code.          *
******************************************

SMALL MOLECULE PARTIAL CHARGE GENERATOR
_________________________________________

```

atravitz:
> We might need to delay the import of pymbar, while testing out the CLI for the partial charge generation I see that it prints a lot of info to the terminal, this gets even worse when using multiprocessing.
> 
> ```
> openfe charge-molecules -M malt1_ligands.sdf -o charged_ligands.sdf -w 6
> Warning on use of the timeseries module: If the inherent timescales of the system are long compared to those being analyzed, this statistical inefficiency may be an underestimate.  The estimate presumes the use of many statistically independent samples.  Tests should be performed to assess whether this condition is satisfied.   Be cautious in the interpretation of the data.
> 
> ****** PyMBAR will use 64-bit JAX! *******
> * JAX is currently set to 32-bit bitsize *
> * which is its default.                  *
> *                                        *
> * PyMBAR requires 64-bit mode and WILL   *
> * enable JAX's 64-bit mode when called.  *
> *                                        *
> * This MAY cause problems with other     *
> * Uses of JAX in the same code.          *
> ******************************************
> 
> SMALL MOLECULE PARTIAL CHARGE GENERATOR
> _________________________________________
> ```

Are we able to suppress just this warning? Otherwise, delaying the import sounds good to me.

atravitz (approved):

atravitz:
Also, adding JAX as a dependency is fine by me

IAlibay:
> > We might need to delay the import of pymbar, while testing out the CLI for the partial charge generation I see that it prints a lot of info to the terminal, this gets even worse when using multiprocessing.
> > ```
> > openfe charge-molecules -M malt1_ligands.sdf -o charged_ligands.sdf -w 6
> > Warning on use of the timeseries module: If the inherent timescales of the system are long compared to those being analyzed, this statistical inefficiency may be an underestimate.  The estimate presumes the use of many statistically independent samples.  Tests should be performed to assess whether this condition is satisfied.   Be cautious in the interpretation of the data.
> > 
> > ****** PyMBAR will use 64-bit JAX! *******
> > * JAX is currently set to 32-bit bitsize *
> > * which is its default.                  *
> > *                                        *
> > * PyMBAR requires 64-bit mode and WILL   *
> > * enable JAX's 64-bit mode when called.  *
> > *                                        *
> > * This MAY cause problems with other     *
> > * Uses of JAX in the same code.          *
> > ******************************************
> > 
> > SMALL MOLECULE PARTIAL CHARGE GENERATOR
> > _________________________________________
> > ```
> 
> Are we able to suppress just this warning? Otherwise, delaying the import sounds good to me.

Oof yeah @jthorton if you don't do the warning supression here, could you open an issue about doing it later? This is OpenMMTools levels of noisy.

jthorton:
Hiding the import warnings from pymbar is proving to be tricking as importing anything from openfe triggers them before I can catch them! The import of the message filter `from openfe.utils.logging_filter import MsgIncludesStringFilter` triggers it

jthorton:
Adding the logging filter to the main init of openfe seems to have done the trick!

IAlibay:
> Adding the logging filter to the main init of openfe seems to have done the trick!

@atravitz could you weigh in on this? I am in favour of it because the alternative is super super noisy, but I do also see the case of "if we do this, then anyone that imports OpenFE at runtime will never see that warning again".

I don't know if there's a way to be more explicit and just catch child process?

atravitz:
> > Adding the logging filter to the main init of openfe seems to have done the trick!
> 
> @atravitz could you weigh in on this? I am in favour of it because the alternative is super super noisy, but I do also see the case of "if we do this, then anyone that imports OpenFE at runtime will never see that warning again".
> 
> I don't know if there's a way to be more explicit and just catch child process?

I looked into it a bit and haven't come up with a nicer way to handle these warnings yet. Curious if @mikemhenry has thoughts as someone who has worked with pymbar. 

@jthorton is the current behavior that it does it not show the warning at all?

I don't want this to block this PR, so please open a follow-up issue and I'll look more into it.

jthorton:
> @jthorton is the current behavior that it does it not show the warning at all?

Yes this now hides the warning completely, I think we could fix this if we changed to only importing pymbar when its used but this is tied to openmmtools as well which might not possible. We could re-do the warning only when we use pymbar so it appears in the protocol execution logs if its critical that we pass on the warning?

news/pymbar_bootstrap.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* The MBAR bootstrap (1000 iterations) error is used to estimate protocol uncertainty and pymbar3 is no longer supported `PR#1077 <https://github.com/OpenFreeEnergy/openfe/pull/1077>`_
```
IAlibay, line 7: @jthorton could you add an "instead of" here too maybe? (i.e. tell them what we used to do for MBAR estimation).

jthorton:
From the OpenFE meeting today, we will drop the perses atom mapper testing to unblock pymbar4 rollout and will come back to this once we have a new perses release.

news/pymbar_bootstrap.rst
jthorton: Done in [44f1b28](https://github.com/OpenFreeEnergy/openfe/pull/1077/commits/44f1b283fb92385549dc26587c5212ba6bed94ac)

IAlibay (approved):

Issue #489, opened 2023-07-17
Allow Openeye for charge generation


jthorton:
PR #479 removed the use of oechem which causes charge generation to go through ambertools sqm which is much slower than openeye and also leads to inconsistent charges as my bespokefit dihedrals are fit against elf10 charges. Would it be possible not to enforce this but allow the passing of some global toolkit registry which can be set using the [context manager](https://github.com/openforcefield/openff-toolkit/blob/a5a90ad5d3a1f6a753b6f1f2335bf2cf05362b31/openff/toolkit/utils/toolkit_registry.py#L378-L379) in the toolkit?

IAlibay:
Duplicate of #480?

IAlibay:
I will point out that the context manager folks keep pointing to is, as far as I can tell, both an undocumented and untested component of the OpenFF toolkit. We've been burned already with private methods getting killed off silently upstream. Ideally OpenFF needs to make this a true supported method before we can use it here.

jthorton:
Good point about the manager in the toolkit causing issues down the line. Another option might be to extend the version implemented in #479 (https://github.com/OpenFreeEnergy/openfe/blob/304d85a82700ad8aa4ad36322d219d743ecd0d5d/openfe/utils/remove_oechem.py#L10) to do something similar to the toolkit version?

richardjgowers:
For context, the reason we have (for now) disabled OEChem backend is that we were getting an error where some molecules were generating different SMILES from rdkit and oechem so charge generation would fail.

I'm a little wary of directly playing with the GLOBAL_TOOLKIT_REGISTRY [as it looks a little in flux](https://github.com/openforcefield/openff-toolkit/issues/493)

We could probably build you a backdoor to do let you use oechem in a hacky way for now

@jthorton not sure I follow your thoughts on toolkit version?

IAlibay:
I'd prefer it if we didn't have to extend our maintenance burden unnecessarily. If someone can ask OpenFF to make that method official & supported that'd be great.

In any case, in order to offer user supported selections I think we'll need to move things into the protocol unit. If so, I will need to look at things after I'm back.

IAlibay:
Note, rather than being hacky I would ask that we instead officially support users passing charges directly to SmallMoleculeComponents. It's something we need anyways. Once that's done we can easily just allow system generator to pick up those charges.

jthorton:
[...]

PR #1082, opened 2025-01-17
Add lomap network generator

### LINKED ISSUE #1070: add support for Lomap network generation
**Is your feature request related to a problem? Please describe.**
To match the feature set of the industry benchmarking scripts, we need to add support for Lomap networks.

**Describe the solution you'd like**
We may want to make changes to Lomap to make it more consistent with the rest of our network generation.

**Describe alternatives you've considered**

**Additional context**

- [x] add cli support for lomap network generation (we already expose it to the API)
- [x] update lomap network generation kwargs and add deprecation warning (https://github.com/OpenFreeEnergy/Lomap/pull/62)
- [x] add a test that calls lomap network generation

atravitz:
closes #1070 

blocked by https://github.com/OpenFreeEnergy/Lomap/pull/62, need to release a minor version of lomap and pin openfe to that version as a minimum.

Checklist
* [x] Added a ``news`` entry


openfecli/tests/commands/test_plan_rbfe_network.py
```diff
@@ -210,9 +210,9 @@ def test_plan_rbfe_network_charge_changes(cdk8_files):
 def custom_yaml_settings():
     return """\
 network:
-  method: generate_minimal_redundant_network
+  method: generate_lomap_network
```
atravitz, line 213: happy to add this as a separate test, but if we were only using `generate_minimal_redundant_network` here to make sure the custom yaml was working, we might as well use the lomap network generator here.

atravitz: I used a debugger to make sure lomap was actually being called in this test, but is stored anywhere that we can check in the test? Or is this motivation for better logging?

IAlibay: Better logging might work, alternatively you could just could the number of edges being generated - you should end up with more than the N_ligands-1 edges you would normally get with an MST.

IAlibay (approved):
This seems find to me thanks!

More tests / logging would be great, but I don't think we have to do them here (or that they should hold off the release).

atravitz:
Thanks @IAlibay - I just switched from the `eg5` dataset to `cdk8` to avoid openeye issues.

IAlibay:
> Thanks @IAlibay - I just switched from the `eg5` dataset to `cdk8` to avoid openeye issues.

Could you possibly open up an issue about the openeye issues? I'm not sure we have any and it feels like we circle around them a lot recently.

atravitz:
> > Thanks @IAlibay - I just switched from the `eg5` dataset to `cdk8` to avoid openeye issues.
> 
> Could you possibly open up an issue about the openeye issues? I'm not sure we have any and it feels like we circle around them a lot recently.

https://github.com/OpenFreeEnergy/openfe/issues/489 I think this is the relevant issue.

IAlibay:
> > > Thanks @IAlibay - I just switched from the `eg5` dataset to `cdk8` to avoid openeye issues.
> > 
> > Could you possibly open up an issue about the openeye issues? I'm not sure we have any and it feels like we circle around them a lot recently.
> 
> https://github.com/OpenFreeEnergy/openfe/issues/489 I think this is the relevant issue.

This shouldn't be happening by default - unless we're using NAGL for charge generation.

PR #1083, opened 2025-01-20
Add support for variable position/velocity trajectory writing

IAlibay:
Depends on: https://github.com/choderalab/openmmtools/pull/767

Also see: https://github.com/OpenFreeEnergy/openfe_analysis/pull/44

TODO:
* [x] Check this works in real life w/ the openfe_analysis patch
* [x] Check that the settings keywords make sense / match other protocols
* [x] Check that the frequency defaults are reasonable / yield sufficiently small trajectories
* [x] Work out the difference in file size compared to current `main`
* [ ] Better document the frequency writing?
* [x] Add pydantic validators (at least positive value checks?)
* [x] Add tests to check that values works (asserting on the reporter's variables) / the right error gets brought up?
* [x] Add ``news`` entry


IAlibay:
>  Check that the frequency defaults are reasonable / yield sufficiently small trajectories

@hannahbaumann the defaults would now yield 50 frames per replica, down from 5000 frames per replica, which should be a large enough reduction in size, what do you think?

hannahbaumann:
Yes, that sounds like a good size!

IAlibay:
I think everything that needs addressing has been addressed, only thing that's blocking is a release of openmmtools and openfe-analysis, but I'll leave it to you to approve @hannahbaumann !

hannahbaumann (approved):
Thanks @IAlibay , lgtm! We'll just need to change the pin after the openmmtools release.

IAlibay:
I'm getting the sense the openmmtools release won't be for a few weeks, so we might have to wait until the next OpenFE release to have this merged.

PR #1086, opened 2025-01-22
Write tests for position and velocity writing frequency

hannahbaumann:
Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
+++ b/openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
@@ -2115,3 +2115,80 @@ def test_structural_analysis_error(tmpdir):
 
     assert 'structural_analysis_error' in ret
     assert 'structural_analysis' not in ret
+
+
+@pytest.mark.parametrize('positions_write_frequency,velocities_write_frequency',
+                         [[100 * unit.picosecond, None],
+                         [None, None],
+                         [None, 100 * unit.picosecond]])
+def test_dry_run_vacuum_write_frequency(benzene_vacuum_system,
+                                        toluene_vacuum_system,
+                                        benzene_to_toluene_mapping,
+                                        positions_write_frequency,
+                                        velocities_write_frequency,
+                                        tmpdir,
+                                        ):
+
+    vac_settings = openmm_rfe.RelativeHybridTopologyProtocol.default_settings()
+    vac_settings.forcefield_settings.nonbonded_method = 'nocutoff'
+    vac_settings.output_settings.positions_write_frequency = positions_write_frequency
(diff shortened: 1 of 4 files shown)


openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -2115,3 +2115,80 @@ def test_structural_analysis_error(tmpdir):
...
+        else:
+            assert reporter.position_interval == 0
+        if velocities_write_frequency:
+            assert reporter.velocity_interval == velocities_write_frequency.m
+        else:
+            assert reporter.velocity_interval == 0
+
+
+@pytest.mark.parametrize('positions_write_frequency,velocities_write_frequency',
+                         [[100.1 * unit.picosecond, 100 * unit.picosecond]])
```
IAlibay, line 2164: It's unclear to me if the intent here was to test both the error being thrown for positions and velocities not being divisible.

If that's the case, then this probably needs modifying to work.

If it was intended for just positions, then I wouldn't do a parameterize here to avoid the setup overheads.

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
```diff
@@ -2115,3 +2115,80 @@ def test_structural_analysis_error(tmpdir):
 
     assert 'structural_analysis_error' in ret
     assert 'structural_analysis' not in ret
+
+
+@pytest.mark.parametrize('positions_write_frequency,velocities_write_frequency',
+                         [[100 * unit.picosecond, None],
+                         [None, None],
+                         [None, 100 * unit.picosecond]])
+def test_dry_run_vacuum_write_frequency(benzene_vacuum_system,
```
IAlibay, line 2124: Do we want to add a similar test for the AHFE protocol? That way we know it's implemented properly in the AFE base code.

IAlibay (changes requested):
Thanks for doing this @hannahbaumann 

I'lll put it under "request changes" since I have two question we should address before merge.

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
hannahbaumann: Added this!

openfe/tests/protocols/test_openmm_equil_rfe_protocols.py
hannahbaumann: Changed this to also check for the error being thrown for the velocities.

IAlibay (approved):
lgtm thanks!

IAlibay (commented):
O actually @hannahbaumann could you add a news entry please?

(please feel free to merge when you're ready!)

PR #1090, opened 2025-01-23
1089 fix tests for missing gather data

### LINKED ISSUE #1089: Fix tests for missing gather data
`test_gather.py` has some tests marked with xfail, and we need to get these working again for #997

atravitz:
Resolves #1089 

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfecli/tests/commands/test_gather.py
+++ b/openfecli/tests/commands/test_gather.py
@@ -1,4 +1,3 @@
-from ast import Str
 from click.testing import CliRunner
 from importlib import resources
 import tarfile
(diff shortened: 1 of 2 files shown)


openfecli/tests/commands/test_gather.py
```diff
@@ -237,7 +243,7 @@ def rbfe_results():
 
 
 @pytest.mark.download
-@pytest.mark.xfail
+@pytest.mark.xfail  # these results are outdated
```
atravitz, line 246: I would be fine with deleting the pooch test for now since it's outdated, with the expectation that we soon implement all our large data being downloaded with pooch or similar.

openfecli/tests/commands/test_gather.py
```diff
@@ -194,31 +195,36 @@ def test_generate_bad_legs_error_message(include):
     for string in expected:
         assert string in msg
 
+class TestGatherFailedEdges:
```
atravitz, line 198: organizing into a class is a personal preference, lmk if you feel strongly.

atravitz:
note: codecov is red because I added a test utils

openfecli/tests/commands/test_gather.py
IAlibay: Just remove for now imho, no need to keep things around if they don't help.

PR #1095, opened 2025-01-24
[WIP] Output all failed edges

atravitz:
note: this works, but I need to make the error message nicer.

Checklist
* [ ] Added a ``news`` entry


openfecli/commands/gather.py
```diff
@@ -147,42 +147,56 @@ def legacy_get_type(res_fn:os.PathLike|str)->Literal['vacuum','solvent','complex
...
+    msg="\nThe following legs are invalid:\n"
+    for leg_types, ligpair in bad_legs:
+        expected_rbfe_types = {'complex', 'solvent'}
+        expected_rhfe_types = {'solvent', 'vacuum'}
+        maybe_rhfe = bool(leg_types & expected_rhfe_types)
+        maybe_rbfe = bool(leg_types & expected_rbfe_types)
+        if maybe_rhfe and not maybe_rbfe:
+            msg += (
+                f"{ligpair}:\n"
+                "\tAssuming this is an RHFE calculation, "
```
IAlibay, line 173: This looks like it might get really noisy, with lots of `Assuming X this is missing`. Could we just set that as a single header and then just `+=` on a list of edge information?

IAlibay (changes requested):
This looks fine to me, but the error message seems like it might be a bit too unweildy.

Ideally we want something that's easy to parse for edges rather than a lot of text around each printed line.

openfecli/commands/gather.py
atravitz: @jthorton @IAlibay how do you feel about this solution?

Here's what the output looks like:

image

My thinking is that we pass through as much info to the user as we can without trying to be overly clever and auto-detecting rbfe vs  rhfe, since it sounds like verbosity hurts more than helps. 

Happy to change it again in the future depending on user needs.

jthorton: This looks good, it is clear from this message which edges have an issue and it is readable. What about adding something at the end with info on restarting the calculations if they have failed? 

For the guessing of the calculation type if we need that info we could require the `AlchemicalNetwork` which would explicitly tell us, this might be useful in future if we wanted to also report missing legs from the network not just failed edges.

jthorton (approved):
One not blocking suggestion for info on restarting, otherwise this is great!

Issue #1104, opened 2025-01-30
gather - check json files


atravitz:
we should have checking similar to what is implemented in industry benchmarks [fix_networks.py](https://github.com/OpenFreeEnergy/IndustryBenchmarks2024/blob/a29a435a04b447f5ddb3beea31fe8168d9d1c617/industry_benchmarks/utils/fix_networks.py#L35)

Key feature here is printing edges that have json files but that are likely failures. Currently, we only flag missing results jsons.

atravitz:
resolved by https://github.com/OpenFreeEnergy/openfe/pull/1112

PR #1112, opened 2025-01-31
1104 gather check json files

atravitz:
#1104 

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -109,7 +109,7 @@ def load_and_check_result(fpath:os.PathLike|str)->dict:
         click.echo(f"{fpath}: Exception found in 'unit_results', assuming to be a failed simulation.", err=True)
 
     if result['estimate'] is None:
-        click.echo(f"No{fpath}: No 'estimate' found, assuming to be a failed simulation", err=True)
+        click.echo(f"{fpath}: No 'estimate' found, assuming to be a failed simulation", err=True)
 
     if result['uncertainty'] is None:
         click.echo(f"{fpath}: No 'uncertainty' found, assuming to be a failed simulation", err=True)
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -78,6 +78,12 @@ def is_results_json(fpath:os.PathLike|str)->bool:
     """Sanity check that file is a result json before we try to deserialize"""
     return 'estimate' in open(fpath, 'r').read(20)
 
+def load_json(fpath):
+    "For ease of testing, we can't mock json.load directly."
+    import json
+    from gufe.tokenization import JSON_HANDLER
+
+    return json.load(open(fpath, 'r'), cls=JSON_HANDLER.decoder)
 
 def load_and_check_result(fpath:os.PathLike|str)->dict:
     """Load the data from a results JSON into a dict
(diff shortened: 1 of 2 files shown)


openfecli/commands/gather.py
```diff
@@ -96,8 +96,25 @@ def load_results(fpath:os.PathLike|str)->dict:
...
+        click.echo(f"{fpath}: No 'unit_results' found, assuming to be a failed simulation.", err=True)
+
+    if all('exception' in u for u in result['unit_results'].values()):
+        click.echo(f"{fpath}: Exception found in 'unit_results', assuming to be a failed simulation.", err=True)
+
+    if result['estimate'] is None:
+        click.echo(f"No{fpath}: No 'estimate' found, assuming to be a failed simulation", err=True)
+
+    if result['uncertainty'] is None:
+        click.echo(f"{fpath}: No 'uncertainty' found, assuming to be a failed simulation", err=True)
```
jthorton, line 115: If an edge has failed this will print all of these messages for that edge? It would be best to have a single message per edge indicating if it has failed.

openfecli/commands/gather.py
```diff
@@ -96,8 +96,25 @@ def load_results(fpath:os.PathLike|str)->dict:
...
+        return None
 
+    if "unit_results" not in result.keys():
+        click.echo(f"{fpath}: No 'unit_results' found, assuming to be a failed simulation.", err=True)
+
+    if all('exception' in u for u in result['unit_results'].values()):
+        click.echo(f"{fpath}: Exception found in 'unit_results', assuming to be a failed simulation.", err=True)
+
+    if result['estimate'] is None:
+        click.echo(f"No{fpath}: No 'estimate' found, assuming to be a failed simulation", err=True)
```
jthorton, line 112:
```suggestion
        click.echo(f"{fpath}: No 'estimate' found, assuming to be a failed simulation", err=True)
```

mikemhenry:
@atravitz are we bumping this to 1.4?

mikemhenry:
I can't remember the providence of that data set, but don't we want some data that has exceptions so can test handling exceptions in our tooling? Or is this dataset supposed to be clean?

atravitz:
@mikemhenry @jthorton disregard-  this is due to an oversight by me when creating the parallel test dataset. this is working as intended, and I just need to update the test dataset on zenodo!

openfecli/commands/gather.py
atravitz: I agree and that's the behavior now 👍

mikemhenry:
@atravitz can you fix the merge conflict? I have a guess on the right DOI but I don't want to be wrong

openfecli/commands/gather.py
```diff
@@ -78,26 +78,63 @@ def is_results_json(fpath:os.PathLike|str)->bool:
...
+        return None
+    if result['estimate'] is None:
+        click.echo(f"{fpath}: No 'estimate' found, assuming to be a failed simulation.", err=True)
+        return None
+    if result['uncertainty'] is None:
+        click.echo(f"{fpath}: No 'uncertainty' found, assuming to be a failed simulation.", err=True)
+        return None
+    if all('exception' in u for u in result['unit_results'].values()):
+        click.echo(f"{fpath}: Exception found in all 'unit_results', assuming to be a failed simulation.", err=True)
+        return None
```
mikemhenry, line 135: few notes/ideas
1. I think using the [`match`](https://docs.python.org/3/reference/compound_stmts.html#overview) statement for all the `if` checks would be cleaner
2. Do we need to worry about the case of `exception`? I'm not sure if we need a `.lower()` call or if we have an agreement with protocol devs that they ensure `exception` will be in the values of `unit_results` if there is one

atravitz: 2. in this case, `openfe gather` is only to be used for protocols built-in to our CLI, so I think this is fine as-is for now. When we want to make this more general later, I'm thinking we'll want to update this function anyway.

atravitz: 1. i'm open to using a match statement, but I'm not sure how to write that for the first and last checks. I can look into it and if it's cleaner add it in a follow-up

mikemhenry (approved):

PR #1131, opened 2025-02-14
Trying to speed up CI a little bit

IAlibay:
My small laptop can't run tests because they take too long :(


IAlibay:
Time to beat is 24 mins.

IAlibay:
> Time to beat is 24 mins.

Now it's roughly 15 mins, so ~ 40% improvement. Still plenty more that could be done, but for the tests I need to run faster this is a win.

openfe/tests/protocols/test_openmmutils.py
```diff
@@ -227,8 +227,9 @@ def get_settings():
 
 class TestFEAnalysis:
 
-    # Note: class scope _will_ cause this to segfault - the reporter has to close
-    @pytest.fixture(scope='function')
+    # Note: scope on this can sometimes cause segfault, may need to revert to
+    # function scope if it happens.
+    @pytest.fixture(scope='class')
```
atravitz, line 232: this is purely for speed-up reasons? I'm hesitant about increasing test fixture scope as a shortcut if we're at all concerned about them being mutated.

openfe/tests/protocols/test_openmmutils.py
```diff
@@ -244,7 +245,7 @@ def reporter(self):
         finally:
             r.close()
 
-    @pytest.fixture()
+    @pytest.fixture(scope='class')
```
atravitz, line 248: same question as above

openfe/tests/protocols/conftest.py
```diff
@@ -10,17 +10,23 @@
 import pooch
 
 
+@pytest.fixture(scope='session')
+def charged_benzene(benzene_modifications):
+    benzene_offmol = benzene_modifications['benzene'].to_openff()
+    benzene_offmol.assign_partial_charges(partial_charge_method='gasteiger')
```
atravitz, line 16: Is it pre-charging the molecule that does the speed-up here? If so, can we just include the charges hard-coded in the file, so that we  1) don't need to generate charges at all and 2) remove any reproducibility issues stemming from the partial charge method?

atravitz (commented):
a few questions

IAlibay:
let's discuss this at today's meeting

openfe/tests/protocols/test_openmmutils.py
IAlibay: (for the future) we discussed this and agreed that we should remove session scopes on everything.

openfe/tests/protocols/test_openmmutils.py
IAlibay: (for the future) this one really benefits from a session scope, we'll try to leave it here for now.

IAlibay:
Decision from today's call: @atravitz and @mikemhenry will work on improving dispatched AWS runners and we'll use this to test the workflow where we manually dispatch long jobs & work through a PR.

openfe/tests/protocols/conftest.py
IAlibay: > remove any reproducibility issues stemming from the partial charge method?

There are no reproducibility issues with gasteiger charges. They are fully reproducible, all the time.

IAlibay: According to @jthorton it's not if the hybridization changes - so it's onlys table if rdkit doesn't change the graph

PR #1137, opened 2025-02-17
adding user message for --n-protocol-repeats transparency

atravitz:
@mikemhenry let me know if this is helpful

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfecli/commands/plan_rbfe_network.py
+++ b/openfecli/commands/plan_rbfe_network.py
@@ -229,10 +229,10 @@ def plan_rbfe_network(
     if overwrite_charges:
         write("\tOverwriting partial charges")
     write("")
+    write(f"\t{n_protocol_repeats=} ({n_protocol_repeats} simulation repeat(s) per protocol unit)\n")
 
     # DO
     write("Planning RBFE-Campaign:")
-    write(f"--n_protocol_repeats={n_protocol_repeats} simulation repeats per protocol unit.")
     alchemical_network, ligand_network = plan_rbfe_network_main(
         mapper=[mapper_obj],
         mapping_scorer=mapping_scorer,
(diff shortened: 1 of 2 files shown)


mikemhenry:
Yes I like it! I think we should look into how we want logging to work and what gets the privilege of writing to std out, but since network planning tends to be interactive/very human in the loop, I think stuff like this is nice to have!

mikemhenry (approved):

openfecli/commands/plan_rbfe_network.py
```diff
@@ -229,6 +229,7 @@ def plan_rbfe_network(
     if overwrite_charges:
         write("\tOverwriting partial charges")
     write("")
+    write(f"\t{n_protocol_repeats=} ({n_protocol_repeats} simulation repeat(s) per protocol unit)\n")
```
IAlibay, line 232:
```suggestion
    write(f"\t{n_protocol_repeats=} ({n_protocol_repeats} simulation repeat(s) per protocol transformation)\n")
```
The wording here is incorrect, it's multiple repeat units in a transformation, not multiple repeats in a unit.

openfecli/commands/plan_rhfe_network.py
```diff
@@ -185,6 +185,7 @@ def plan_rhfe_network(molecules: List[str], yaml_settings: str, output_dir: str,
     if overwrite_charges:
         write("\tOverwriting partial charges")
     write("")
+    write(f"\t{n_protocol_repeats=} ({n_protocol_repeats} simulation repeat(s) per protocol unit)\n")
```
IAlibay, line 188: As above.

IAlibay (changes requested):
"request changes" for wording update, please discard review after change (i.e. this doesn't need a re-review).

PR #1144, opened 2025-02-19
Test openmm 8.2 again 

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,7 +25,7 @@ concurrency:
 
 defaults:
   run:
-    shell: bash -l {0}
+    shell: bash -leo pipefail {0}
 
 jobs:
   tests:
--- a/.github/workflows/docker-cron.yaml
+++ b/.github/workflows/docker-cron.yaml
@@ -12,7 +12,7 @@ on:
 
 defaults:
   run:
-    shell: bash -l {0}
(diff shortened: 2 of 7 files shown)


mikemhenry:
Testing this using the GPU runner but I think the action needs updating

mikemhenry:
Looks like we are seeing the same error:
```
cls = <class 'mdtraj.core.topology.Bond'>, atom1 = ACE0-C, atom2 = ACE0-CH3
type = None, order = Single

    def __new__(cls, atom1, atom2, type=None, order=None):
        """Construct a new Bond.  You should call add_bond()
        on the Topology instead of calling this directly.
    
        Must use __new__ constructor since this is an immutable class
        """
        bond = super().__new__(cls, atom1, atom2)
        assert isinstance(type, Singleton) or type is None, "Type must be None or a Singleton"
>       assert order is None or 1 <= order <= 3, "Order must be int between 1 to 3 or None"
E       TypeError: '<=' not supported between instances of 'int' and 'Single'
```

So something changed with the way openmm does the bond data structure

mikemhenry:
Going to test installing mdtraj from main

mikemhenry:
Locally still seeing the same error with mdtraj from main, will need to make a small example before reporting the bug

Issue #1132, opened 2025-02-14
Move large test datasets out of git repo


atravitz:
**Is your feature request related to a problem? Please describe.**
We are currently storing an increasing amount of data (even though it's zipped) in this git repo. 

**Describe the solution you'd like**
We should switch to storing data on figshare or zenodo. 

**Additional context**
One complication is that `openfe test` won't be able to be run on hpcs that restrict web access. We should think carefully about how to have reasonable test coverage without needing full test data.

mikemhenry:
> One complication is that openfe test won't be able to be run on hpcs that restrict web access. We should think carefully about how to have reasonable test coverage without needing full test data.

I think pooch would work if you drop the data it needs in the right spot since then it would be a cache hit -- so I think we just need a guide "how to download test data on an air-gapped network" that will go something like "download all these files, and put them in `.pooch/openfe/`"

Then that should work, but we will then need a flag that disables the "do I have internet test" since that will skip the tests and it won't even check the cache. I think making a `OFE_USE_LOCAL_TEST_DATA` envar that a user would export would be best which would then disable the internet check but still run the tests the "need" internet.

Issue #1130, opened 2025-02-14
fix parallel gather test dataset


atravitz:
**Describe the bug**
When I mocked up a parallel test data structure, I only carried over `attempt_0`, which causes our CI to fail in some new test cases. 

**To Reproduce**
testing bug shows up in CI here: https://github.com/OpenFreeEnergy/openfe/pull/1112

PR #1148, opened 2025-02-19
move test data to zenodo

atravitz:
resolves https://github.com/OpenFreeEnergy/openfe/issues/1132

also resolves https://github.com/OpenFreeEnergy/openfe/issues/1130

Checklist
* [x] switch from using local data to fetching from zenodo
* [x] support running tests offline if data already exists in the cache
*  match existing `openfe fetch rbfe-tutorial-results` behavior: this will require a bit more work on the cli fetching side, I'll address in a follow-up PR


mikemhenry:
Lets add an import of `from .conftest import HAS_INTERNET` and then a mark `@pytest.mark.skipif(not HAS_INTERNET, reason="Internet seems to be unavailable")` So that we don't get a test fail if someone runs this on an air gapped network

atravitz:
my motivation for putting pooch in a fixture is that it won't get called for tests that don't require it. e.g. if you want to just run the `test_format_estimate_uncertainty` test, you dont want pooch downloading a test dataset.

atravitz:
> Lets add an import of `from .conftest import HAS_INTERNET` and then a mark `@pytest.mark.skipif(not HAS_INTERNET, reason="Internet seems to be unavailable")` So that we don't get a test fail if someone runs this on an air gapped network

I'd like to support these tests in instances where the data exists in the cache, but `HAS_INTERNET=False`? This would cover most of us doing active development, so that we can 1) do offline development if we already have the data cached 2) test on air-gapped networks by manually uploading data into the cache. 

I can brute-force this, but I'm curious if you have an elegant way of going about this?

mikemhenry:
Let me think about it -- my first thought is that we check if `HAS_INTERNET` or `OFE_USE_DATA_CACHE` (which we make a new envar for) is true, since if either of those are true then we can run the tests that need pooch, but the tests that check if we can fetch tutorial data don't use pooch so we would want to keep the way `HAS_INTERNET` behaves for them.

Can you test this?
1. Run a test that uses pooch
2. Turn off your internet
3. Run test again

And see if that actually works?

atravitz:
> Let me think about it -- my first thought is that we check if `HAS_INTERNET` or `OFE_USE_DATA_CACHE` (which we make a new envar for) is true, since if either of those are true then we can run the tests that need pooch, but the tests that check if we can fetch tutorial data don't use pooch so we would want to keep the way `HAS_INTERNET` behaves for them.
> 
> Can you test this?
> 
>     1. Run a test that uses pooch
> 
>     2. Turn off your internet
> 
>     3. Run test again
> 
> 
> And see if that actually works?

Yes that works as expected!

openfecli/tests/commands/test_gather.py
```diff
@@ -134,30 +135,29 @@ def test_get_column(val, col):
...
+@pytest.fixture
+def rbfe_result_dir()->pathlib.Path:
+    def _rbfe_result_dir(dataset)->str:
+        ZENODO_RBFE_DATA.fetch(f'{dataset}.tar.gz', processor=pooch.Untar())
+        cache_dir = pathlib.Path(pooch.os_cache('openfe'))/f'{dataset}.tar.gz.untar/{dataset}/'
+        return  cache_dir
+
+    return _rbfe_result_dir
+
+@pytest.mark.skipif(not os.path.exists(POOCH_CACHE) and not HAS_INTERNET,reason="Internet seems to be unavailable and test data is not cached locally.")
```
atravitz, line 156: @mikemhenry this works but misses edge cases where the openfe cache exists, but the test data inside has been deleted (I think this would only happen if a user manually messed with it).

mikemhenry: I don't think it really misses the edge case, since if someone did that then the test will fail since the data is missing

openfecli/tests/data/rbfe_results.tar.gz
atravitz, line 1: see here: https://github.com/OpenFreeEnergy/openfe/pull/1121/files#r1964427513

same data, just correctly(?) compressed.

mikemhenry (approved):
LGTM -- I don't think we need a news entry since this is pretty back end stuff and functionally it should all work. The only thing I can think of is if we don't add it to the changelog, somewhere we need to document how to manually load the test data on an air gapped system.

atravitz:
> LGTM -- I don't think we need a news entry since this is pretty back end stuff and functionally it should all work. The only thing I can think of is if we don't add it to the changelog, somewhere we need to document how to manually load the test data on an air gapped system.

https://github.com/OpenFreeEnergy/openfe/pull/1156 working on it here

PR #1154, opened 2025-02-21
remove macos x86_64 from single file installer list

atravitz:


mikemhenry:
Can you also update the action so we don't build it anymore?

mikemhenry (changes requested):
Okay -- Can we add a change log entry saying we are dropping support for MacOSX-x86_64? Something about support will be done on a best effort basis or something but we will not prioritize fixing platform specific bugs.

atravitz:
> Can you also update the action so we don't build it anymore?

this is already the current behavior.

added a news entry!

mikemhenry (approved):

Issue #1155, opened 2025-02-21
add instructions for uploading test data to air gapped system


atravitz:
now that our data lives on zenodo

PR #1156, opened 2025-02-21
starting to add air gapped data upload docs

atravitz:
resolves https://github.com/OpenFreeEnergy/openfe/issues/1155

Checklist
* [ ] Added a ``news`` entry


docs/installation.rst
```diff
@@ -650,6 +650,16 @@ The output of the script will also be printed to standard out as it is executed.
 While no sensitive information is extracted, it is good practice to review the output before sending it or posting it to ensure that nothing needs to be redacted.
 For example, if your python path was ``/data/SECRET_COMPOUND_NAME/python`` then that would show up in ``debug.log``.
 
+Testing on an Air Gapped Network (most HPCs)
+--------------------------------------------
```
mikemhenry, line 654:
```suggestion
Testing on an Air Gapped Network
--------------------------------
```

I would just leave out the bit out HPCs since users might just assume they have this problem and they don't. Also we used air gapped all the time, but will users understand that? Maybe "Testing on systems without internet access" would better explain our intent for this section.

atravitz: I agree that is clearer!

PR #1165, opened 2025-02-26
Set all FloatQuantity default settings to float values

### LINKED ISSUE #1164: Fix default value for `MultiStateOutputSettings.positions_write_frequency`
The enforced type of `positions_write_frequency` is an optional `FloatQuantity`, but the default value provided is an integer. Since the default values of a field are not validated (by default), the resulting object will contain the wrong type in that field. The `GufeKey` of a `GufeTokenizable` containing this incorrect type will not be persevered after round-tripping through an intermediate format even though, functionally, the two objects represent the same thing.

https://github.com/OpenFreeEnergy/openfe/blob/91724bc86680e4f26664c5ac7b1f42eb09cb4644/openfe/protocols/openmm_utils/omm_settings.py#L454

The easiest solution to this problem is updating that single line to read `100.0` instead of `100`.

An issue upstream was raises here: https://github.com/OpenFreeEnergy/gufe/issues/491

IAlibay:
Fixes #1164 

### DIFF
--- a/openfe/protocols/openmm_utils/omm_settings.py
+++ b/openfe/protocols/openmm_utils/omm_settings.py
@@ -338,8 +338,8 @@ class IntegratorSettings(SettingsBaseModel):
     class Config:
         arbitrary_types_allowed = True
 
-    timestep: FloatQuantity['femtosecond'] = 4 * unit.femtosecond
-    """Size of the simulation timestep. Default 4 * unit.femtosecond."""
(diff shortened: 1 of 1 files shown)


ianmkenney (approved):
LGTM

mikemhenry:
brings me back to the python 2.7 days when using `/` and ints you want to be floats was going to be a bad time

PR #1190, opened 2025-03-13
make gufe lowercase in docs

atravitz:
while we're addressing this in the gufe docs overhaul, I thought I'd update in openfe docs as well.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/docs/reference/api/index.rst
+++ b/docs/reference/api/index.rst
@@ -2,7 +2,7 @@
 
 .. note::
    We have reproduced API documentation from the `gufe`_ package here for convenience.
-   Gufe serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
+   `gufe`_ serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
 
 OpenFE API Reference
 ====================
--- a/docs/reference/api/index.rst
+++ b/docs/reference/api/index.rst
@@ -1,8 +1,8 @@
 .. _api:
 
 .. note::
-   We have reproduced API documentation from the `GUFE`_ package here for convenience.
-   The GUFE package serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
+   We have reproduced API documentation from the `gufe`_ package here for convenience.
+   Gufe serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
 
 OpenFE API Reference
 ====================
@@ -20,4 +20,4 @@ OpenFE API Reference
     openmm_md
     openmm_protocol_settings
 
-.. _GUFE: https://gufe.readthedocs.io/en/latest/api.html
+.. _gufe: https://gufe.readthedocs.io/en/latest/api.html
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -6,7 +6,7 @@ command line interface.
 
 .. note::
    We have reproduced API documentation from the `gufe`_ package here for convenience.
-   Gufe serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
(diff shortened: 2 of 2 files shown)


mikemhenry:
Is this settled? Feels like we need a team wide discussion for this, it can be async. @jthorton @hannahbaumann see https://github.com/OpenFreeEnergy/gufe/issues/508

Perhaps without opening the whole can of worms with naming things, I think really the only change in the PR that really changes the "name" is `The GUFE package` -> `Gufe` since the other changes I agree with even if it is an acronym (I like `lowercase` stylistically for software packages). Maybe we could go with ``gufe`` in that spot?

This is a minor preference, I am okay with `Gufe` for now in that spot until we decide what we want to do name wise.

atravitz:
> Is this settled? Feels like we need a team wide discussion for this, it can be async. @jthorton @hannahbaumann see [OpenFreeEnergy/gufe#508](https://github.com/OpenFreeEnergy/gufe/issues/508)

I thought this was settled (see "visual style" [here](https://github.com/OpenFreeEnergy/openfe/discussions/1115#discussion-7915729))

But if it's just for the package name, that' s fine by me.

mikemhenry:
AH--sorry, I thought this was motivated with acronym discussion -- my bad! 

I would argue then that it should be `gufe` even if it is the start of a sentence (that is what the visual style says IMHO).

docs/reference/api/index.rst
```diff
@@ -1,8 +1,8 @@
 .. _api:
 
 .. note::
-   We have reproduced API documentation from the `GUFE`_ package here for convenience.
-   The GUFE package serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
+   We have reproduced API documentation from the `gufe`_ package here for convenience.
+   Gufe serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
```
mikemhenry, line 5:
```suggestion
   `gufe`_ serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
```

docs/reference/index.rst
```diff
@@ -5,13 +5,13 @@ This contains details of the Python API as well as a reference to the
 command line interface.
 
 .. note::
-   We have reproduced API documentation from the `GUFE`_ package here for convenience.
-   The GUFE package serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
+   We have reproduced API documentation from the `gufe`_ package here for convenience.
```
mikemhenry, line 8: Probably excessive but we can use a macro to not have to copy and paste this note in two places

docs/reference/index.rst
```diff
@@ -5,13 +5,13 @@ This contains details of the Python API as well as a reference to the
...
+   Gufe serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
```
mikemhenry, line 9:
```suggestion
   `gufe`_ serves as a foundation layer for openfe, providing abstract base classes and object models, and so might be more useful for developers.
```

atravitz:
> I would argue then that it should be `gufe` even if it is the start of a sentence (that is what the visual style says IMHO).

thanks! the start of sentence thing was ambiguous to me - I like keeping it lowercase.

mikemhenry:
I agree it is ambiguous and not clearly stated in the spec, so I would say there is an element of interpretation to it. We've got 2 votes now for `gufe` to be lower case at the start of a sentence. I'll post a comment on the discussion to see if we can get more consensus on it (I am also fine with merging this in)

docs/reference/index.rst
atravitz: once we're >2 then I'll bother with a macro?

mikemhenry: That sounds like a good metric, see:
https://xkcd.com/1319/
https://xkcd.com/1205/

dotsdl:
I'm partial to **lowercase** or `lowercase` for software packages. Thank you for making this change!

PR #1191, opened 2025-03-14
remove deps in pyproject.toml

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

Same as https://github.com/OpenFreeEnergy/gufe/pull/511

These deps don't really help anyone install or test the package and just get in the way of tooling


mikemhenry:
The docs might break since they do use pyproject.toml to install deps if they are listed, but this could actually speed up the build since locally when I created the env for the docs, I only used the yaml, but RTD uses the yaml + the pyproject.toml

mikemhenry:
looks like that didn't break anything!

PR #1199, opened 2025-03-19
1002: improve missing data error handling

### LINKED ISSUE #1002: improve gather error message when output JSON is missing "data" entry
If job fails and so the resulting JSON's ['protocol_result']['data'] is an empty dict, `gather` then fails with an indexing error, which does not make it clear that the script is failing due to an incomplete run.

We should either have better error handling, or return "unknown" (in the case that a user would want a raw output).

atravitz:
resolves #1002

I also added some cleanup to the other tests in this class.

Checklist
* [x] Added a ``news`` entry


mikemhenry (approved):
LGTM!

jthorton (approved):
LGTM!

PR #1208, opened 2025-03-24
only show --allow-partial suggestion if user hasn't used it

atravitz:
Checklist
* [ ] Added a ``news`` entry


mikemhenry (approved):
LGTM -- one idea that is non-blocking, a test that has and doesn't have the `allow_partial` flag and we assert the error is raised and the warning is raised, depending on the case

atravitz:
> LGTM -- one idea that is non-blocking, a test that has and doesn't have the `allow_partial` flag and we assert the error is raised and the warning is raised, depending on the case

good idea - a check already exists for the error, but I added a `pytest.warns(match)` for good measure

PR #1210, opened 2025-03-25
switch kcal/mol to kilocalorie_per_mole for roundtrip consistency

atravitz:
gufe added new tests to ``GufeTokenizableTestsMixin`` for json and msgpack roundtripping, and it caught that roundtripping `RelativeHybridTopologyProtocol` and `AbsoluteSolvationProtocol` does not reproduce the same gufe key. 

This is because the original object's `early_termination_target_error`'s units are `kilocalorie_per_mole`, but when it's loaded back in, it's in units of `kilocalorie / mole`

We think this is because of pydantic casting, and changing the pydantic class to expect `kilocalorie_per_mole` solves this.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/tests/protocols/test_rfe_tokenization.py
+++ b/openfe/tests/protocols/test_rfe_tokenization.py
@@ -3,6 +3,7 @@
 
 from openfe.protocols import openmm_rfe
 from gufe.tests.test_tokenization import GufeTokenizableTestsMixin
+from openff.units import unit
 import pytest
 
 """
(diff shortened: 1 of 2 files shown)


hannahbaumann (approved):
Thanks @atravitz , lgtm!

ianmkenney (approved):
Under the assumption that this is reasonable for the other reviewers, as they are more familiar with downstream effects, this looks good to me.

It might also be good to check to see what happens if a user provides a quantity in `kcal/mol`, rather than relying on the default, and if this comes up again.

mikemhenry:
> It might also be good to check to see what happens if a user provides a quantity in kcal/mol, rather than relying on the default, and if this comes up again.

That exact point was raised during our meeting so we will add a test to make sure it works

mikemhenry (changes requested):
Just need a test to make sure if a user sets something with `'kcal/mol'` style it works -- it absolutely should since that is how this works BUT we will want that test since if we ever change the way our models work, we could introduce a regression

mikemhenry (approved):
LGTM!

mikemhenry:
Also, no need for a news entry since this really does not affect users at all

openfe/tests/protocols/test_rfe_tokenization.py
```diff
@@ -36,15 +44,30 @@ class TestRelativeHybridTopologyProtocolResult(GufeTokenizableTestsMixin):
     def instance(self):
         pass
 
+class TestRelativeHybridTopologyProtocolOtherUnits(GufeTokenizableTestsMixin):
```
IAlibay, line 47: I was looking at the tests earlier today and stumbled upon this - the name is really confusing. Can we maybe consider switching it out to something else? (especially having it in tokenization just threw me off).

PR #1220, opened 2025-03-31
fix docker build workflow

atravitz:
Now that we're running with `shell: bash -leo pipefail {0}`, the singularity calls to `--version` and `--help` are failing because we were actually calling them incorrectly, but it was just failing silently before.

This should fix it.

Checklist
* [ ] Added a ``news`` entry


atravitz:
workflow running here: https://github.com/OpenFreeEnergy/openfe/actions/runs/14178124827

mikemhenry (approved):
Thanks! No news entry needed, just some internal testing tooling updates

Issue #1211, opened 2025-03-25
cli gather: add alchemical_network.json awareness


atravitz:
**Is your feature request related to a problem? Please describe.**

As discussed in openfe dev meeting, we should add the alchemical network as a cli gather argument, so that we can use it as the source of truth when checking for failed or missing runs. 

The industry benchmarking scripts already do this, and we can achieve this without an api break by:
- make alchemical_network cli arg optional
- if alchemical_network arg is _not_ passed, raise a warning that strongly encourages the user to pass in alchemical_network

in 2.0, we can discuss removing the legacy behavior.

PR #1227, opened 2025-04-09
Pass through failed sims

atravitz:
This PR is work that should be done in preparation of resolving https://github.com/OpenFreeEnergy/openfe/issues/1211.

The purpose of using the AlchemicalNetwork in the gathering stage is to better report on missing legs. 
The desired end state is for output tables (raw, dg, ddg) to include failed or missing simulations. 

Before we check for missing simulations though, we can add support for passing through failed simulation information such that it is reported in these outputs. 

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -366,6 +366,69 @@ def _write_dg_mle(legs:dict, writer:Callable, allow_partial:bool):
         writer.writerow([ligA, DG, unc_DG])
 
 
+def _collect_result_jsons(results: List[os.PathLike | str]) -> List[pathlib.Path]:
+    """Recursively collects all results JSONs from the paths in ``results``,
+    which can include directories and/or filepaths.
+    """
+    import glob
+
+    def collect_jsons(results: List[os.PathLike]):
+        all_jsons = []
+        for p in results:
+            if str(p).endswith("json"):
+                all_jsons.append(p)
+            elif p.is_dir():
+                all_jsons.extend(glob.glob(f"{p}/**/*json", recursive=True))
+
+        return all_jsons
+
+    def is_results_json(fpath: os.PathLike | str) -> bool:
+        """Sanity check that file is a result json before we try to deserialize"""
+        return "estimate" in open(fpath, "r").read(20)
+
+    results = sorted(
+        results
+    )  # ensures reproducible output order regardless of input order
+
+    # 1) find all possible jsons
+    json_fns = collect_jsons(results)
+
+    # 2) filter only result jsons
+    result_fns = filter(is_results_json, json_fns)
+    return result_fns
+
+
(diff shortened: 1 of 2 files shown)


openfecli/commands/gather.py
```diff
@@ -296,36 +353,42 @@ def _write_raw(legs:dict, writer:Callable, allow_partial=True):
...
     g = nx.DiGraph()
     nm_to_idx = {}
     DDGbind_count = 0
     for ligA, ligB, DDGbind, bind_unc, DDGhyd, hyd_unc in DDGs:
-        if DDGbind is None:
+        for lig in (ligA, ligB):
+            if lig not in expected_ligs:
+                expected_ligs.append(lig)
+
+        if DDGbind is None or DDGbind == FAIL_STR:
```
IAlibay, line 391: If you have one or more failed edges, there is a risk that you might end up with a disconnected graph. We should have a is_connected check on `g` below before we call the mle. If it's disconnected then for now I would suggest we just fail saying it is disconnected.

IAlibay (changes requested):
Requested change mostly on the dG writing section.

openfecli/commands/gather.py
atravitz: good call! I added basic handling here, but it's something I'd like to handle more nicely when we do a rewrite.

openfecli/commands/gather.py
```diff
@@ -4,12 +4,13 @@
...
 import os
 import pathlib
+import sys
 from typing import Callable, Literal, List
-import warnings
 
 from openfecli import OFECommandPlugin
 from openfecli.clicktypes import HyphenAwareChoice
 
+FAIL_STR = "NaN" # string used to indicate a failed run in output tables.
```
mikemhenry, line 13: I think we should use 
```suggestion
FAIL_STR = "Err" # string used to indicate a failed run in output tables.
```
since "NaN" (IMHO) implies some sort of numerical issue, but here we are just saying "there is an error here"

IAlibay: That's a good point, I'm not sure which would be better :/ However, if we do go for "Err", given the width of the field, should we just got for the full "Error"?

atravitz: I agree and switched it to `Error`

openfecli/commands/gather.py
```diff
@@ -349,9 +410,13 @@ def _write_dg_mle(legs:dict, writer:Callable, allow_partial:bool):
         g.add_edge(
             idA, idB, calc_DDG=DDGbind, calc_dDDG=bind_unc,
         )
+
     if DDGbind_count > 2:
+        if not nx.is_weakly_connected(g):
+            # TODO: dump the network for debugging?
+            click.secho("The resulting graph is not connected, likely due to failed edges.")
```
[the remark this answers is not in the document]
atravitz: done!

PR #1244, opened 2025-04-23
use pandas

atravitz:
by using pandas, we can streamline our table writing, and make it easier to adjust formatting in a `--report``-agnostic way.

rdkit already pulls in pandas, so this doesn't add a new dependency.

Checklist
* [ ] Added a ``news`` entry


IAlibay (approved):
As per slack - we gave up on lightweight a loooong time ago (we can clean up later if we want).

PR #1245, opened 2025-04-23
print message to stderr when no valid results files are found

atravitz:
currently, if you pass in an empty dir or mistype the your directory name, ``openfe gather`` will generate an empty table and even silently print it to `-o`.

I find this can be confusing and lead to unexpected behavior, so I'm adding a check up front, so that these scenarios are more transparent.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfecli/tests/commands/test_gather.py
+++ b/openfecli/tests/commands/test_gather.py
@@ -116,7 +116,7 @@ class TestResultLoading:
             captured = capsys.readouterr()
             assert result == (None, None)
             assert "Missing ligand names and/or simulation type. Skipping" in captured.err
-    
+
     def test_get_legs_from_result_jsons(self, capsys, sim_result):
         """Test that exceptions are handled correctly at the _get_legs_from_results_json level."""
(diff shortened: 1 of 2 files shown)


openfecli/tests/commands/test_gather.py
```diff
@@ -275,7 +275,7 @@ def test_cmet_failed_edge(self, cmet_result_dir, report, file_regression):
 
     @pytest.mark.parametrize("allow_partial", [True, False])
     def test_cmet_too_few_edges_error(self, cmet_result_dir, allow_partial):
-        results = [str(cmet_result_dir / f"result_{i}_failed_edge") for i in range(3)]
+        results = [str(cmet_result_dir / f"results_{i}_failed_edge") for i in range(3)]
```
atravitz, line 278: this is a fix, not a behavior change from this PR. 
before, it wasn't detecting any edges because of a typo in the file path. 

I think the fact that this PR caught this proves its value!

IAlibay (approved):

PR #1246, opened 2025-04-23
pretty print tables with rich

atravitz:
downstream of https://github.com/OpenFreeEnergy/openfe/pull/1244

I'm open to making this a `--pretty` flag if we want to preserve backwards compatibility. This would be a good idea if users regularly use `>> output.csv` instead of `-o output.csv`. This isn't anything we suggest in the docs, so I'm not sure if it's something people actually do.

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -624,10 +624,9 @@ def gather(results:List[os.PathLike|str],
     # write output
     if isinstance(output, click.utils.LazyFile):
         click.echo(f"writing {report} output to '{output.name}'")
-        df.to_csv(output, sep="\t", lineterminator='\n', index=False)
+
     # TODO: we can use rich to make this output prettier
-    else:
-        click.echo(df.to_string(output, index=False, justify='left', col_space=15))
+    df.to_csv(output, sep="\t", lineterminator='\n', index=False)
 
 PLUGIN = OFECommandPlugin(
     command=gather,
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -554,11 +554,12 @@ def rich_print_to_stdout(df: pd.DataFrame) -> None:
     from rich import box
 
     table = Table(box=box.MINIMAL_HEAVY_HEAD)
+
     for col in df.columns:
         table.add_column(col)
 
-    for value_list in df.values.tolist():
-        row = [str(x) for x in value_list]
+    for row_values in df.values:
+        row = [str(val) for val in row_values]
         table.add_row(*row)
 
     console = Console()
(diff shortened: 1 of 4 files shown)


openfecli/commands/gather.py
```diff
@@ -524,6 +532,26 @@ def _get_legs_from_result_jsons(
...
+    from rich.table import Table
+    from rich import box
+
+    table = Table(box=box.MINIMAL_HEAVY_HEAD)
+    for col in df.columns:
+        table.add_column(col)
+
+    for value_list in df.values.tolist():
+        row = [str(x) for x in value_list]
+        table.add_row(*row)
```
mikemhenry, line 549: This is all super clear, but does rich have any built in functions to work with a pandas df?

IAlibay (approved):

openfecli/commands/gather.py
```diff
@@ -582,31 +610,31 @@ def gather(results:List[os.PathLike|str],
...
+        "ddg": _generate_ddg,
+        "raw": _generate_raw,
     }[report.lower()]
-    writing_func(legs, writer, allow_partial)
+    df = report_func(legs, allow_partial)
+    # write output
+    if isinstance(output, click.utils.LazyFile):
+        df.to_csv(output, sep="\t", lineterminator="\n", index=False)
+
+    # TODO: we can add a --pretty flag if we want this to be optional/preserve backwards compatibility
```
mikemhenry, line 630: I think I rather have a flag that is like `--tsv` that then pints the table out like that and keep rich as the default

mikemhenry (changes requested):
Not sure if you just haven't written tests yet (since it is a draft) but I think that is the only thing missing

openfecli/commands/gather.py
atravitz: surprisingly, no!

openfecli/commands/gather.py
atravitz: i think @mikemhenry just means for stdout, not writing to file. (see what I wrote in the description about users possibly using `>>`)

openfecli/tests/commands/test_gather.py
mikemhenry, line 1: do we have some tests still testing the `-o` flag? (i didn't look to check this)

mikemhenry (changes requested):
looks good! One test question and needs a news item but other than that LGTMM!

mikemhenry:
oh and a test for the rich printing ad well

openfecli/tests/commands/test_gather.py
atravitz: we do now!

atravitz:
> oh and a test for the rich printing as well

this is tricky because I'll need to mock the terminal width, since that affects text wrapping. do you know how to do this off the top of your head? otherwise i'll open an issue to follow up later, and leave it as a smoke test for now.

mikemhenry (approved):
Oh btw since the smoke test is in there, I will give this an approval so it can get merged in now

PR #1254, opened 2025-04-24
immediate print planning cli output

atravitz:
Imports can take 3-6 seconds, which is around the time that you think something is broken on the CLI. I'm moving our imports to right after the header prints to make it clear that the code is being executed.

Longer term,  I need to see if I can speed up our imports.

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/news/cli_responsiveness.rst
+++ b/news/cli_responsiveness.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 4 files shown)


mikemhenry:
yes this is good, I think things are slow b/c of the way we are using plugcli:
https://github.com/dwhswenson/plugcli/blob/main/plugcli/plugin_management.py

I am pretty sure we end up searching through every python package for plugins

mikemhenry (changes requested):
Add a news item, something like "Improve CLI responsiveness"

mikemhenry (approved):

PR #1260, opened 2025-04-29
Import vendored openff.models from gufe

mikemhenry:
Checklist
* [x] Added a ``news`` entry


mikemhenry:
I also just bump the python version to sync with google colab

atravitz (changes requested):
looks good! could you add a news item, if this is technically the commit that makes us numpy 2 compatible?

mikemhenry:
> looks good! could you add a news item, if this is technically the commit that makes us numpy 2 compatible?

Eh, like our code has been numpy 2 compatible since the day it was written so I am always a bit weary about saying that is what this actually does... I think it is worth mentioning in a news bit that we now import a vendored copy of openff.models from gufe, it is on the edge of being kinda internal but I could see someone going down a rabbit hole on how they installed openff-models but things are being weird, so the change log would be a good hint for that

IAlibay:
> > looks good! could you add a news item, if this is technically the commit that makes us numpy 2 compatible?
> 

> I think it is worth mentioning in a news bit that we now import a vendored copy of openff.models from gufe

I strongly disagree, we shouldn't have a news item for this.

I do however agree with Alyssa that we should just put a news item that says "makes toolkit compatible with numpy 2.0". The news items aren't meant to be pure accurate representation of reality but rather a summary of "what am I getting out of this" for users. My compromise here is that you just don't add a PR number and leave it as just an entry in the news.

IAlibay (approved):

atravitz (approved):

PR #1294, opened 2025-05-23
Protocols test migration & MD Protocol simulation tests

### LINKED ISSUE #1295: Evaluate need for `set_openmm_threads_1` in simulation tests
We now restrict the CPU count in vacuum simulations: 

https://github.com/OpenFreeEnergy/openfe/blob/9626bbf4c56010e3c44b4e797e23738bc81c844a/openfe/protocols/openmm_rfe/equil_rfe_methods.py#L972-L977

and

https://github.com/OpenFreeEnergy/openfe/blob/9626bbf4c56010e3c44b4e797e23738bc81c844a/openfe/protocols/openmm_afe/base.py#L196-L202

We _shouldn't_ have to do this in the tests anymore: https://github.com/OpenFreeEnergy/openfe/blob/9626bbf4c56010e3c44b4e797e23738bc81c844a/openfe/tests/protocols/test_openmm_rfe_slow.py#L21-L34

### LINKED ISSUE #1298: MD Protocol protocol unit output includes a PDB for the NVT equilibrated structure even NVT equil was not run
If we set `simulation_settings.equililbration_length_nvt=None`, no NVT simulation will be run.
However, the ProtocolUnit output will still contain a path to a non-existant PDB file for the NVT equilibrated structure, because `output_settings.equil_nvt_structure` is not `None` by default.

### LINKED ISSUE #1299: MD ProtocolUnit outputs point to a checkpoint file that can sometimes not exist.
If the checkpoint interval is greater than the length of the simulation, it is possible that the simulation checkpoint file does not get written.

In that case, the protocolunit output will point to a checkpoint file that doesn't exist.

My proposal is to set this entry to `None` if the file does not exist.

IAlibay:
Fixes #1298 
Fixes #1295 
Fixes #1299

It started as just the AHFE protocol, but I thought it'd be better to just move everything.

TODO:
  - [x] MD Protocol tokenization tests.
  - [x] MD Protocol slow tests.

Checklist
* [x] Added a ``news`` entry - internal API only, no news needed


hannahbaumann (commented):
Thanks @IAlibay , this looks good to me, maybe we can add a test for the number of frames once that mystery is solved... =)

IAlibay:
> Thanks @IAlibay , this looks good to me, maybe we can add a test for the number of frames once that mystery is solved... =)

Yup, I'm adding this in the follow up PR #1302 :)

hannahbaumann (approved):

openfe/tests/protocols/openmm/plain_md/test_simulation_slow.py
```diff
@@ -0,0 +1,144 @@
...
+
+@pytest.mark.integration
+@pytest.mark.parametrize('platform', ['CPU', 'CUDA'])
+def test_vacuum_sim(
+    benzene_vacuum_system,
+    platform,
+    available_platforms,
+    tmpdir
+):
+    if platform not in available_platforms:
```
mikemhenry, line 18: This is nice, eventually we will want to set things up so we can "force" the GPU tests so we can catch issues like the GPU drivers not working BUT for fast tests, I think this is a great idea

mikemhenry (approved):
LGTM! +1 to @atravitz file layout suggestions

openfe/tests/protocols/openmm/plain_md/test_simulation_slow.py
IAlibay: Not me unfortunately, there's a good chance this was you from a previous test :P

IAlibay:
> as far as file organization, I'd prefer to have the tests more closely match the code's file structure. so something like:
> 
> thoughts? (I can make this commit if you agree)

So I started with that file structure, but the only "issue" I found was that when you do abfes & ahfes you want to have them as separate folders but code-wise they exist within the one protocols folder. So that structure starts failing really quickly.

IAlibay:
> > as far as file organization, I'd prefer to have the tests more closely match the code's file structure. so something like:
> > thoughts? (I can make this commit if you agree)
> 
> So I started with that file structure, but the only "issue" I found was that when you do abfes & ahfes you want to have them as separate folders but code-wise they exist within the one protocols folder. So that structure starts failing really quickly.

Just to follow-up here, assuming we don't really mind being loose with that structure (e.g. for abfes), I'm perfectly happy with that structure @atravitz

IAlibay:
> > So I started with that file structure, but the only "issue" I found was that when you do abfes & ahfes you want to have them as separate folders but code-wise they exist within the one protocols folder. So that structure starts failing really quickly.
> 
> I'm not sure I understand the issue - I'm fine with us adding `tests/protocols/openmm_afe/test_abfe_protocol.py` etc. - I still think that it's clearer which tests are supposed to cover which pieces of code. But let me know if I'm misunderstanding!

To avoid confusion, I would liek to have `tests/protocols/openmm_abfe` and `tests/protocols/openmm_ahfe`, even though there's only a single `protocols/openmm_afe`.

atravitz:
ah okay, i'm fine with that! it still keeps things clear imo. i'll make those changes shortly.

openfe/tests/protocols/openmm/plain_md/test_tokenization.py
```diff
@@ -0,0 +1,82 @@
...
+    @pytest.fixture()
+    def instance(self, protocol_result):
+        return protocol_result
+    
+    def test_repr(self, instance):
+        """
+        Overwrites the base `test_repr` call to do a bit more.
+        """
+        assert isinstance(repr(instance), str)
+        assert self.repr in repr(instance)
```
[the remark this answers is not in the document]
IAlibay: I'm going to raise an issue for this. It feels like we should just upstream the check to gufe if we're overriding the same method everywhere.

PR #1302, opened 2025-05-26
Adjust plain MD simulation PDB writing & production step counter

### LINKED ISSUE #1291: MD trajectory (when using the PlainMDProtocol) does not write final frame
Just triaging this, but it looks like our trajectories don't write the final frame (something which I expect coming from GROMACS and AMBER-land).

It might be that we need to manually force writing the final frame to the XTC.

IAlibay:
Fixes #1291 

This PR does two things:
- Make it so that PDBs are written with the same periodicity shift as the XTC file
- Make it so that the step counter resets to 0 during the production phase. This is partly to ensure that the XTC writing is predictable but also a QoL improvement for the simulation log file (since it otherwise starts from a high number and it's unclear to users what's going on).

Checklist
* [x] Added a ``news`` entry


IAlibay:
Note: this PR depends on #1294

hannahbaumann:
Thanks @IAlibay , that looks good to me! Only thing I was wondering is that if the user doesn't want waters in the traj, if it would still be useful to save a full system for the final frame.

atravitz:
as far as file organization, I'd prefer to have the tests more closely match the code's file structure. so something like:

image


thoughts? 
(I can make this commit if you agree)

mikemhenry (approved):

mikemhenry:
> Thanks @IAlibay , that looks good to me! Only thing I was wondering is that if the user doesn't want waters in the traj, if it would still be useful to save a full system somewhere. But maybe if the user specifies 'no water' that's what you get?

So a user sets what they want to store, are you thinking about a case where after a simulation is over, they might change their mind?

hannahbaumann:
> Is this an enhancement request to save a fully solvated PDB of the system too (I thought we were already doing that with preminimized_structure) or did we use to save a fully solvated PDB and we're no longer doing this here?

I meant a fully solvated last frame of the MD, in case people want to use the MD protocol to equilibrate before running FEP, but not want to save big trajectories.

PR #1305, opened 2025-05-27
Include tests in coverage

mikemhenry:
We should include the tests in our code coverage. This will help us see if any tests we made are not running, check if the skips are working how we expect, and help us check for regressions in changes in testing infrastructure. 

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,6 @@ ignore_missing_imports = true
 [tool.coverage.run]
 omit = [
     "openfe/due.py",
-    "*/tests/dev/*py",
     "*/tests/protocols/test_openmm_rfe_slow.py"
 ]
 


mikemhenry:
@IAlibay @atravitz Let me know what you guys think -- if this requires more discussion we can set aside some time to talk it out. See: https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html for some more talking points.

pyproject.toml
```diff
@@ -47,7 +47,6 @@ ignore_missing_imports = true
 [tool.coverage.run]
 omit = [
     "openfe/due.py",
-    "*/tests/dev/*py",
```
IAlibay, line 50: Re: coverage for tests - yes, this is something Swenson enforced early on in the project's history, we should be doing everywhere imho.

In this particular case, I don't think it should be in the coverage count, because this is a purely "devtool" thing, that will never get tested / covered.

i.e. someone saying "hey I'm going to add tests for this file", wouldn't be something that I'd ever consider to be a good use of time.

IAlibay (changes requested):
I'm going to put a soft block on this for further discussion - I'm not convinced we shouldn't just delete that folder if we care about its coverage.

PR #1306, opened 2025-05-27
first pass on a QA script

mikemhenry:
First pass on a script to QA using our `RelativeHybridTopologyProtocol` 

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/gen-lock-file.yaml
+++ b/.github/workflows/gen-lock-file.yaml
@@ -1,0 +1,100 @@
+name: Create OpenFE Conda-Lock File
+
+on:
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash -leo pipefail {0}
+
+jobs:
+  create-conda-lock-file-and-test-linux:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


mikemhenry:
@IAlibay let me know if you have any feedback, I think pulling down the benchmark repo or using a local copy would be a nice improvement.

IAlibay:
🤔 This might be worth discussing, I'm wondering if this is the right place for this to exist.

Ideally we would want to have this in a place where we can also store results. We still need to clean up openfe-benchmarks, but maybe there?

To explain what I'm thinking a bit more, I can see this becoming a thing that grows to be more automated over time, especially if you do things like pulling down the benchmark repo automatically etc... and so maybe it should be somewhere we can test it etc..

mikemhenry:
I agree, I think it needs to live somewhere, my thoughts putting it here is it isn't shipped with the package and isn't in a very public place, I worry about putting it in the benchmark repo right now since it doesn't have a ton of visibility, I do think it should get tested, happy to put it somewhere else!

PR #1170, opened 2025-02-28
test larger cpu runner

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

## Developers certificate of origin
- [ ] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

PR #1346, opened 2025-06-05
clean up CI workflow files

atravitz:
some items in issue: https://github.com/OpenFreeEnergy/openfe/issues/1333?issue=OpenFreeEnergy%7Copenfe%7C1193

Checklist
* [ ] Added a ``news`` entry


.github/workflows/cpu-long-tests.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
           # Set the OFE_SLOW_TESTS to True if running a Cron job
           OFE_SLOW_TESTS: "true"
           DUECREDIT_ENABLE: 'yes'
-          OFE_INTEGRATION_TESTS: TRUE
+          OFE_INTEGRATION_TESTS: FALSE
```
atravitz, line 94: we want CPU runner for long tests, and GPU runner for long tests + integration tests... right? @mikemhenry @IAlibay 

also why do we use `"true" 'yes' FALSE` @mikemhenry

IAlibay: Yeah on the integration tests for GPU only - I think @mikemhenry is adjusting this in his CPU runner PR

mikemhenry: We can still merge this one in first, that PR is here https://github.com/OpenFreeEnergy/openfe/pull/1170

mikemhenry: I would also have to double check the code, but we might be checking for the existence of the envar instead of what they are set to OR allow yes/no/true/false as options

mikemhenry (approved):

.github/workflows/cpu-long-tests.yaml
atravitz: https://github.com/OpenFreeEnergy/openfe/blob/ecfbd65f5636f1ce311c38b679d7999cb2a940ee/openfe/tests/conftest.py#L74-L75 

looks like we check for true/false!

Issue #1193, opened 2025-03-14
optimize CI 


atravitz:
- [x] `conda_cron.yaml` : create environment all in one step, run at 5 UTC. eventually run long tests here? (https://github.com/OpenFreeEnergy/openfe/pull/1432)
- [x] `package-test.yaml`: turn this into`main_cron.yaml`, run long tests here every night at 5 UTC - make our working dir such that we'll catch `MANIFEST.in` issues. (https://github.com/OpenFreeEnergy/openfe/pull/1435)
- [x] `docker_cron.yaml`: make sure this builds a docker image from main and not the latest conda forge. clean up any unnecessary tasks. (fine as-is, I don't believe any work is needed)
- [x] `mypy.yaml`: is this too permissive? check defaults and enforce `ignore_missing_imports` (https://github.com/OpenFreeEnergy/openfe/pull/1433)
- [x] `installer.yaml`: review this and update micromamba version etc. (https://github.com/OpenFreeEnergy/openfe/pull/1434)
- [x] `gpu-runner.yaml` and `cpu-runner.yaml`: rename these to `gpu-integration-tests.yaml` and `cpu-long-test.yaml`
- [x] `clean_cache.yaml`: nothing needed here
- [x] `change.yaml`: rename to `api_break_check.yaml`

for all:
- [x] unpin micromamba (resolved by https://github.com/OpenFreeEnergy/openfe/pull/1348)

atravitz:
@mikemhenry maybe we can also use `if: github.event.pull_request.draft == false` to avoid running expensive tests on draft PRs.

atravitz:
re: mypy defaults, I believe this is the list of defaults used: https://mypy.readthedocs.io/en/stable/error_code_list.html

atravitz:
re: `docker-cron` - I _think_ this is saying that it's running on `main` (as opposed to the latest conda forge) https://github.com/OpenFreeEnergy/openfe/actions/runs/15096818271

the fact that it failed a few times also makes me thing it's running not on the latest conda forge release 😁

atravitz:
closing since all items are addressed, and any issues have been opened for new items.

PR #1348, opened 2025-06-05
pin micromamba version to latest

atravitz:
as discussed in https://github.com/OpenFreeEnergy/openfe/issues/1193

Checklist
* [x] Added a ``news`` entry


atravitz:
> Is this different from the default? If it is the default, do we just want to remove it?

I was 50/50 on this - I know in the past we've had confusion where a newer version of micromamba was pulling in a different package for whatever reason, so I thought it would be helpful to be explicit here. Happy to remove completely if you feel otherwise.

mikemhenry:
we print out the micromamba version 
image
so that will help with debugging, this will also be nice since then we will get updates without having to remember to check for them

mikemhenry (approved):

PR #1362, opened 2025-06-11
Decouple `checkpoint_interval` from `online_analysis_interval`

### LINKED ISSUE #1309: Adapt settings for new checkpoint frequency changes
With @ianmkenney's great work in https://github.com/choderalab/openmmtools/pull/779 we can now write checkpoint files at a different frequency to the `online_analysis_interval`. This means we can write much much smaller checkpoint files.

We should change our settings defaults to match this.

IAlibay:
Fixes #1309 

Work done:
 - Clean up docs
 - Increased checkpoint interval frequency to 1 * unit.nanosecond

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
   - openff-units==0.2.0
   - openmm >=8.0.0,!=8.1.0,<8.2.0
   - openmmforcefields
-  - openmmtools >=0.24.1
+  - openmmtools >=0.25
   - packaging
   - pandas
   - perses>=0.10.3
--- a/news/checkpoint.rst
+++ b/news/checkpoint.rst
@@ -1,0 +1,27 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* The checkpoint interval default frequency has been increased to every
+  nanosecond. `real_time_analysis_interval` no longer needs to be divisible
+  by the checkpoint interval, allowing users of the HybridTopologyProtocol
+  and AbsoluteSolvationProtocol to write checkpoints less frequently and
+  yielding smaller file sizes.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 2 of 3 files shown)


IAlibay:
JSON roundtrips are failing - investigating -- fixed

IAlibay:
Integration tests pass, this should be good to go!

mikemhenry:
Can this be merged in and keep us a 1.x or do we need to start a `next` or `2.x` branch?

IAlibay:
> Can this be merged in and keep us a 1.x or do we need to start a `next` or `2.x` branch?

This is still 1.x - checkpoints don't affect science, especially given that we don't use checkpoints for anything (it would be different if we actually used them).

hannahbaumann (approved):
Thanks @IAlibay , lgtm! The only thing I can think of that this would change, is that a recalculation of MBAR energies using the checkpoint file (as we did for the industry benchmarking) would not be possible any more with these new default settings. Is that correct?

IAlibay:
Duplicating from slack - the reduced potentials are stored in the main storage file. The only thing that will be affected here is that we will store fewer "fulll system" sampler states.

mikemhenry:
> > Can this be merged in and keep us a 1.x or do we need to start a `next` or `2.x` branch?
> 
> This is still 1.x - checkpoints don't affect science, especially given that we don't use checkpoints for anything (it would be different if we actually used them).

Well with this argument, can we make CUDA the default platform then?

IAlibay:
> > > Can this be merged in and keep us a 1.x or do we need to start a `next` or `2.x` branch?
> > 
> > 
> > This is still 1.x - checkpoints don't affect science, especially given that we don't use checkpoints for anything (it would be different if we actually used them).
> 
> Well with this argument, can we make CUDA the default platform then?

Without breaking 1.x? Sure, we will want to bring it to the board for discussion first and be rather loud about it, but as long as folks are well informed, I don't think it's a problem - the science doesn't change and 99% of folks run on CUDA.

mikemhenry:
Cool, I will make an issue about it and we can then see if people are happy with that.

PR #1374, opened 2025-06-13
bump openff-units and pint pins

atravitz:
I _think_ we were pinned to `openff-units==0.2.0` and `pint<0.22` because we were waiting on this: https://github.com/OpenFreeEnergy/gufe/pull/533.

Trying an unpin now to see where we're at.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/analysis/plotting.py
+++ b/openfe/analysis/plotting.py
@@ -182,30 +182,30 @@ def plot_convergence(
     ax.yaxis.set_ticks_position("left")
 
     # Set the overall error bar to the final error for the reverse results
-    overall_error = forward_and_reverse['reverse_dDGs'][-1].m
-    final_value = forward_and_reverse['reverse_DGs'][-1].m
+    overall_error = forward_and_reverse['reverse_dDGs'][-1].m  # type: ignore
+    final_value = forward_and_reverse['reverse_DGs'][-1].m  # type: ignore
     ax.fill_between([0, 1],
                     final_value - overall_error,
                     final_value + overall_error,
                     color='#D2B9D3', zorder=1)
 
     ax.errorbar(
-        forward_and_reverse['fractions'],
+        forward_and_reverse['fractions'],  # type: ignore
         [val.m
          for val in forward_and_reverse['forward_DGs']],
         yerr=[err.m
-              for err in forward_and_reverse['forward_dDGs']],
+              for err in forward_and_reverse['forward_dDGs']],  # type: ignore
         color="#736AFF", lw=3, zorder=2,
         marker="o", mfc="w", mew=2.5,
         mec="#736AFF", ms=8, label='Forward'
     )
 
     ax.errorbar(
-        forward_and_reverse['fractions'],
+        forward_and_reverse['fractions'],  # type: ignore
         [val.m
-         for val in forward_and_reverse['reverse_DGs']],
+         for val in forward_and_reverse['reverse_DGs']],  # type: ignore
         yerr=[err.m
-              for err in forward_and_reverse['reverse_dDGs']],
+              for err in forward_and_reverse['reverse_dDGs']],  # type: ignore
         color="#C11B17", lw=3, zorder=2,
         marker="o", mfc="w", mew=2.5,
         mec="#C11B17", ms=8, label='Reverse',
--- a/openfe/analysis/plotting.py
+++ b/openfe/analysis/plotting.py
@@ -1,11 +1,10 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/openfe
-from itertools import chain
 import matplotlib.pyplot as plt
 from matplotlib.axes import Axes
 import numpy as np
 import numpy.typing as npt
-from openff.units import unit, Quantity
+from openff.units import Quantity
 from typing import Optional, Union
 import warnings
 
--- a/openfe/analysis/plotting.py
+++ b/openfe/analysis/plotting.py
@@ -5,7 +5,7 @@ import matplotlib.pyplot as plt
 from matplotlib.axes import Axes
 import numpy as np
 import numpy.typing as npt
-from openff.units import unit
+from openff.units import unit, Quantity
 from typing import Optional, Union
 import warnings
 
@@ -129,8 +129,8 @@ def plot_lambda_transition_matrix(matrix: npt.NDArray) -> Axes:
 
 
 def plot_convergence(
-    forward_and_reverse: dict[str, Union[npt.NDArray, unit.Quantity]],
-    units: unit.Quantity
+    forward_and_reverse: dict[str, Union[npt.NDArray, Quantity]],
+    units: Quantity
 ) -> Axes:
     """
     Plot a Reverse and Forward convergence analysis of the
@@ -142,7 +142,7 @@ def plot_convergence(
       A dictionary containing the reverse and forward
       values of the free energies sampled along a given fraction
       of the sample size.
-    units : unit.Quantity
+    units : openff.units.Quantity
       The units the free energies are provided in.
 
     Returns
--- a/openfe/analysis/plotting.py
+++ b/openfe/analysis/plotting.py
@@ -192,7 +192,7 @@ def plot_convergence(
     ax.errorbar(
         forward_and_reverse['fractions'],  # type: ignore
         [val.m
-         for val in forward_and_reverse['forward_DGs']],
+         for val in forward_and_reverse['forward_DGs']],  # type: ignore
         yerr=[err.m
               for err in forward_and_reverse['forward_dDGs']],  # type: ignore
         color="#736AFF", lw=3, zorder=2,
--- a/openfe/protocols/openmm_rfe/equil_rfe_methods.py
+++ b/openfe/protocols/openmm_rfe/equil_rfe_methods.py
@@ -258,7 +258,7 @@ class RelativeHybridTopologyProtocolResult(gufe.ProtocolResult):
 
     @staticmethod
     def compute_mean_estimate(dGs:list[Quantity]) -> Quantity:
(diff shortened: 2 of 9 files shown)


IAlibay:
I believe by doing this bump, we will have to fully migrate to pydantic v2+ (iirc interchange v0.4 and above is pydantic 2 only and openff-units 0.3 is does not work with pre-0.4 versions of interchange).

IAlibay:
Regarding the mypy failures, this is what we're seeing here: https://github.com/OpenFreeEnergy/gufe/pull/533

We'll need to import Quantity directly from openff.units for typing.

IAlibay:
For the EitherQuantity failure, the best answer will be to just directly call "from_openmm" instead. kT in ThermodynamicState should be in OpenMM units.

atravitz:
the last nagging mypy errors are:
```
> openfe/protocols/openmm_utils/multistate_analysis.py:370: error: "type[Quantity]" has no attribute "from_list"  [attr-defined]
> openfe/protocols/openmm_rfe/equil_rfe_methods.py:261: error: "Quantity" has no attribute "u"  [attr-defined]
```

I'm going to ignore these to get this PR merged, but im curious if this is expected or fixable @mattwthompson (I see from openff-units that this inherits directly from pint and therefore might be out of your control).

openfe/analysis/plotting.py
```diff
@@ -5,7 +5,7 @@
 from matplotlib.axes import Axes
 import numpy as np
 import numpy.typing as npt
-from openff.units import unit
+from openff.units import unit, Quantity
```
mattwthompson:
```suggestion
from openff.units import Quantity
```
might not need this any more, here and in some other files

mattwthompson, line 8: It shouldn't make a difference in import times in this case, but if shifting from `unit.Quantity` to `Quantity`, the `unit` namespace shouldn't be needed

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -257,20 +257,20 @@ def __init__(self, **data):
             raise NotImplementedError("Can't stitch together results yet")
 
     @staticmethod
-    def compute_mean_estimate(dGs:list[unit.Quantity]):
-        u = dGs[0].u
+    def compute_mean_estimate(dGs:list[Quantity]):
+        u = dGs[0].u  # type: ignore
```
IAlibay:
```suggestion
        u = dGs[0].u 
```
Checked locally, this is uncessary

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -279,15 +279,15 @@ def get_estimate(self) -> unit.Quantity:
         return self.compute_mean_estimate(dGs)
 
     @staticmethod
-    def compute_uncertainty(dGs:list[unit.Quantity]):
-        u = dGs[0].u
+    def compute_uncertainty(dGs:list[Quantity]):
+        u = dGs[0].u  # type: ignore
```
IAlibay:
```suggestion
        u = dGs[0].u
```
As above

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -257,20 +257,20 @@ def __init__(self, **data):
             raise NotImplementedError("Can't stitch together results yet")
 
     @staticmethod
-    def compute_mean_estimate(dGs:list[unit.Quantity]):
-        u = dGs[0].u
+    def compute_mean_estimate(dGs:list[Quantity]):
```
IAlibay:
```suggestion
    def compute_mean_estimate(dGs:list[Quantity]) -> Quantity:
```
Whilst we're here, at least this will help folks know the expected return.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -279,15 +279,15 @@ def get_estimate(self) -> unit.Quantity:
         return self.compute_mean_estimate(dGs)
 
     @staticmethod
-    def compute_uncertainty(dGs:list[unit.Quantity]):
-        u = dGs[0].u
+    def compute_uncertainty(dGs:list[Quantity]):
```
IAlibay:
```suggestion
    def compute_uncertainty(dGs:list[Quantity]) -> Quantity:
```
As above

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -279,15 +279,15 @@ def get_estimate(self) -> unit.Quantity:
...
+        u = dGs[0].u  # type: ignore
         # convert all values to units of the first value, then take average of magnitude
         # this would avoid a screwy case where each value was in different units
         vals = [dG.to(u).m for dG in dGs]
```
IAlibay:
```suggestion
        vals = np.asarray([dG.to(u).m for dG in dGs])
```
Doing this will make mypy happy and it's what happens when you call np.std internally (they use asanyarray to avoid creating an array if there's one already, but this is always a list): https://github.com/numpy/numpy/blob/1fefc5c6b767e86c2642641e7ba7a22ab0cf554b/numpy/_core/_methods.py#L152

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -279,15 +279,15 @@ def get_estimate(self) -> unit.Quantity:
...
 
-        return np.std(vals) * u
+        return np.std(vals) * u  # type: ignore
```
IAlibay:
```suggestion
        return np.std(vals) * u
```
With the above fix this works fine.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -257,20 +257,20 @@ def __init__(self, **data):
             raise NotImplementedError("Can't stitch together results yet")
 
     @staticmethod
-    def compute_mean_estimate(dGs:list[unit.Quantity]):
-        u = dGs[0].u
+    def compute_mean_estimate(dGs:list[Quantity]):
+        u = dGs[0].u  # type: ignore
         # convert all values to units of the first value, then take average of magnitude
         # this would avoid a screwy case where each value was in different units
         vals = [dG.to(u).m for dG in dGs]
```
IAlibay:
```suggestion
        vals = np.asarray([dG.to(u).m for dG in dGs])
```
As with np.std, doing this will make mypy happy and it's what np.average does internally just with `asanyarray`.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -257,20 +257,20 @@ def __init__(self, **data):
...
 
-        return np.average(vals) * u
+        return np.average(vals) * u  # type: ignore
```
IAlibay:
```suggestion
        return np.average(vals) * u
```
With the above fix this works fine.

openfe/protocols/openmm_utils/settings_validation.py
```diff
@@ -89,8 +89,8 @@ def get_simsteps(sim_length: unit.Quantity,
       The number of simulation timesteps.
     """
 
-    sim_time = round(sim_length.to('attosecond').m)
-    ts = round(timestep.to('attosecond').m)
+    sim_time = round(sim_length.to('attosecond').m)  # type: ignore
```
IAlibay, line 92: Oddly enough, mypy does define `__round__`, so we should be able to call round on the attosecond Quantity and then call its magnitude, but something is going wrong here.

Looks like there's some stalled PRs on mypy to deal partially with this.

openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -363,10 +364,10 @@ def get_forward_and_reverse_analysis(
 
         forward_reverse = {
             'fractions': np.array(fractions),
-            'forward_DGs': unit.Quantity.from_list(forward_DGs),
-            'forward_dDGs': unit.Quantity.from_list(forward_dDGs),
-            'reverse_DGs': unit.Quantity.from_list(reverse_DGs),
-            'reverse_dDGs': unit.Quantity.from_list(reverse_dDGs)
+            'forward_DGs': Quantity.from_list(forward_DGs),  # type: ignore
```
IAlibay, line 367: These are all because the pyi from openff-units doesn't define the method. Manually adding it fixes the problem. Raising the issue here: https://github.com/openforcefield/openff-units/issues/127

IAlibay (changes requested):
Couple of things, otherwise lgtm.

openfe/protocols/openmm_rfe/equil_rfe_methods.py
atravitz: strange, this was complaining locally for me (see above) but if the CI is happy, i'm happy!

atravitz:
thank you for the additional fixes, @IAlibay! it looks like the two remaining items are tracked in [openff ](https://github.com/openforcefield/openff-units/issues/127) and mypy, so we're good to merge here?

openfe/protocols/openmm_rfe/equil_rfe_methods.py
```diff
@@ -257,20 +257,20 @@ def __init__(self, **data):
             raise NotImplementedError("Can't stitch together results yet")
 
     @staticmethod
-    def compute_mean_estimate(dGs:list[unit.Quantity]):
-        u = dGs[0].u
+    def compute_mean_estimate(dGs:list[Quantity]) -> Quantity:
+        u = dGs[0].u
```
IAlibay:
```suggestion
        u = dGs[0].u
```

IAlibay (approved):
just one spare space that needs removing

PR #1379, opened 2025-06-18
move to minimum of pydantic v2

atravitz:
we need to move to a minimum of pydantic v2 to support openff-units 0.3.0+ (see [this conversation](https://github.com/OpenFreeEnergy/openfe/pull/1374#issuecomment-2985026159))

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,8 +46,7 @@ jobs:
             openeye: "yes"
           - os: "macos-latest"
             python-version: "3.12"
-            omff-version: ">0.13"
-            openeye: ["no"]
+            openeye: "yes"
 
     env:
       OE_LICENSE: ${{ github.workspace }}/oe_license.txt


.github/workflows/ci.yaml
```diff
@@ -30,31 +30,22 @@ defaults:
...
-            python-version: "3.11"
-            pydantic-version: "<2"
           - os: "ubuntu-latest"
             python-version: "3.11"
-            pydantic-version: ">1"
             openeye: "yes"
           - os: "macos-latest"
             python-version: "3.12"
-            pydantic-version: ">1"
             omff-version: ">0.13"
```
atravitz: what is this?

atravitz:
i'll update the required checks once this is approved.

.github/workflows/ci.yaml
mikemhenry: looks like a leftover from when we supported different openmm forcefield versions

mikemhenry:
So we can still import from the V1 namespace for things, just not install the pydantic < 2 version. That should be fine!

mikemhenry:
This will make it impossible to install openfe and openff-bespokefit into the same env, but that should be okay (at least until bespokefit is updated to pydantic 2)

mikemhenry (approved):

atravitz:
> This will make it impossible to install openfe and openff-bespokefit into the same env, but that should be okay (at least until bespokefit is updated to pydantic 2)

@mattwthompson are you aware of this? not sure who owns that maintenance.

.github/workflows/ci.yaml
atravitz: okay, i've removed that and replaced it with a macos w/ openeye check. do you agree with that?

mattwthompson:
Yeah, Pydantic 2 support (API level) for Bespokefit is somewhere in our backlog. Right now it requires v1 API I think.

Stub issue is https://github.com/openforcefield/openff-bespokefit/issues/381, managed somewhat jointly between Jeff and myself

PR #1382, opened 2025-06-24
Update userguide's ligand network wording for how scores are used.

IAlibay:

### DIFF
--- a/docs/guide/setup/creating_ligand_networks.rst
+++ b/docs/guide/setup/creating_ligand_networks.rst
@@ -8,9 +8,8 @@ For example, a :class:`.LigandNetwork` with drug candidates as nodes can be used
 **openfe** includes an interface to common :any:`Ligand Network Planners`, which are implemented in OpenFE's `konnektor <https://github.com/OpenFreeEnergy/konnektor>`_ package.
 (See `konnektor's documentation <https://konnektor.openfree.energy/en/latest/>`_ for more information on network generators.)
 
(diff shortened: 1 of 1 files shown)


docs/guide/setup/creating_ligand_networks.rst
```diff
@@ -8,9 +8,8 @@ For example, a :class:`.LigandNetwork` with drug candidates as nodes can be used
 **openfe** includes an interface to common :any:`Ligand Network Planners`, which are implemented in OpenFE's `konnektor <https://github.com/OpenFreeEnergy/konnektor>`_ package.
 (See `konnektor's documentation <https://konnektor.openfree.energy/en/latest/>`_ for more information on network generators.)
 
-Because each edge in a :class:`.LigandNetwork` is a :class:`.LigandAtomMapping`, each edge can be assigned a score that indicates the mapping's quality.
-Some network generators use these scores to construct more efficient network topologies.
-
+A :class:`.LigandNetwork` is constructed from :class:`.LigandAtomMapping`, which represent the edges of the network.
```
atravitz: this is not always true, a `LigandNetwork` could just be a bunch of unconnected `SmallMoleculeComponent` nodes. so I think this is less clear than what we currently have.

IAlibay: Fixed now it covers both.

Generally I disagree that this is less clear than the current state, the current state is inherently wrong imho.

atravitz: I like the current state, thank you!

Issue #1296, opened 2025-05-23
pin OpenFE_ExampleNotebooks


atravitz:
We should keep our docs versioned as much as possible. also, see https://github.com/OpenFreeEnergy/ExampleNotebooks/pull/208. Remove the duplicate notebook when doing  this pin.

PR #1385, opened 2025-06-25
updating environment yaml

atravitz:
removing openff-models since we vendored [here](https://github.com/OpenFreeEnergy/openfe/pull/1260), plus updating some other pins.

Also pinning us to a specific example notebooks release to resolve https://github.com/OpenFreeEnergy/openfe/issues/1296

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,7 +17,7 @@ dependencies:
   - openff-nagl-models>=0.1.2
   - openff-toolkit>=0.16.2
   - openff-units==0.3.1  # https://github.com/OpenFreeEnergy/openfe/pull/1374
-  - openmm >=8.1.1, <8.3.0
+  - openmm ~=8.2.0  # 8.2 is qualified, but 8.3 is not yet.
   - openmmforcefields
   - openmmtools >=0.25.0
   - packaging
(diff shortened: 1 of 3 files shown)


environment.yml
```diff
@@ -13,12 +13,11 @@ dependencies:
   - numpy<2.0.0
   - openfe-analysis>=0.3.1
   - openff-interchange
-  - openff-models>=0.0.5
   - openff-nagl-base >=0.3.3
   - openff-nagl-models>=0.1.2
   - openff-toolkit>=0.16.2
   - openff-units==0.3.1
-  - openmm >=8.2.0
+  - openmm !=8.1.0, <8.3.0
```
atravitz: @mikemhenry is there any reason we'd want a min of 8.2 here? (i know we don't want the 8.2 min on the feedstock)

IAlibay: We probably want a min of 8.1.1, there's stuff we use for salvation that needs it.

mikemhenry: So this is for our CI, we don't have openmm versions in the CI matrix so we are not testing on more than one version. I think we should probably just have it at `==8.2.0` since that is the latest version we have QA'ed and support. Unless something goes crazy with the env/other deps `- openmm !=8.1.0, <8.3.0` is the same thing as I proposed.

atravitz: okay, how about `openmm ~=8.2.0` so we're testing against the lasted qualified version, and also catching any patches that come in?

mikemhenry: Sounds good!

Issue #1406, opened 2025-06-30
Update docs to indicate we support openmm 8.2


mikemhenry:
![Image](https://github.com/user-attachments/assets/bbfae2bb-14bf-41f6-81c4-e1ef8adc7f2f)

PR #1407, opened 2025-06-30
remove openmm exclusion from docs

atravitz:
resolves https://github.com/OpenFreeEnergy/openfe/issues/1406

@mikemhenry opinion on rereleasing 1.5.0 so that this shows up on `stable`?

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -6,7 +6,6 @@ Installation
 We try to follow `SPEC0 <https://scientific-python.org/specs/spec-0000/>`_ as far as minimum supported dependencies, with the following caveats:
 
 - Python 3.10, 3.11, 3.12 - **we do not yet support Python 3.13**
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -6,7 +6,6 @@ Installation
 We try to follow `SPEC0 <https://scientific-python.org/specs/spec-0000/>`_ as far as minimum supported dependencies, with the following caveats:
 
 - Python 3.10, 3.11, 3.12 - **we do not yet support Python 3.13**
-- OpenMM 8.0, 8.1.1, 8.1.2 - **we do not yet support OpenMM v8.2.0**
```
IAlibay, line 9: I would keep "we do not support OpenMM v8.3.0"

atravitz: ah good point

IAlibay (approved):

IAlibay:
Runners won't run because it's docs only.

PR #1425, opened 2025-07-09
Test against 0.4.3

mattwthompson:
ref https://github.com/openforcefield/openff-interchange/pull/1253

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -52,3 +52,4 @@ dependencies:
   - threadpoolctl
   - pip:
     - git+https://github.com/OpenFreeEnergy/gufe@main
+    - git+https://github.com/openforcefield/openff-interchange@release-0.4.3


mattwthompson:
Nothing obviously broken

IAlibay:
Thanks for testing this @mattwthompson - just a heads up, we don't actually do anything directly with Interchange here (at least for now). So as long as relevant tests pass in openmmforcefields, we're probably fine.

The place where testing might be more useful is in pontibus. There we do use Interchange directly, so that might be a good canary for "will things break", especially after we finish implementing that temporary rbfe protocol.

mattwthompson:
Thanks for the added context - this was definitely me just plucking one downstream package to make sure nothing major broke. Your suggestions sound good for other places to check and I'll do those in the future

PR #1435, opened 2025-07-17
updating daily CI runners

atravitz:
Both the package install and ci os matrix tests should run daily with long tests. I'm removing the package install long test every time we push to main, since a daily check on that should be plenty to catch any packaging issues before they show up when we cut a release.

Checklist
* [ ] Added a ``news`` entry


atravitz:
@mikemhenry I'm not sure if the current working directory (`dist/`) would catch `MANIFEST.in` issues. do you have any thoughts on this?

mikemhenry:
That should work!

mikemhenry (approved):

Issue #1418, opened 2025-07-04
restraints: Move things so that dssp filter is ignored if there are too few atoms after we call FindHostAtoms


IAlibay:
This way we can filter by distance and then check if there's not enough atoms.

PR #1440, opened 2025-07-19
Evaluate host idxs atoms in DSSP filters after distance evaluation [Issue #1418]

IAlibay:
When finding host atoms, rather than just looking at the number of atoms in the atomgroup *prior* to evaluating how far they are from the guest anchor, we now check how many atoms there are after distance filtering.

### DIFF
--- a/openfe/tests/protocols/restraints/test_geometry_boresch_host.py
+++ b/openfe/tests/protocols/restraints/test_geometry_boresch_host.py
@@ -34,13 +34,14 @@ def test_host_atom_candidates_dssp(eg5_protein_ligand_universe):
         host_idxs=host_atoms.ix,
         # hand picked
         guest_anchor_idx=5508,
-        host_selection="backbone and resnum 15:25",
+        host_selection="backbone and resnum 212:221",
         dssp_filter=True,
     )
     expected = np.array(
-        [136, 134, 135, 133, 104, 159, 160, 119, 157, 103, 120, 158,  71,
-         118, 102, 101, 117,  70,  87,  86,  85,  69,  68,  88,  51,  50,
-         52,  49,  38,  37,  16,  36,  15,  35,  14,  13]
+        [3144, 3146, 3145, 3143, 3162, 3206, 3200, 3207, 3126, 3201, 3127,
+         3163, 3199, 3202, 3164, 3125, 3165, 3177, 3208, 3179, 3124, 3216,
+         3209, 3109, 3107, 3178, 3110, 3180, 3108, 3248, 3217, 3249, 3226,
+         3218, 3228, 3227, 3250, 3219, 3251, 3229]
     )
     assert_equal(idxs, expected)
 
@@ -62,6 +63,7 @@ def test_host_atom_candidates_dssp_too_few_atoms(eg5_protein_ligand_universe):
             guest_anchor_idx=5508,
             host_selection="backbone and resnum 15:25",
             dssp_filter=True,
+            max_search_distance=2*unit.nanometer
         )
 
 


openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -80,55 +129,65 @@ def find_host_atom_candidates(
...
+            filtered_host_idxs = _host_atoms_search(
+                atomgroup=protein_chain_selection(selected_host_ag),
+                guest_anchor_idx=guest_anchor_idx,
+                rmsf_cutoff=rmsf_cutoff,
+                min_search_distance=min_search_distance,
+                max_search_distance=max_search_distance,
+            )
 
-        if len(stable_ag) < 20:
+        if len(filtered_host_idxs) < 20:
```
atravitz, line 161: naive question, why 20?

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -80,55 +129,65 @@ def find_host_atom_candidates(
...
+    if len(filtered_host_idxs) == 0:
         errmsg = (
             f"No host atoms found within the search distance "
-            f"{min_distance}-{max_distance} consider widening the search window."
+            f"{min_search_distance}-{max_search_distance} consider widening the search window."
         )
         raise ValueError(errmsg)
 
-    # Now we sort them!
+    # Now we sort them from their distance from the guest anchor
```
atravitz:
```suggestion
    # Now we sort them by their distance from the guest anchor
```

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -80,55 +129,65 @@ def find_host_atom_candidates(
...
+            )
 
-        if len(stable_ag) < 20:
+        if len(filtered_host_idxs) < 20:
             wmsg = (
-                "Secondary structure filtering: "
-                "Too few atoms found in protein residue chains, will just "
-                "use all atoms."
+                "Restraint generation: protein chain filter found too few "
+                "host atoms. Will attempt to use all host atoms in "
```
atravitz:
```suggestion
               f"host atoms ({len(filtered_host_idxs)} found). Will attempt to use all host atoms in "
```
same as above

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -80,55 +129,65 @@ def find_host_atom_candidates(
...
+        )
 
-        if len(stable_ag) < 20:
+        if len(filtered_host_idxs) < 20:
             wmsg = (
-                "Secondary structure filtering: "
-                "Too few atoms found via secondary structure filtering will "
-                "try to only select all residues in protein chains instead."
+                "Restraint generation: DSSP filter found too few host atoms "
+                "will attempt to use all protein chains."
```
atravitz:
```suggestion
                f"({len(filtered_host_idxs)} found). Will attempt to use all protein chains."
```
probably useful for user debugging to include # found?

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -80,55 +129,65 @@ def find_host_atom_candidates(
...
+            max_search_distance=max_search_distance,
+        )
 
-    if not atom_finder.results.host_idxs.any():
+    # Crash out if no atoms were found
+    if len(filtered_host_idxs) == 0:
         errmsg = (
             f"No host atoms found within the search distance "
-            f"{min_distance}-{max_distance} consider widening the search window."
+            f"{min_search_distance}-{max_search_distance} consider widening the search window."
```
atravitz:
```suggestion
            f"{min_search_distance}-{max_search_distance}. Consider widening the search window."
```

openfe/protocols/restraint_utils/geometry/boresch/host.py
IAlibay: It's just a heuristic that Hannah and I decided upon. We'll probably refine in the future, but for now it's ~ 6 residues if you're using the backbone.

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -29,15 +29,64 @@
...
     host_idxs: list[int],
-    l1_idx: int,
+    guest_anchor_idx: int,
     host_selection: str,
     dssp_filter: bool = False,
     rmsf_cutoff: Quantity = 0.1 * unit.nanometer,
-    min_distance: Quantity = 1 * unit.nanometer,
-    max_distance: Quantity = 3 * unit.nanometer,
+    min_search_distance: Quantity = 1 * unit.nanometer,
+    max_search_distance: Quantity = 3 * unit.nanometer,
```
hannahbaumann: As you had mentioned, we should think about whether we want to change the default to 2.

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -80,55 +129,65 @@ def find_host_atom_candidates(
...
-        guest_atoms=universe.atoms[l1_idx],
-        min_search_distance=min_distance,
-        max_search_distance=max_distance,
-    )
-    atom_finder.run()
+            filtered_host_idxs = None
+
+    if filtered_host_idxs is None:
+        filtered_host_idxs = _host_atoms_search(
+            atomgroup=selected_host_ag,
```
hannahbaumann, line 172: Would we also want to try out the protein chain selection (as in line 154) when no dssp filter is applied, or just use the selected host ag?

IAlibay: Can you raise an issue for this? We can address it separately.

openfe/protocols/restraint_utils/geometry/boresch/host.py
IAlibay: This is technically being handled in #1463 but I'll also update the numbers here to match the new settings defaults. Longer term I think we'll just remove all the defaults for these methods (there's an issue open somewhere for that).

IAlibay:
```suggestion
    min_search_distance: Quantity = 0.5 * unit.nanometer,
    max_search_distance: Quantity = 1.5 * unit.nanometer,
```

openfe/protocols/restraint_utils/geometry/boresch/host.py
hannahbaumann: Raised an issue!

hannahbaumann:
I think the two tests are failing because the selected residues are now outside the 1.5nm distance
`host_selection="backbone and resnum 15:25",`
maybe we could use
`host_selection="backbone and resnum 212:221"` instead.

hannahbaumann (approved):

PR #1450, opened 2025-07-24
add espaloma tests

### LINKED ISSUE #1449: Add espaloma tests
we have tests for `espaloma_charge` but don't have tests for `espaloma`

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

Fixes #1449

### DIFF
--- a/openfe/tests/protocols/openmm_md/test_plain_md_protocol.py
+++ b/openfe/tests/protocols/openmm_md/test_plain_md_protocol.py
@@ -198,7 +198,7 @@ def test_dry_run_gaff_vacuum(benzene_vacuum_system, tmpdir):
     pytest.param(
         'espaloma', 'rdkit', 'espaloma',
         marks=pytest.mark.skipif(
-            not HAS_ESPALOMA, reason='needs espaloma',
+            not HAS_ESPALOMA, reason='needs espaloma charge',
         ),
     ),
 ])
--- a/openfe/tests/protocols/openmm_md/test_plain_md_protocol.py
+++ b/openfe/tests/protocols/openmm_md/test_plain_md_protocol.py
@@ -15,7 +15,7 @@ from openfe.protocols.openmm_md.plain_md_methods import (
     PlainMDProtocol, PlainMDProtocolUnit, PlainMDProtocolResult,
 )
 from openfe.protocols.openmm_utils.charge_generation import (
-    HAS_NAGL, HAS_OPENEYE, HAS_ESPALOMA
+    HAS_NAGL, HAS_OPENEYE, HAS_ESPALOMA_CHARGE
 )
 import json
 import openfe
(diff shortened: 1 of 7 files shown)


mikemhenry:
We will need to update the logic when we start testing on python 3.14

.github/workflows/ci.yaml
```diff
@@ -61,6 +61,24 @@ jobs:
         run: echo "date=$(date +%Y-%m-%d)" >> "${GITHUB_OUTPUT}"
 
       - name: "Setup Micromamba"
+        if: ${{ matrix.python-version != '3.13' }}
```
atravitz, line 64: do you expect this to work with python 3.14, or should this be `<'3.13'`?

atravitz (commented):
just one question

openfe/tests/protocols/openmm_md/test_plain_md_protocol.py
```diff
@@ -15,13 +15,14 @@
...
 from openfe.protocols.openmm_utils.charge_generation import (
-    HAS_NAGL, HAS_OPENEYE, HAS_ESPALOMA
+    HAS_NAGL, HAS_OPENEYE, HAS_ESPALOMA_CHARGE
 )
 import json
 import openfe
 from openfe.protocols import openmm_md
 import pathlib
 import logging
+from openfe.tests.conftest import HAS_INTERNET, HAS_ESPALOMA
```
IAlibay:
```suggestion
from openfe.tests.conftest import HAS_ESPALOMA
```
Not using internet check anywhere?

openfe/tests/protocols/openmm_md/test_plain_md_protocol.py
```diff
@@ -180,6 +181,27 @@ def test_dry_run_gaff_vacuum(benzene_vacuum_system, tmpdir):
         system = unit.run(dry=True)["debug"]["system"]
 
 
+@pytest.mark.skipif(not HAS_ESPALOMA, reason='espaloma is not available')
+def test_dry_run_espaloma_vacuum(benzene_vacuum_system, tmpdir):
```
IAlibay, line 185: Ideally, we should check more things, like if the partial charges made it through, etc.. I'm ok with this as a smoke test, but we should raise a separate issue to try it out in a production setting.

IAlibay (changes requested):

.github/workflows/ci.yaml
mikemhenry: idk how to do a less than for a string comparison, it really depends what happens in conda-forge land for rebuilds for different python versions.

openfe/tests/protocols/openmm_md/test_plain_md_protocol.py
mikemhenry: Agreed, I'll raise an issue, maybe we can add a GPU test where we run some MD with parameters from espaloma.

PR #1477, opened 2025-08-13
Change bonded host strategy to use lowest variance restraint

### LINKED ISSUE #1471: Derive single-residue host anchor boresch restraint via variance tiebreak
Follow-up from #1470

IAlibay:
Fixes #1471 

### DIFF
--- a/openfe/protocols/restraint_utils/geometry/boresch/host.py
+++ b/openfe/protocols/restraint_utils/geometry/boresch/host.py
@@ -27,6 +27,7 @@ from openfe.protocols.restraint_utils.geometry.utils import (
     stable_secondary_structure_selection,
 )
 from openff.units import Quantity, unit
+from scipy.stats import circvar
 
 
 def _host_atoms_search(
@@ -196,14 +197,13 @@ def find_host_atom_candidates(
 
 class EvaluateBoreschAtoms(AnalysisBase):
     """
-    Class to evaluate the suitability of the atoms in a Boresch
-    restraint.
+    Class to evaluate the geometric suitability of Boresch restraints.
 
     Parameters
     ----------
-    restraint : MDAnalysis.AtomGroup
-      An AtomGroup defining the H2-H1-H0-G0-G1-G2 atoms, in that
-      order.
+    restraints : list[MDAnalysis.AtomGroup]
+      A list of AtomGroup defining the H2-H1-H0-G0-G1-G2 atoms,
+      in that order.
     angle_force_constant : openff.units.Quantity
       The force constant for the angle.
     temperature : openff.units.Quanity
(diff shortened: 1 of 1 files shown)


openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -533,6 +548,70 @@ def _conclude(self):
...
+            ])
+
+            ang_variance = sum([
+                circvar(angles, high=np.pi, low=0)
+                for angles in restraints_eval.results.angles[ridx]
+            ])
+
+            bond_variance = np.var(restraints_eval.results.bonds[ridx])
+
+            valid_variances.append(dih_variance + ang_variance + bond_variance)
```
IAlibay, line 605: Only using a linear combination of variances here. In the past, this "worked", but probably just as well as any other approaches. Happy to change it to something else if there's a better combination we can try.

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -196,122 +197,136 @@ def find_host_atom_candidates(
...
-        restraint: mda.AtomGroup,
+        restraints: list[mda.AtomGroup],
         angle_force_constant: Quantity,
         temperature: Quantity,
         **kwargs,
     ):
-        super().__init__(restraint.universe.trajectory, **kwargs)
+        if len(restraints) < 1:
+            errmsg = "Need to have at least one restraint"
+            raise ValueError(errmsg)
```
IAlibay, line 221: TODO: I want to add a few tests for these ValueErrors

openfe/protocols/restraint_utils/geometry/boresch/host.py
```diff
@@ -196,122 +197,136 @@ def find_host_atom_candidates(
...
-        self.results.valid = all(
-            [
-                angle_bounds,
-                angle_variance,
-                dihed_bounds,
-                dihed_variance,
-                not_collinear,
-            ]
-        )
+            not_collinear = not all(self.results.collinear)
```
hannahbaumann:
```suggestion
            not_collinear = not all(self.results.collinear[ridx])
```
This part is currently failing.

hannahbaumann (commented):
Thanks @IAlibay , lgtm, just the one thing. I tested this on one trajectory and it gave reasonable atoms. It does, however, take quite some time.

IAlibay:
> It does, however, take quite some time.

Do you have a ballpark timing @hannahbaumann ? i.e. is it order minute or tens of minutes or closer to hour? 

This being quite expensive is somewhat expected, the number of restraints it searches through pretty much explodes with the number of potential host anchor atoms it can find. How large was the max cutoff distance for host atoms? You'd want to keep that quite small, probably no larger than 1.5 nm.

hannahbaumann:
More like 30sec, so actually not that bad!

IAlibay:
> More like 30sec, so actually not that bad!

O yeah, I was really expecting up to 10s of minutes! I think the search space grows by ~ 10-12 potential restraints per residue (if only doing backbone atoms), so it gets to pretty large numbers quickly if we're not careful.

openfe/protocols/restraint_utils/geometry/boresch/host.py
IAlibay: 🤔 I'm confused as to why the long tests didn't pick this up, will need to look into it.

PR #1479, opened 2025-08-15
Move ligand utils to gufe

atravitz:
partially addresses #560, along with https://github.com/OpenFreeEnergy/gufe/pull/602

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -51,4 +51,4 @@ dependencies:
   # Control blas/openmp threads
   - threadpoolctl
   - pip:
-    - git+https://github.com/OpenFreeEnergy/gufe@migrate_ligand_utils_from_openfe
+    - git+https://github.com/OpenFreeEnergy/gufe@main
--- a/environment.yml
+++ b/environment.yml
@@ -51,4 +51,4 @@ dependencies:
   # Control blas/openmp threads
   - threadpoolctl
   - pip:
-    - git+https://github.com/OpenFreeEnergy/gufe@main
+    - git+https://github.com/OpenFreeEnergy/gufe@migrate_ligand_utils_from_openfe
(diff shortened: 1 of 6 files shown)


mikemhenry:
This looks good -- it will require us to pin to the version of gufe we release with `LigandAtomMapping.get_alchemical_charge_difference()` I will approve this BUT something to maybe consider is we keep the deprecation warning in for a release before we require an update to the latest gufe. 

P.S. This argument only holds water if there is some reason a user wouldn't want to update to the latest gufe, AFAIK there isn't any new deps or major changes that warrant this so I believe it is a moot point

environment.yml
```diff
@@ -51,4 +51,4 @@ dependencies:
   # Control blas/openmp threads
   - threadpoolctl
   - pip:
-    - git+https://github.com/OpenFreeEnergy/gufe@main
+    - git+https://github.com/OpenFreeEnergy/gufe@migrate_ligand_utils_from_openfe
```
mikemhenry: Set this back to main before merging :)

mikemhenry (approved):

atravitz:
> This looks good -- it will require us to pin to the version of gufe we release with `LigandAtomMapping.get_alchemical_charge_difference()` I will approve this BUT something to maybe consider is we keep the deprecation warning in for a release before we require an update to the latest gufe.
> 
> P.S. This argument only holds water if there is some reason a user wouldn't want to update to the latest gufe, AFAIK there isn't any new deps or major changes that warrant this so I believe it is a moot point

I agree - if this were moving to any package _other_ than gufe, I would do a deprecation cycle before moving the code. But since gufe and openfe are closely tied and we make releases together, I think this is fine.

PR #1487, opened 2025-08-22
Bump number of retires when fetching data with pooch

mikemhenry:
I've seen a few cases in the wild where CI fails because a file failed to download. I think we should try bumping to 5 since with each retry it increases the wait time and I think that ~3s might be too fast so I want to try 5s.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,9 +1,9 @@
 version: 2
 
 build:
-  os: "ubuntu-20.04"
+  os: "ubuntu-24.04"
   tools:
-    python: "mambaforge-4.10"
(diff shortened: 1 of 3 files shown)


mikemhenry:
For example, this release failed in CI https://github.com/conda-forge/openfe-feedstock/pull/98 with `E           requests.exceptions.HTTPError: 503 Server Error: Service Unavailable for url: https://files.rcsb.org/download/6CZJ.pdb`

atravitz:
worth a shot!

atravitz (approved):

PR #1506, opened 2025-09-10
update partial charge tests now that a production NAGL model is available

atravitz:
I'm assuming we still want to have a helpful error message in case a production NAGL model isn't found - or we can omit that check if we trust that an upstream openff error message will make sense to an openfe user.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/protocols/openmm_utils/charge_generation.py
+++ b/openfe/protocols/openmm_utils/charge_generation.py
@@ -142,12 +142,11 @@ def assign_offmol_nagl_charges(
         prod_models = get_models_by_type(
             model_type='am1bcc', production_only=True
         )
-        # Currently there are no production models so expect an IndexError
         try:
             nagl_model = prod_models[-1]
         except IndexError:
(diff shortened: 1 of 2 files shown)


openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -139,15 +140,14 @@ def assign_offmol_nagl_charges(
         raise ImportError(errmsg)
 
     if nagl_model is None:
-        prod_models = get_models_by_type(
+        prod_models = openff.nagl_models.get_models_by_type(
             model_type='am1bcc', production_only=True
```
IAlibay, line 143: If you change `model_type` to something else like `AM1`, it may trigger the same issue (we would need to update the error message though).

IAlibay: Nevermind ignore me, I'm being silly - thinking of something ellse (possible follow-up work, but not here)

PR #1530, opened 2025-09-22
Change the default atom mapping in the CLI to Kartograf

### LINKED ISSUE #1493: Change default atom mapper to Kartograf?
The default CLI atom mapper should probably match what we used in the industry benchmarks.

Lomap is still default in the CLI, we should revisit as we barely benchmark the Lomap atom mapper.

jthorton:
Fixes #1493 by changing the default atom mapper used in the CLI to Kartograf.

Note:
- I added the flag to map hydrogens to hydrogens only, as these are later removed in the protocol anyway, it should make the mapping visualisation clearer as well, and this is planned to be the new default setting in kartograf in the future. 

Checklist
* [x] Added a ``news`` entry


jthorton:
We do have a planning test for the CLI, but in this case we are lucky and got the same edges (tyk2 has good alignment and we made the edge order reproducible!), so no test failed. I have added a check for the default settings used in the CLI by matching some info in the output stream. This could be improved in future when we have provenance on the `LigandNetwork` and `AlchemicalNetwork` objects.

mikemhenry:
wow called out 
image
did we decide that this isn't a 2.0 breaking change since we don't promise that the CLI defaults don't change?

IAlibay:
> did we decide that this isn't a 2.0 breaking change since we don't promise that the CLI defaults don't change?

Yeah - board approved it too. We're selling it as "this is what was used in the industry benchmarks".

atravitz (approved):
LGTM, and I agree with @jthorton about adding more tests once we store this info.

openfecli/parameters/plan_network_options.py
```diff
@@ -241,8 +244,8 @@ def load_yaml_planner_options(path: Optional[str], context) -> PlanNetworkOption
 
 \b
 Supported atom mapper choices are:
-    - ``LomapAtomMapper`` (default)
-    - ``KartografAtomMapper``
+    - ``KartografAtomMapper`` (default)
```
atravitz: one thought
```suggestion
    - ``KartografAtomMapper`` (default as of v1.7.0)
```

PR #1536, opened 2025-09-24
Septop pydantic v2

atravitz:
NOTE: this targets the `to_pydantic_v2` branch so that I can get @hannahbaumann's review on just the septop pydanticv2 updates.

### DIFF
--- a/openfe/tests/protocols/test_openmm_settings.py
+++ b/openfe/tests/protocols/test_openmm_settings.py
@@ -77,7 +77,7 @@ class TestOpenMMSolvationSettings:
                 [0.0, 1.0, 0.0],
                 [1.0, 0.0, 0.0],
             ],
-        # add units, openfe doesn't duck-type but interchange does
+        # add units, openfe doesn't assume nanometers but interchange does
         ) * unit.nanometer
 
         with pytest.raises(ValueError, match="not in OpenMM reduced form"):


openfe/protocols/openmm_septop/equil_septop_method.py
```diff
@@ -773,7 +773,7 @@ def _default_settings(cls):
             forcefield_settings=settings.OpenMMSystemGeneratorFFSettings(),
             thermo_settings=settings.ThermoSettings(
                 temperature=298.15 * unit.kelvin,
-                pressure=1 * unit.bar,
+                pressure=1 * unit.bar,  # TODO: this is converted to atm as of gufe v1.7.0, do we still want 1 bar, or 1 atm?
```
atravitz: @hannahbaumann - how do you feel about this? Is `bar` necessary here, or can we define this in `atm`? 
On the gufe side, we decide to convert all pressure units to atm for consistency.

atravitz:
@hannahbaumann I'd like your feedback here - I see you use openmm units in a few places here, and I'm wondering if we can convert these all to openff units?

openfe/protocols/openmm_septop/equil_septop_method.py
hannahbaumann: atm would be fine too! Would you then also change that in the other protocols? I think as long as we're consistent between protocols it should be fine.
Atm number would just be less pretty than just having a 1?!

openfe/tests/protocols/openmm_septop/test_septop_protocol.py
```diff
@@ -674,7 +674,7 @@ def test_dry_run_benzene_toluene(benzene_toluene_dag, tmpdir):
         assert isinstance(
             solv_sampler._thermodynamic_states[0].barostat, MonteCarloBarostat
         )
-        assert solv_sampler._thermodynamic_states[1].pressure == 1 * openmm.unit.bar
+        assert solv_sampler._thermodynamic_states[1].pressure == 1 * unit.bar
```
hannahbaumann: This won't work since the sampler is returning an openmm unit. Would need a to_openmm, or the openmm unit. The other protocols are also using the openmm unit here.

openfe/tests/protocols/openmm_septop/test_septop_protocol.py
```diff
@@ -447,9 +447,9 @@ def three_particle_system():
     sigmas = 1.1, 1.2, 1.3
     epsilons = 210, 220, 230
 
-    force.addParticle(charges[0], sigmas[0] * openmm.unit.angstrom, epsilons[0])
-    force.addParticle(charges[1], sigmas[1] * openmm.unit.angstrom, epsilons[1])
-    force.addParticle(charges[2], sigmas[2] * openmm.unit.angstrom, epsilons[2])
+    force.addParticle(charges[0], sigmas[0] * unit.angstrom, epsilons[0])
```
hannahbaumann: These need to be openmm unit since it's an OpenMM force

hannahbaumann (commented):
Thanks @atravitz !
Regarding the switch of openmm to openff units, for the tests that are using openmm (sampler, forces,..) I think I would stick to openmm units, otherwise you'd need a `to_openmm` for many things. I think other protocols are also sticking to openmm units for these.

openfe/protocols/openmm_septop/equil_septop_method.py
hannahbaumann: @atravitz we just discussed this in  a meeting and decided that for now it would be better to stick with bar in all protocols.

openfe/tests/protocols/openmm_septop/test_septop_protocol.py
IAlibay: Roughly, the advice here is to avoid to/from unit conversions in tests and just test with native units where possible. Partly because openff-units' to/from methods are not guaranteed stable, so we're just adding an extra layer of things that could potentially fail.

Note: this isn't consistent across our tests, it's something we can update when someone has a spare moment.

openfe/protocols/openmm_septop/equil_septop_method.py
atravitz: got it! we should decide this at the gufe level, so PR is here: https://github.com/OpenFreeEnergy/gufe/pull/637.

IAlibay: I vote bar over atm, it's more representative of what Gromacs and OpenMM does by default - i.e. they both internally represent things in Bar.

openfe/tests/protocols/openmm_septop/test_septop_protocol.py
atravitz: makes sense to me, thanks!

openfe/tests/protocols/openmm_septop/test_septop_protocol.py
```diff
@@ -482,7 +483,7 @@ def interaction_energy_fn(
         np.array(
             [[0.0, 0.0, 0.0], [distances[0][1], 0.0, 0.0], [0.0, distances[0][2], 0.0]]
         )
-        * openmm.unit.angstrom
+        * unit.angstrom
```
IAlibay:
```suggestion
        * openmm.unit.angstrom
```
Or whatever variant, I'm pretty sure this gets thrown into `compute_energy` directly, which expects OpenMM units.

atravitz: ah, thanks for catching this!

atravitz:
CI running at: https://github.com/OpenFreeEnergy/openfe/actions/workflows/ci.yaml?query=branch%3Aseptop_pydantic_v2 (not automatic because this isn't merging into `main`)

~merge blocked by https://github.com/OpenFreeEnergy/gufe/pull/637~ no longer blocked

ready for review @hannahbaumann!

PR #1538, opened 2025-09-24
Only run itergration tests on AWS GPU and CPU runners

### LINKED ISSUE #1527: add gpu test labels
we want to be able to run only gpu tests easily

mikemhenry:
Checklist
* [ ] Added a ``news`` entry
-m MARKEXPR           Only run tests matching given mark expression. For example: -m 'mark1 and not mark2'.

Fixes #1527

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,12 @@
+# https://github.com/OpenFreeEnergy/openfe/pull/1604 - ruff formatting part 1
+3a08b6809fc57662e4146db3c7ccedfbc7c7c8df  
+
+# https://github.com/OpenFreeEnergy/openfe/pull/1610 - ruff formatting part 2
+2311a2f2d956dd30e95c180841ce19b921d89e1f
+
+# https://github.com/OpenFreeEnergy/openfe/pull/1622 - ruff formatting part 3
+d7196d119e2f65d88e488afc665f2521e4f68042
+
+# https://github.com/OpenFreeEnergy/openfe/pull/1623 - ruff formatting part 4
+036869ae81670c6dcfa2532f125ee88c3f35936c
+


mikemhenry:
Testing CPU here: https://github.com/OpenFreeEnergy/openfe/actions/runs/17988568861

.github/workflows/ci.yaml
```diff
@@ -9,6 +9,8 @@ on:
     paths-ignore:
       - "docs/*"
       - ".readthedocs.yaml"
+      - ".github/workflows/cpu-long-tests.yaml"
+      - ".github/workflows/gpu-integration-tests.yaml"
```
mikemhenry, line 13: Adding skips here so that when we make some small change to our action that runs on AWS we don't fire off the whole testing matrix.

mikemhenry:
I actually forgot that we don't run the integration tests on the CPU runner -- do we only want to run the slow tests on the CPU runner @IAlibay @atravitz ? Right now we run all the "normal" tests + the slow tests on the CPU aws runner.

mikemhenry:
GPU testing here https://github.com/OpenFreeEnergy/openfe/actions/runs/17989853748

mikemhenry:
GPU test here: https://github.com/OpenFreeEnergy/openfe/actions/runs/17991765097/job/51183190807

Looks like one failed -- should we add some retry logic to the test?
```
FAILED openfe/tests/protocols/openmm_ahfe/test_ahfe_slow.py::test_openmm_run_engine[CPU] - openmmtools.multistate.utils.SimulationNaNError: Propagating replica 0 at state 0 resulted in a NaN!
The state of the system and integrator before the error were saved in /tmp/pytest-of-root/pytest-0/popen-gw3/test_openmm_run_engine_CPU_3/shared_AbsoluteSolvationSolventUnit-
```

mikemhenry:
Took 40 minutes:
```
======= 1 failed, 7 passed, 158 warnings, 3 rerun in 2346.37s (0:39:06) ========
```

mikemhenry:
@atravitz @IAlibay Ready for review! The goal here was to just run the integration tests and not our other tests when using the GPU and CPU runner.

atravitz:
> GPU test here: https://github.com/OpenFreeEnergy/openfe/actions/runs/17991765097/job/51183190807
> 
> Looks like one failed -- should we add some retry logic to the test?
> 
> ```
> FAILED openfe/tests/protocols/openmm_ahfe/test_ahfe_slow.py::test_openmm_run_engine[CPU] - openmmtools.multistate.utils.SimulationNaNError: Propagating replica 0 at state 0 resulted in a NaN!
> The state of the system and integrator before the error were saved in /tmp/pytest-of-root/pytest-0/popen-gw3/test_openmm_run_engine_CPU_3/shared_AbsoluteSolvationSolventUnit-
> ```

tagging @IAlibay to ask how expected this is.

atravitz:
> I actually forgot that we don't run the integration tests on the CPU runner -- do we only want to run the slow tests on the CPU runner @IAlibay @atravitz ? Right now we run all the "normal" tests + the slow tests on the CPU aws runner.

Hmm, I think the CPU runner can just be slow tests - I know @IAlibay specifically wanted this for when he was doing development away from his workstation. I think unit tests locally and then slow tests on the runner should still meet his need?

mikemhenry:
Okay CPU runner just doing the slow tests now, testing that here:
https://github.com/OpenFreeEnergy/openfe/actions/runs/18412521347

IAlibay:
> > GPU test here: https://github.com/OpenFreeEnergy/openfe/actions/runs/17991765097/job/51183190807
> > Looks like one failed -- should we add some retry logic to the test?
> > ```
> > FAILED openfe/tests/protocols/openmm_ahfe/test_ahfe_slow.py::test_openmm_run_engine[CPU] - openmmtools.multistate.utils.SimulationNaNError: Propagating replica 0 at state 0 resulted in a NaN!
> > The state of the system and integrator before the error were saved in /tmp/pytest-of-root/pytest-0/popen-gw3/test_openmm_run_engine_CPU_3/shared_AbsoluteSolvationSolventUnit-
> > ```
> 
> tagging @IAlibay to ask how expected this is.

Shouldn't be happening very often, we can throw a retry in there though, wouldn't hurt.

.github/workflows/cpu-long-tests.yaml
```diff
@@ -91,7 +92,7 @@ jobs:
           DUECREDIT_ENABLE: 'yes'
           OFE_INTEGRATION_TESTS: FALSE
         run: |
-          pytest -n logical -vv --durations=10 openfecli/tests/ openfe/tests/
+          pytest -n logical -vv --durations=10 -m slow openfecli/tests/ openfe/tests/
```
IAlibay: Why do we need `-m slow`? I would have though the env variable would be enough.

IAlibay:
> > I actually forgot that we don't run the integration tests on the CPU runner -- do we only want to run the slow tests on the CPU runner @IAlibay @atravitz ? Right now we run all the "normal" tests + the slow tests on the CPU aws runner.
> 
> Hmm, I think the CPU runner can just be slow tests - I know @IAlibay specifically wanted this for when he was doing development away from his workstation. I think unit tests locally and then slow tests on the runner should still meet his need?

Here's what my view is on what we want:

### AWS CPU runners

Normal tests: yes
Slow tests: yes
Integration tests: ~ no (see below)

### AWS GPU runners

Normal tests: if we have to
Slow tests: no
Integration tests: ~ yes (see below)

### What's wrong with our integration tests?

The most important thing is that we don't use GPU runners in places we don't need them, i.e. we don't want to use GPU dollars running a bunch of CPU-only tests.

What we would want, is an additional flag that ONLY runs GPU tests, and then we can run any integration tests that need CPU on a specific CPU integration runner.

.github/workflows/gpu-integration-tests.yaml
```diff
@@ -95,7 +97,7 @@ jobs:
           DUECREDIT_ENABLE: 'yes'
           OFE_INTEGRATION_TESTS: TRUE
         run: |
-          pytest -n logical -vv --durations=10
+          pytest -n logical -vv --durations=10 -m integration openfecli/tests/ openfe/tests/
```
IAlibay, line 99: We don't want to run slow tests with the GPU runners, otherwise we're spending lots of $$ doing non-gpu things.

mikemhenry:
Okay to expand on what I wrote here: https://github.com/OpenFreeEnergy/openfe/issues/1527#issuecomment-3330549036

The `-m` flag 
> Only run tests matching given mark expression. For example: -m 'mark1 and not mark2'.

Which means `-m integration` ONLY runs tests with the `@pytest.mark.integration` (so since normal and slow tests do not have that mark, they do not get run). 

This will satisfy what you want for the GPU runners. 

For the CPU runners, since we want slow + normal, that is easy to do with flags/args.

> What we would want, is an additional flag that ONLY runs GPU tests, and then we can run any integration tests that need CPU on a specific CPU integration runner.

The `-m integration` flag does exactly that! So we should be good on the GPU runner :sunglasses:

.github/workflows/gpu-integration-tests.yaml
mikemhenry: The `-m` flag will only run tests with the `@pytest.mark.integration` decoration

.github/workflows/cpu-long-tests.yaml
mikemhenry: I had thought we only wanted to run the slow tests (and not the normal ones) on the CPU runner, will fix this!

IAlibay: That could be ok too.

IAlibay:
Ah ok thanks for the explanation @mikemhenry , I hadn't caught that.

mikemhenry:
No worries! I could have sworn I explained the `-m` flag but I didn't, and even in the issue I linked I actually didn't explain it either. I will add a comment to the workflow since I will forget about the flag I'm sure

.github/workflows/gpu-integration-tests.yaml
atravitz: does this mean that anything marked with `@pytest.mark.integration` is a gpu test? If so, I'd prefer the naming explicitly say `gpu` in the name, since "integration" is a bit of a grey area.

IAlibay: No, not everything named integration is a GPU test. A lot of them are both GPU and CPU. However, untangling them might not be within scope, since it'll require us to work out what requires CPU coverage and what doesn't.

mikemhenry:
I guess it depends if all integration tests are GPU tests, or if all GPU tests will be integration tests. If we don't think there is anything interesting there, then we can rename `integration -> gpu` 

Thoughts @IAlibay

IAlibay:
> I guess it depends if all integration tests are GPU tests, or if all GPU tests will be integration tests. If we don't think there is anything interesting there, then we can rename `integration -> gpu`
> 
> Thoughts @IAlibay

All integration tests are GPU, but not all integration tests are ONLY GPU.

Very roughly, if we can control the OpenMM platform with pytest marks, then that's the answer. We stick on a gpu and cpu mark and then use the name of the active mark to pick the platform (or skip the test).

If you gimme some pseudo code I'm happy to show a demo of what I mean.

IAlibay:
This! https://stackoverflow.com/a/74804492

We just need to create a gpu and cpu mark and then check for the mark's presence and pick from a dictionary, with a prerference for GPU over CPU if both are available.

mikemhenry:
I think the three of us are either kinda talking past each other or I don't fully understand the scope of what we are trying to do. As this PR currently stands:


* Only tests marked `integration` run on the GPU runner
* "normal" tests and tests marked slow run on the CPU runner.

If this isn't quite what we are trying to do, let me know! 

It sounds like there are some `integration` tests that we want to run on a CPU, which ones are those (do you have some in mind or is there some heuristic to choosing these)?

IAlibay:
> I think the three of us are either kinda talking past each other or I don't fully understand the scope of what we are trying to do. As this PR currently stands:
> 
> * Only tests marked `integration` run on the GPU runner
> * "normal" tests and tests marked slow run on the CPU runner.
> 
> If this isn't quite what we are trying to do, let me know!
> 
> It sounds like there are some `integration` tests that we want to run on a CPU, which ones are those (do you have some in mind or is there some heuristic to choosing these)?

I agree, we're going out of scope, the untangling of CPU & GPU tests for the integration tests can be done in a separate PR. @atravitz are you happy with that?

mikemhenry:
I think @atravitz main point was to re-name the `integration` mark to `gpu` since it functional (now) marks tests we want to run on the GPU. I think we should keep them labeled as `integration` and then we can untangle them (maybe `integration-gpu` and `integration-cpu`) in another PR.

atravitz:
>I think we should keep them labeled as `integration` and then we can untangle them (maybe `integration-gpu` and `integration-cpu`) in another PR.

I agree!

PR #1539, opened 2025-09-24
Removing unnecessary limit on resids

ijpulidos:
This change would allow not having an unnecessary limit on the `resids` when getting mappings from a topology. This is needed for current implementation of the protein mutation protocol.

Checklist
* [x] Added a ``news`` entry


IAlibay (approved):
Thanks @ijpulidos - sorry it took a bit longer, was really confused by the docstring.

Issue #1473, opened 2025-08-12
Store the selection indices in protocol results


IAlibay:
That way we can map the selected indices to the full sytem indices.

hannahbaumann:
In numpy.array format. Put it into the results object. Add method to get them out.

PR #1557, opened 2025-10-08
Disable user supplied host & guest atoms in restraint settings

### LINKED ISSUE #1555: Disable user-supplied Boresch restraints

IAlibay:
Fixes #1555 
Partially addresses #1473 

### DIFF
--- a/openfe/protocols/openmm_septop/equil_septop_method.py
+++ b/openfe/protocols/openmm_septop/equil_septop_method.py
@@ -745,8 +745,9 @@ class SepTopProtocolResult(gufe.ProtocolResult):
         -------
         indices : dict[str, list[npt.NDArray]]
           A dictionary keyed as `complex` and `solvent` for each
-          state, each containing a list of NDArrays containing the
-          indices for each replica.
+          state, each containing a list of NDArrays containing the corresponding
+          full system atom indices for each atom written in the production
+          trajectory files for each replica.
         """
         indices: dict[str, list[Optional[npt.NDArray]]] = {}
 


IAlibay:
@hannahbaumann I need to re-create the ProtocolResults for SepTop. How did you create them? We probably should add a relevant script to `devtools`.

openfe/protocols/openmm_septop/equil_septop_method.py
```diff
@@ -736,6 +736,29 @@ def restraint_geometries(self) -> tuple[list[dict[str, Any]], list[dict[str, Any
...
+        """
+        Get the system selection indices used to write PDB and
+        trajectory files.
+
+        Returns
+        -------
+        indices : dict[str, list[npt.NDArray]]
+          A dictionary keyed as `complex` and `solvent` for each
+          state, each containing a list of NDArrays containing the
+          indices for each replica.
```
hannahbaumann: Maybe this could be more specific, like the output atom indices or similar?

IAlibay:
> what an old atom index (in the full system) is in the output indices, e.g. if there were crystallographic waters between the ligand and the protein. Or maybe I'm missing something?

@hannahbaumann My understanding of the original request is that we didn't have a good way to work out what atoms the indices from the Boresch restraints matched the indices in the written trajectory files. This effectively allows you to do that, i.e. you can look at the Boresch restraint values, load up your trajectory and find the equivalence in the indices based on where they are in the selection_indices.

Making an equivalence between selection indices and the input Components is a lot harder - I'm not sure it'd be in scope, unless I misunderstood the original ask.

hannahbaumann:
Ah sorry, I had a brain freeze, thinking the indices stored will be the "new" indices, but you're right, since it's the original atom indices, based on their position we can easily backtrack.

openfe/protocols/openmm_septop/equil_septop_method.py
IAlibay: Done, I _think_, words are a bit hard today, so please do tell me if it still makes no sense.

hannahbaumann: This sounds great!

PR #1564, opened 2025-10-09
Fix test to check if we are printing `--allow-partial` when a user uses `--allow-partial`

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfecli/tests/commands/test_gather.py
+++ b/openfecli/tests/commands/test_gather.py
@@ -413,10 +413,10 @@ class TestRBFEGatherFailedEdges:
             assert "The results network is disconnected" in str(result.stderr)
 
 
-    def test_missing_leg_allow_partial_(self, results_paths_serial_missing_legs: str):
-        runner = CliRunner()
+    def test_allow_partial_msg_not_printed(self, results_paths_serial_missing_legs: str):
         # we *dont* want the suggestion to use --allow-partial if the user already used it!
-        with pytest.warns(match=r'[^using the \-\-allow\-partial]'):
-            args =  ["--report", "ddg", "--allow-partial"]
-            result = runner.invoke(gather, results_paths_serial_missing_legs + args + ['--tsv'])
-            assert_click_success(result)
+        runner = CliRunner()
+        args =  ["--report", "ddg", "--allow-partial"]
+        result = runner.invoke(gather, results_paths_serial_missing_legs + args + ['--tsv'])
+        assert_click_success(result)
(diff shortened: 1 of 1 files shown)


mikemhenry:
@atravitz I am a bit confused by this regex, what are we to do with this test? I feel like `# we *dont* want the suggestion to use --allow-partial if the user already used it!` implies we are trying to make sure that if a user has `--allow-partial` we DON'T want it to warn us that they can use that option. But I think the regex is the same as `match='[^using the\-alowpr]'` (the test still passes if I use that match string) Which is like "match if there is a warning that doesn't have any of these chars" 

So then when we get a warning like `ResourceWarning("unclosed file <_io.TextIOWrapper name='/home/mmh/.cache/openfe/rbfe_results_serial_repeats.tar.gz.untar/rbfe_results_serial_repeats/rbfe_lig_ejm_46_solvent_lig_jmc_28_solvent/shared_RelativeHybridTopologyProtocolUnit-60df96a488094f63978880d19487d299_attempt_1/db.json' mode='r' encoding='UTF-8'>")` since it has some chars that don't match, the regex matches so then the test passes.

I don't think `pytest.warns` works for warnings that don't come from python API and I think we are using it here to check for warnings that we expect to be printed to std out/err, so what we need to do is loop though what the command returns and check to make sure we don't see that string in the output. 

At least that is what I took away from a quick look at this, I could be totally wrong!

mikemhenry:
I saw the warning about the escape syntax when I was working on an unrelated thing and you know how I like a good investigation on an esoteric code path!

This fixes the escape warning, but what are we trying to test with this test?

openfecli/tests/commands/test_gather.py
```diff
@@ -416,7 +416,7 @@ def test_missing_leg_allow_partial_disconnected(self, results_paths_serial_missi
     def test_missing_leg_allow_partial_(self, results_paths_serial_missing_legs: str):
         runner = CliRunner()
         # we *dont* want the suggestion to use --allow-partial if the user already used it!
```
atravitz, line 417: @mikemhenry is this comment unclear? I added this test because when we ran with `--allow-partial`, it was still adding the `you can use --allow-partial` to stdout, which I think is confusing.

atravitz: I missed your earlier comment before, and now I understand why this test is insufficient.  Is your suggestion that we assert that `you can use --allow-partial` is not in stdout?

PR #1567, opened 2025-10-10
add yaml syntax check precommit hooks

atravitz:
This is a first pass at moving us to a manual trigger for precommit so that we can add more opinionated hooks (like ruff) down the line.

- **add more checks**
- **make precommit manual**

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,6 +9,8 @@ on:
     paths-ignore:
       - "docs/*"
       - ".readthedocs.yaml"
+      - ".github/workflows/cpu-long-tests.yaml"
+      - ".github/workflows/gpu-integration-tests.yaml"
   push:
     branches:
       - main
--- a/.github/workflows/cpu-long-tests.yaml
+++ b/.github/workflows/cpu-long-tests.yaml
@@ -60,6 +60,7 @@ jobs:
             espaloma_charge==0.0.8
             espaloma==0.4.0
             openeye-toolkits
+            python=3.12
 
       - name: "Check if OpenMM can get a GPU"
         run: python -m openmm.testInstallation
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,22 +1,23 @@
...
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v6.0.0
   hooks:
   - id: check-added-large-files
+  - id: check-case-conflict
+  - id: check-executables-have-shebangs
+  - id: check-symlinks
+  - id: check-toml
+  # - id: check-yaml  # TODO: resolve violation in devtools/installer/construct.yaml
```
atravitz: @mikemhenry do you have thoughts on this?

mikemhenry: I think we just need to mark that file as a skip, it isn't a true yaml file.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ classifiers = [
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
   "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
```
IAlibay, line 28: Aren't we blocked by OpenMM 8.4?

atravitz: resolved by https://github.com/OpenFreeEnergy/openfe/pull/1629

mikemhenry:
This PR actually doesn't make it manual, but enables a YAML syntax check + disables that check for a file that is a mix of yaml and jinja

mikemhenry:
I just changed the tile to make it clear at a glance in the log

atravitz:
> This PR actually doesn't make it manual, but enables a YAML syntax check + disables that check for a file that is a mix of yaml and jinja

ah, thank you! the original changes in this ended up getting merged with the first ruff PR. Thanks for keeping the commit log clean

PR #1572, opened 2025-10-12
Fix validation in AHFE Protocol

IAlibay:
1. Move the validation to the `validate` method.
2. Add a check to disallow net charged ligands.
3. Fix some tests.

Checklist
* [x] Added a ``news`` entry


openfe/protocols/openmm_afe/equil_solvation_afe_method.py
```diff
@@ -559,26 +524,55 @@ def _validate_alchemical_components(
           SmallMoleculeComponents.
         * Currently doesn't support more than one alchemical component
           being desolvated.
+        * Currently doesn't support charged alchemical components.
+        * Solvent must always be present in both end states.
         """
+        # Check that there are no protein components
+        if stateA.contains(ProteinComponent) or stateB.contains(ProteinComponent):
+            errmsg = ("Protein components are not allowed for "
+                      "absolute solvation free energies")
```
hannahbaumann: Maybe are not allowed in the `AbsoluteSolvationProtocol` to be more specific?

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
```diff
@@ -559,26 +524,55 @@ def _validate_alchemical_components(
...
             raise ValueError(errmsg)
 
-        if len(alchemical_components['stateA']) > 1:
-            errmsg = ("More than one alchemical components is not supported "
-                      "for absolute solvation free energies")
+        # Now we check the alchemical Components
+        diff = stateA.component_diff(stateB)
+
+        # Check that there's only one state A unique Component
+        if len(diff[0]) > 1 or len(diff[0]) == 0:
```
hannahbaumann: Could this be `!= 1`?

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
IAlibay: Unless you feel strongly about this, I would prefer not, couple of reasons:
1. This entire thing gets subclassed by the ASFEProtocol in pontibus. So I expect it to be shared.
2. The error will stem from execution of the transformation, so it should (hopefully) be obvious what Protocol is emitting the error.
3. There are a lot of other warnings/errors all over the place that aren't that specific, especially if the above is untrue, that might need a rethink of how we do errors.

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
IAlibay: Ah yes thanks, looks like a holdover from when I was trying to separate it into two different error types and then gave up.

IAlibay:
> Should we adapt the ABFE and SepTop protocols as well?

The ABFE Protocol also implements `_validate` (it's why this PR exists, I was fixing things for both at the same time since they share a ton of common components).

For SepTop and HybridTop, yes eventually but not for v1.7. As-in, I don't consider not having it a blocker for SepTop, and the HybridTop one would require more effort / rewritting (validation there is messy because it was our first Protocol).

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
```diff
@@ -559,26 +524,55 @@ def _validate_alchemical_components(
           SmallMoleculeComponents.
         * Currently doesn't support more than one alchemical component
           being desolvated.
+        * Currently doesn't support charged alchemical components.
+        * Solvent must always be present in both end states.
         """
+        # Check that there are no protein components
+        if stateA.contains(ProteinComponent) or stateB.contains(ProteinComponent):
```
jthorton, line 531: Love to see the new `contains` methods out in the wild!

openfe/tests/protocols/openmm_ahfe/test_ahfe_settings.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@pytest.fixture()
+def default_settings():
+    return AbsoluteSolvationProtocol.default_settings()
+
+
+def test_create_default_settings():
+    settings = AbsoluteSolvationProtocol.default_settings()
+    assert settings
```
jthorton, line 17: Should we be checking some of these settings in case they are changed accidentally?

IAlibay: Field deletion gets checked via serialization tests (under the results check and also the tokenization ones). Field addition (if non-optional) are also checked that way.

Field changes or new defaults aren't (they used to be via testing the GufeTokenizable key, but that wasn't stable). I'm not against testing some settings, but that probably should be done else on a settings by settings manner.

openfe/protocols/openmm_afe/equil_solvation_afe_method.py
IAlibay: Both the `contains` and `diff` methods are very convenient!

PR #1579, opened 2025-10-14
run CI workflows for feat/ branches

atravitz:
In conversation with @ethanholz, we discussed running CI for branches that start with `feat/`, so that we can use CI for feature branches that might have many smaller PRs going into them.

Also, I'm proposing that we don't auto-run CI for PRs in draft-mode to save the rainforest.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,6 +13,7 @@ on:
   push:
     branches:
       - main
+      - 'feat/**'
   schedule:
     - cron: "0 4 * * *"
   workflow_dispatch:
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -3,9 +3,11 @@ on:
   pull_request:
     branches:
       - main
+      - 'feat/**'
   push:
     branches:
       - main
+      - 'feat/**'
 
 concurrency:
   group: "${{ github.workflow }}-${{ github.ref }}"


mikemhenry:
+1 to the feat rule (we need to document this somewhere) 

-1 to not running CI on draft PRs, at least without someway to running them via a label or comment. I like to use draft PRs to mean "this isn't ready yet for review" not "don't run CI on this". Sometimes I am not ready for review but I do want to see if it passes CI. But there are times when I don't want to run CI on a PR so I think in a perfect world if I could have a label that I can add to a draft PR that kicks off CI, that would be grand (and the label is a no op when the PR isn't in draft mode)

ethanholz:
I agree with Mike, running on draft PRs often helps to inform when I should request a review. If things are looking green, that's a good sign that I am actually ready to have this seriously looked at.

atravitz:
> +1 to the feat rule (we need to document this somewhere)

feat/ rule documented in this draft of contributing guidelines: https://github.com/OpenFreeEnergy/openfe/pull/1580. 
I plan to update the PR  and issue templates as well soon.

> -1 to not running CI on draft PRs, at least without someway to running them via a label or comment. I like to use draft PRs to mean "this isn't ready yet for review" not "don't run CI on this". Sometimes I am not ready for review but I do want to see if it passes CI. But there are times when I don't want to run CI on a PR so I think in a perfect world if I could have a label that I can add to a draft PR that kicks off CI, that would be grand (and the label is a no op when the PR isn't in draft mode)

This is a good point, are you saying "don't run tests on drafts by default, but add a comment or label that overrides it and turns on testing"?

.github/workflows/ci.yaml
```diff
@@ -3,15 +3,18 @@ on:
...
     branches:
       - main
+      - 'feat/**'
    # Skip CI if changed files only affect the following folders
    # - docs: documentation changes don't need code validation
    # See here for more details: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths
     paths-ignore:
       - "docs/*"
       - ".readthedocs.yaml"
+    types: [ready_for_review]  # don't run for draft PRs
```
IAlibay: See my comment in #1580, also in my opinion this discourages the use of draft PRs. If I open a draft PR, I also want to run CI to see what happens.

If we're concerned about running CI too much on draft, then we should just reduce the number of CI runners we run (we *don't* need to run codecov for every flavour of Python, and we have way too many CI runner types).

.github/workflows/ci.yaml
```diff
@@ -3,15 +3,18 @@ on:
   pull_request:
     branches:
       - main
+      - 'feat/**'
```
IAlibay: Why just feat? I often write fixes going into other PRs that would need CI to run on them.

mikemhenry:
> This is a good point, are you saying "don't run tests on drafts by default, but add a comment or label that overrides it and turns on testing"?

Yes, but only if there is some push for not wanting to run CI on drafts b/c it is an obvious waste and most of the time, the author of the PR didn't care about CI while in draft mode.

atravitz:
addressed by github.com/OpenFreeEnergy/openfe/issues/1649

PR #1590, opened 2025-10-17
Fix sampler setup for dry runs

hannahbaumann:
The sampler setup is very slow for dry runs of membrane systems.
We are using the `LocalEnergyMinimizer` to do a short minimization, but don't want to do that in the dry run. 
`minimization_steps` are set to zero in the dry run, but for the `LocalEnergyMinimizer` that leads to running minimizations until converged:
`maxIterations (int) – the maximum number of iterations to perform. If this is 0, minimation is continued until the results converge without regard to how many iterations it takes. The default value is 0.`

quick fix would be just to change it to 1, but we could also turn off minimization completely during dry runs?

### DIFF
--- a/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
+++ b/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
@@ -289,7 +289,7 @@ def minimize(thermodynamic_state: states.ThermodynamicState,
         The positions and accompanying state following minimization
     """
     # Only run a minimization if max_iterations is not None
-    if max_iterations:
+    if max_iterations is not None:
         # we won't take any steps, so use a simple integrator
         integrator = openmm.VerletIntegrator(1.0)
         platform = openmm.Platform.getPlatformByName(platform_name)


IAlibay:
Putting the None check / skip here would do fine: https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/protocols/openmm_rfe/_rfe_utils/multistate.py#L291

openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
```diff
@@ -288,20 +288,22 @@ def minimize(thermodynamic_state: states.ThermodynamicState,
...
-    platform = openmm.Platform.getPlatformByName(platform_name)
-    dummy_cache = cache.DummyContextCache(platform=platform)
-    context, integrator = dummy_cache.get_context(
-        thermodynamic_state, integrator
-    )
-    try:
-        sampler_state.apply_to_context(
-            context, ignore_velocities=True
+    # Only run a minimization if max_iterations is not None
+    if max_iterations:
```
IAlibay:
```suggestion
    if max_iterations is not None:
```
Tiny thing - usually prefered to clarify the `None` check when it's not a bool.

openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
```diff
@@ -288,20 +288,22 @@ def minimize(thermodynamic_state: states.ThermodynamicState,
...
-    integrator = openmm.VerletIntegrator(1.0)
-    platform = openmm.Platform.getPlatformByName(platform_name)
-    dummy_cache = cache.DummyContextCache(platform=platform)
-    context, integrator = dummy_cache.get_context(
-        thermodynamic_state, integrator
-    )
-    try:
-        sampler_state.apply_to_context(
-            context, ignore_velocities=True
+    # Only run a minimization if max_iterations is not None
```
IAlibay, line 291: Could you update the type hint on `minimize` to be `int | None` or `Optional[int]`?

IAlibay (changes requested):
Thanks, too small things, otherwise it looks good to me!

openfe/protocols/openmm_rfe/_rfe_utils/multistate.py
hannahbaumann: Done!

IAlibay:
Failing tests are down to AWS issues it looks.

PR #1592, opened 2025-10-17
deprecate openfe utils 3dviz in favor of gufe method

atravitz:
Checklist
* [x] Added a ``news`` entry


openfe/utils/visualization_3D.py
```diff
@@ -135,6 +136,7 @@ def view_components_3d(mols: Iterable[ExplicitMoleculeComponent],
     py3Dmol.view
         view containing all component coordinates
     """
+    warnings.warn("Use gufe.LigandAtomMapping.get_alchemical_charge_difference() instead.", DeprecationWarning)
```
IAlibay: We should have ourselves some `functools.wraps` thing that deprecates a method, that way you can just stick it on the method with minimal text and it does the rest.

PR #1611, opened 2025-10-23
Update docker image to use python 3.12 and rdkit==2025.09.1

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/test-feedstock-pkg-build.yaml
+++ b/.github/workflows/test-feedstock-pkg-build.yaml
@@ -1,0 +1,71 @@
+name: Test conda-forge package build
+
+on:
+  workflow_dispatch:
+
+jobs:
+  test-conda-build:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout openfe repository
+        uses: actions/checkout@v4
+        with:
+          path: openfe
+      
+      - name: Checkout conda-forge feedstock
+        uses: actions/checkout@v4
+        with:
+          repository: conda-forge/openfe-feedstock
+          path: openfe-feedstock
+      
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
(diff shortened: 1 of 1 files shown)


mikemhenry:
This is a quick fix, I was thinking about this yesterday that we should use a script to keep our "production" python version we use in our docker, single file installer, and lock file all in sync. One way to do that would be to use the lockfile for everything which I think would be our best bet. 

We might need 2 lock files, one that has what we recommend and then one that has some extra stuff like a Jupyter notebook for example.

mikemhenry:
https://github.com/OpenFreeEnergy/openfe/actions/runs/18752115646/job/53494281139

looks like it is passing

.github/workflows/docker.yaml
```diff
@@ -29,6 +29,45 @@ jobs:
       packages: write
 
     steps:
+      - name: Aggressive cleanup
+        run: |
```
atravitz, line 34:
```suggestion
        run: |
        # remove unneeded packages to avoid running out of memory
```

atravitz:
NOTE: related to https://github.com/OpenFreeEnergy/gufe/issues/664

mikemhenry:
it worked https://github.com/OpenFreeEnergy/openfe/actions/runs/18755688975/job/53506909380

mikemhenry:
and the containers have been pushed https://github.com/OpenFreeEnergy/openfe/pkgs/container/openfe

not sure why the "latest" tag is on the older version

IAlibay:
> not sure why the "latest" tag is on the older version

<s>@ijpulidos I'm not seeing this on my end?</s>

edit: sorry Ivan, that was a careless ping!

IAlibay:
Oops I copied in Ivan when I meant @mikemhenry ! (sorry I'm jumping between slack and github)

PR #1613, opened 2025-10-23
Test example notebooks on demand

mikemhenry:
Checklist
* [ ] Added a ``news`` entry


.github/workflows/test-example-notebooks.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Environment Information
+        run: |
+          micromamba info
+          micromamba list
+
+      - name: Run example notebooks
+        run: |
+          cd example-notebooks
+          python -m pytest -v setup/ showcase/ cookbook/ openmm_rbfe/ networks/ --nbval-lax --nbval-cell-timeout=3000 -n auto --dist loadscope
```
atravitz: if we add / reorganize the directory this test will pass in a sneaky way - do we need to be so specific here?

atravitz (changes requested):
Just One More Thing

.github/workflows/test-example-notebooks.yaml
mikemhenry: That is how we test it upstream, it is a good question but without specifying a subset of notebooks, they all get tested. Which we might want to do. Looking at the PR where I set this up, I don't see anything like "don't test this notebook" so maybe here (and upstream) we should change it to test everything?

atravitz: hm, my understanding is that we inherently _do_ want to test anything that is in the example notebooks repo. otherwise it shouldn't be there

mikemhenry:
IDK how long this takes to run, but this should be added to the review checklist since we want to make sure no PR breaks out example notebooks

PR #1629, opened 2025-10-27
pin to python < 3.14 in pyproject.toml

atravitz:
https://github.com/OpenFreeEnergy/openfe/pull/1567#discussion_r2462134045

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,8 @@ readme = "README.md"
 license = "MIT"
 license-files = [ "LICENSE" ]
 authors = [ { name = "The OpenFE developers", email = "openfe@omsf.io" } ]
-requires-python = ">=3.10"
+# blocked by OpenMM 8.4 for python 3.14 support
+requires-python = ">=3.10,<3.14"
 
 classifiers = [
   "Development Status :: 1 - Planning",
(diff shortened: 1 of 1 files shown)


mikemhenry (changes requested):
I don't think we need an upper bound on our python version support. IMHO it just makes it hard to install in a python 3.14 environment and I am not sure what the upsides are. I saw https://github.com/OpenFreeEnergy/openfe/pull/1567#discussion_r2462134045 and it actually looks like we are blocked by the `openff-*` ecosystem https://github.com/conda-forge/openff-toolkit-feedstock/blob/main/recipe/meta.yaml 

I know we didn't have an upper bound on our last release, but if we did, that means if tomorrow `openff-*` supported python 3.14, a user wouldn't be able to install `openfe` in python 3.14

tl;dr I think our requirements should be what _our_ code requires and not what functionally happens (sone dependency not supporting a python version) at some point in time.

P.S. Right now openmm 8.4 doesn't have python 3.14 builds (there is an RC) BUT if tomorrow they did, we still couldn't install our software in a python 3.14 environment b/c of `openf-*`

atravitz:
@mikemhenry - I think your opinion here follows our general best-practice, so I'm closing this PR 👍

PR #1632, opened 2025-10-27
refactor: remove result_server references in favor of gufe ExternalStorage

ethanholz:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,14 +1,15 @@
 name: "CI"
 on:
   pull_request:
-    branches:
-      - main
    # Skip CI if changed files only affect the following folders
    # - docs: documentation changes don't need code validation
    # See here for more details: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths
     paths-ignore:
       - "docs/*"
+      - "news/*"
       - ".readthedocs.yaml"
+      - ".github/workflows/cpu-long-tests.yaml"
+      - ".github/workflows/gpu-integration-tests.yaml"
   push:
     branches:
       - main
--- a/.github/workflows/clean_cache.yaml
+++ b/.github/workflows/clean_cache.yaml
@@ -1,6 +1,7 @@
 # from https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries
 name: cleanup caches by a branch
 on:
+  workflow_dispatch:
   pull_request:
     types:
       - closed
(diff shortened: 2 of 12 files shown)


ethanholz:
@atravitz I believe this is ready for review but I am unsure of how this should be marked as in the news (or if it should be at all).

openfe/storage/resultclient.py
```diff
@@ -5,7 +5,6 @@
 import re
 from typing import Any
 
-from .resultserver import ResultServer
 from .metadatastore import JSONMetadataStore
```
atravitz: ### 
```suggestion

```

This is no longer used.

atravitz (commented):
As far as news item, `ResultServer` is not in our api documentation anywhere and I don't believe it's used downstream, so if the user-facing behavior is otherwise the same, this doesn't need a news item.

Can we remove `resultserver.py` and `metadatastore.py`? ormally I would do a deprecation cycle, but if this is not advertised to our users, we can consider it an internal refactor.

I'd like sign-off from @IAlibay or @dwhswenson who know the history of the storage code and might know if this is actually used somewhere I'm not seeing.

dwhswenson:
Main point: As far as I know, this is not being used anywhere. It should be fine to remove it.

Historical context: IIRC, this was developed (originally in gufe) with the idea that things would function with the following layers:

1. Lowest layer: Writes raw bytes to some object (key-value) storage (where obvious "keys" are filenames). This still exists as `ExternalResource` in gufe.
2. Middle layer: Similar API to lowest layer, but add in metadata tracking to validate the received bytes (md5, possibly last modified). In a client-server model, this would be on the server. Hence `ResultServer`. To keep the nature of the metadata storage modular, we made it a separate component, but I think only ever implemented a `JSONMetadataStore`, which dumped the metadata to a JSON file (honestly, not a great approach).
3. User-facing layer: Focus on a more friendly API, hide the fact that we're doing all this metadata stuff. This was `ResultClient`.

At some point, these parts got moved from gufe to openfe.

My opinion now is that the middle layer was overkill, and removing it makes sense to me. If someone wants to add it as part of an `ExternalResource` backend, they can do it in their own tool; we don't need an explicit middle layer. All of this seems good to me!

ethanholz:
> Can we remove `resultserver.py` and `metadatastore.py`? ormally I would do a deprecation cycle, but if this is not advertised to our users, we can consider it an internal refactor.

I have half of this currently pending in another PR (#1646) but I am more than okay moving this into this one instead.

atravitz:
@ethanholz either (separate PR or here) is fine - I just missed it!

PR #1633, opened 2025-10-27
pin openmm to <8.3

atravitz:
pinning the development environment.yaml until we verify that https://github.com/openmm/openmm/pull/5069 is fixed.

because this is GPU-related, I see the tests fail locally but it doesn't affect our PR CPU tests.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ dependencies:
   - cinnabar ~=0.5.0
   - click >=8.2.0
   - coverage
-  - dask>=2025  # for debugging
+  - dask>=2025  # temporary fix for https://github.com/openforcefield/openff-units/issues/140
   - duecredit<0.10
   - kartograf>=1.2.0
   - konnektor~=0.2.0


environment.yml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - openff-nagl-models>=0.1.2
   - openff-toolkit-base >=0.16.2
   - openff-units==0.3.1  # https://github.com/OpenFreeEnergy/openfe/pull/1374
-  - openmm ~=8.3.0
+  - openmm ~=8.2.0  # omit 8.3.0 and 8.3.1 due to https://github.com/openmm/openmm/pull/5069, unpin once we've qualified 8.3.2
```
atravitz, line 21: I prefer this over the `!=8.1.0,>=8.0.0,<8.3.0 ` use on the feedstock ([here](https://github.com/atravitz/openfe-feedstock/blob/5e513788f3ee3e6ac57939a2337a0f9fb578426a/recipe/meta.yaml#L43-L44))
because we want to know that things are building and testing with the latest version of openmm that we can.

atravitz:
image

well that doesn't seem right

atravitz:
> image
> 
> well that doesn't seem right

ah, this is the `pip list`

atravitz:
@mikemhenry I think this is good for now, but we want to follow up with a perses  vendor and a perses feedstock pin to `dask >=2024` so that we're not pulling in a version incompatible with numpy 2

environment.yml
```diff
@@ -5,6 +5,7 @@ dependencies:
   - cinnabar ~=0.5.0
   - click >=8.2.0
   - coverage
+  - dask>=2025  # temporary fix for https://github.com/openforcefield/openff-units/issues/140
```
mikemhenry, line 8: Is it 2024 or 2025?

mikemhenry (approved):
LGTM but I had one question about the dask version

environment.yml
atravitz: i'm saying 2025 to be safe, since that's what is typically pulled in.

PR #1638, opened 2025-10-29
feat: support `openfe gather` for septop

atravitz:
soon to be followed by abfe support once we mock up a way to handle results networks

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfecli/commands/gather_septop.py
+++ b/openfecli/commands/gather_septop.py
@@ -309,8 +309,8 @@ def generate_dg_raw(results_dict: dict[str, dict[str, list]]) -> pd.DataFrame:
     default="dg",
     show_default=True,
     help=(
-        "What data to report. 'dg' gives maximum-likelihood estimate of "
-        "absolute deltaG,  'ddg' gives delta-delta-G, and 'raw' gives "
+        "What data to report. 'dg' gives the maximum-likelihood estimate derived "
+        "absolute deltaG value,  'ddg' gives delta-delta-G, and 'raw' gives "
         "the raw result of the deltaG for a leg."
     ),
 )


openfecli/commands/gather_septop.py
```diff
@@ -0,0 +1,362 @@
...
+    legs: dict[str, dict[str, list]]
+        Data extracted from the given result JSONs, organized by the leg's ligand names and simulation type.
+    """
+    from collections import defaultdict
+
+    ddgs = defaultdict(lambda: defaultdict(list))
+
+    for result_fn in result_fns:
+        names, result = _load_valid_result_json(result_fn)
+        if names is None:  # this means it couldn't find names and/or simtype
```
hannahbaumann, line 90: Maybe we should have some warning or so in this case?

openfecli/commands/gather_septop.py
```diff
@@ -0,0 +1,362 @@
...
+        Ligand names corresponding to the results.
+    """
+    # TODO: use gather's _get_names once it is improves (uses input.stateA/B.name)
+    try:
+        nm = list(result["unit_results"].values())[0]["name"]
+
+    except KeyError:
+        raise ValueError("Failed to guess names")
+
+    toks = nm.split(",")
```
hannahbaumann, line 141: This would lead to problems if there is a `,` in the ligand name. We could try to get the name from the smc instead?

openfecli/commands/gather_septop.py
atravitz: the warnings happen in `_load_valid_result_json`!

openfecli/commands/gather_septop.py
atravitz: yes, the plan is to use the smc - that'll be a follow-up PR.

atravitz:
thanks @hannahbaumann!

tests are currently failing due to marginal rounding differences - we're discussing this offline.

As far as mixed networks, that's the ultimate goal, but this PR is to get the MVP working first.

atravitz:
ready for re-review @hannahbaumann - note that this is just to be merged into a `feat` branch as the baseline that I will then make improvements to - this isn't going directly into `main`.

openfecli/commands/gather_septop.py
hannahbaumann: Depending on how quickly you're planning on making that follow up PR, one could also change this slightly to allow for ligand names that have a comma.
This is how the name is specified in the protocol:
`name=(
                        f"SepTop RBFE Setup, transformation {alchname_A} to "
                        f"{alchname_B}, {leg} leg: repeat {i} generation 0"
                    )`
`name=(
                        f"SepTop RBFE Run, transformation {alchname_A} to "
                        f"{alchname_B}, {leg} leg: repeat {i} generation 0"
                    ),`
So one could split maybe at `, transformation` and `complex` and `solvent`, similarly to the ABFE gather? 
But maybe this is not necessary with the plans of the follow up PR.

openfecli/commands/gather_septop.py
```diff
@@ -0,0 +1,359 @@
...
+    type=click.Path(dir_okay=True, file_okay=True, path_type=pathlib.Path),
+    required=True,
+)
+@click.option(
+    "--report",
+    type=HyphenAwareChoice(["dg", "ddg", "raw"], case_sensitive=False),
+    default="dg",
+    show_default=True,
+    help=(
+        "What data to report. 'dg' gives maximum-likelihood estimate of "
```
hannahbaumann:
```suggestion
        "What data to report. 'dg' gives the maximum-likelihood estimate derived "
```

openfecli/commands/gather_septop.py
```diff
@@ -0,0 +1,359 @@
...
+        "absolute deltaG,  'ddg' gives delta-delta-G, and 'raw' gives "
```
hannahbaumann:
```suggestion
        "absolute deltaG value,  'ddg' gives delta-delta-G, and 'raw' gives "
```

openfecli/tests/commands/test_gather/test_septop_full_results_raw_.csv
```diff
@@ -0,0 +1,37 @@
...
+19,solvent,1,7b,3.1000000000000001,1.3
+20,standard_state_correction_A,1,7b,-9,0
+21,standard_state_correction_A,1,7b,-9.3000000000000007,0
+22,standard_state_correction_B,1,7b,9.0999999999999996,0
+23,standard_state_correction_B,1,7b,9.3000000000000007,0
+24,complex,7a,7b,3.1699999999999999,0.80000000000000004
+25,complex,7a,7b,1.6499999999999999,0.68000000000000005
+26,complex,7a,7b,-2.0600000000000001,0.57999999999999996
+27,solvent,7a,7b,-1.2,1.3
+28,solvent,7a,7b,-0.90000000000000002,1.3
```
hannahbaumann, line 30: Would it make sense to report the same number of digits here and in the other output files?

hannahbaumann (commented):
Thanks @atravitz , this looks great! Only thing is that the outputs have lots of digits for the ddg and dgs, but I'm not sure if this is related to the rounding discussion?

atravitz:
> Thanks @atravitz , this looks great! Only thing is that the outputs have lots of digits for the ddg and dgs, but I'm not sure if this is related to the rounding discussion?

Yes this is related to the rounding discussion! This PR matches the behavior of the example notebooks as they exist today. After the [pandas PR](https://github.com/OpenFreeEnergy/openfe/pull/1690), we can unify rounding across all outputs.

openfecli/tests/commands/test_gather/test_septop_full_results_raw_.csv
atravitz: yes, see my comment addressing this below!

openfecli/commands/gather_septop.py
atravitz: Here is the follow-up PR: https://github.com/OpenFreeEnergy/openfe/pull/1691

If there's any reason that PR gets blocked, we can fall back to your suggestion.

hannahbaumann (approved):
Thanks @atravitz , lgtm! At some point it would be great to have tests for all the edge cases/failure modes, but you're probably planning on doing that some time later in a follow up/when things are more settled, so approving the PR!

PR #1640, opened 2025-10-29
Switch to using skip_if_missing

mikemhenry:
This PR unifies how we skip tests. 

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/tests/data/6CZJ.pdb
+++ b/openfe/tests/data/6CZJ.pdb
@@ -1,0 +1,2190 @@
+HEADER    DE NOVO PROTEIN                         09-APR-18   6CZJ              
+TITLE     STRUCTURE OF A REDESIGNED BETA BARREL, B10                            
+COMPND    MOL_ID: 1;                                                            
+COMPND   2 MOLECULE: B10;                                                       
+COMPND   3 CHAIN: A, B;                                                         
+COMPND   4 ENGINEERED: YES                                                      
+SOURCE    MOL_ID: 1;                                                            
+SOURCE   2 ORGANISM_SCIENTIFIC: SYNTHETIC CONSTRUCT;                            
+SOURCE   3 ORGANISM_TAXID: 32630;                                               
+SOURCE   4 EXPRESSION_SYSTEM: ESCHERICHIA COLI;                                 
+SOURCE   5 EXPRESSION_SYSTEM_TAXID: 562                                         
+KEYWDS    BETA BARREL, ROSSETTA, COMPUTATIONAL, DE NOVO, DE NOVO PROTEIN        
+EXPDTA    X-RAY DIFFRACTION                                                     
+AUTHOR    L.A.DOYLE,B.L.STODDARD                                                
+REVDAT   6   13-MAR-24 6CZJ    1       REMARK                                   
+REVDAT   5   01-JAN-20 6CZJ    1       REMARK                                   
+REVDAT   4   17-APR-19 6CZJ    1       COMPND                                   
+REVDAT   3   10-OCT-18 6CZJ    1       COMPND JRNL                              
+REVDAT   2   26-SEP-18 6CZJ    1       JRNL                                     
+REVDAT   1   19-SEP-18 6CZJ    0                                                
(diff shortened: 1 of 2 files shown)


atravitz:
I agree with unification, but I would generally prefer we use a built-in pytest functionality over an openff utils dependency. is there a benefit to openff's implementation here?

mikemhenry:
It started with me thinking that [`pytest.importskip`](https://docs.pytest.org/en/stable/deprecations.html#pytest-importorskip-default-behavior-regarding-importerror) was deprecated in 8.3, BUT a closer read is that the behavior of how it handles `ImportError` differently than `ModuleNotFoundError`

I like how the openff-utils version only skips on a `ModuleNotFoundError` and instead would let an `ImportError` bubble up. I also like how the openff-utils version is a decorator where the pytest version should really be used like this:

```python
np = pytest.importskip("numpy")

def test_cos(rad):
    assert np.cos(0) == 1
```

so we were not really using it correctly (and often our tests don't really have an import of the optinal package so it looks a bit weird to have `pytest.importskip("numpy")` inside each test we want to skip). 

I am happy to switch to using the built in pytest version, it has some nice things like a `minversion` arg so if we had a test that needs a newer version of a package but we still supported older versions, this method would just work (assuming we had 2 CI legs testing the new and old version). 


I am not really worried about the openff-utilities dep since other packages need it as well:

```
 openff-interchange-base 0.4.8   pyhd8ed1ab_1   openff-utilities >=0.1.5 conda-forge noarch
 openff-toolkit-base     0.17.1  pyhd8ed1ab_0   openff-utilities         conda-forge noarch
 openff-units            0.3.1   pyhd8ed1ab_2   openff-utilities         conda-forge noarch
```

atravitz:
Thanks for articulating this @mikemhenry  - I think the benefits (especially decorator) of openff utils is worth it here (and probably why openff added it in the first place 😄) i'm fine with this solution!

mikemhenry:
As a bonus, b/c we switch to this method, the output looks like this now:
```
$ pytest -v -k perses                                                                                                                                                       
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.13.9, pytest-8.4.2, pluggy-1.6.0 -- /home/mmh/micromamba/envs/openfe-1.7.0/bin/python3.13
cachedir: .pytest_cache
rootdir: /home/mmh/Projects/openfe
configfile: pyproject.toml
plugins: datadir-1.8.0, anyio-4.11.0, regressions-2.8.3
collected 1513 items / 1506 deselected / 7 selected                                                                                                                                          

openfe/tests/setup/atom_mapping/test_perses_atommapper.py::test_simple SKIPPED (Package perses is required, but was not found.)                                                        [ 14%]
openfe/tests/setup/atom_mapping/test_perses_atommapper.py::test_generator_length SKIPPED (Package perses is required, but was not found.)                                              [ 28%]
openfe/tests/setup/atom_mapping/test_perses_atommapper.py::test_empty_atommappings SKIPPED (Package perses is required, but was not found.)                                            [ 42%]
openfe/tests/setup/atom_mapping/test_perses_atommapper.py::test_dict_round_trip SKIPPED (Package perses is required, but was not found.)                                               [ 57%]
openfe/tests/setup/atom_mapping/test_perses_scorers.py::test_perses_normalization_not_using_positions SKIPPED (Package perses is required, but was not found.)                         [ 71%]
openfe/tests/setup/atom_mapping/test_perses_scorers.py::test_perses_not_implemented_position_using SKIPPED (Package perses is required, but was not found.)                            [ 85%]
openfe/tests/setup/atom_mapping/test_perses_scorers.py::test_perses_regression SKIPPED (Package perses is required, but was not found.)                                                [100%]
```

instead of 
```
$ pytest -v -k perses                                                                                                                                                       
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.13.9, pytest-8.4.2, pluggy-1.6.0 -- /home/mmh/micromamba/envs/openfe-1.7.0/bin/python3.13
cachedir: .pytest_cache
rootdir: /home/mmh/Projects/openfe
configfile: pyproject.toml
plugins: datadir-1.8.0, anyio-4.11.0, regressions-2.8.3
collected 1500 items / 1500 deselected / 3 skipped / 0 selected          
```

PR #1642, opened 2025-10-29
Fix import error from pr# 1640

mikemhenry:
There was one mistake in #1640, this PR fixes that mistake.
We need to wait to import things from perses directly until after we check we have the package available.  

Checklist
* [ ] Added a ``news`` entry


atravitz (approved):

mikemhenry:
Locally, the tests pass!
All the tests in `test_perses_scorers.py` are set to XPASS so they don't really run.

PR #1643, opened 2025-10-29
Fix pytest warnings from the flaky marker

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/tests/setup/atom_mapping/test_perses_scorers.py
+++ b/openfe/tests/setup/atom_mapping/test_perses_scorers.py
@@ -11,8 +11,6 @@ from openfe.setup import perses_scorers
 
 from ....utils.silence_root_logging import silence_root_logging
 
-with silence_root_logging():
-    from perses.rjmc.atom_mapping import AtomMapper, AtomMapping
 
 USING_OLD_OFF = False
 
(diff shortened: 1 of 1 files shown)


atravitz:
pre-commit.ci autofix

mikemhenry:
Still lots of warnings but no warnings from pytest anymore!

pyproject.toml
```diff
@@ -68,6 +68,11 @@ lint.ignore = [
 ]
 lint.isort.known-first-party = [ "openfe" ]
 
+[tool.pytest.ini_options]
+markers = [
+  "flaky: marks tests that need to be re-tried sometimes because of iterment errors (such as http errors)",
```
atravitz, line 73:
```suggestion
  "flaky: marks tests that need to be re-tried sometimes because of internet errors (such as http errors)",
```

mikemhenry:
Yes and IDK why the auto register isn't working actually https://github.com/pytest-dev/pytest-rerunfailures/blob/b3220b9/src%2Fpytest_rerunfailures.py#L324-L330

pyproject.toml
mikemhenry: Ha! I meant to spell intermittent

mikemhenry:
Okay I figured it out! The problem is we don't have `pytest-rerunfailures` as a dep in the feedstock BUT when it is installed, this PR isn't needed.

So I don't think we actually want this, since that warning is a hint that `pytest-rerunfailures` isn't installed, and this PR covers up that error. 

Instead we should add the dependency to the feedstock.

mikemhenry:
Fixed here https://github.com/conda-forge/openfe-feedstock/pull/108

PR #1644, opened 2025-10-30
add workflow_dispatch event trigger for cleaning up cache by branch

atravitz:
I'd like to have this for when we know we need to clear a cache to address test failures on a specific PR.

Checklist
* [ ] Added a ``news`` entry


mikemhenry:
Sounds good, also makes it easy to test it if we need to make changes, BTW, this will also let you clear the chache as well:
https://github.com/OpenFreeEnergy/openfe/actions/caches

PR #1647, opened 2025-10-30
Create Pooch Cache

mikemhenry:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -123,7 +123,7 @@ jobs:
           path: |
             # linux cache location
             ~/.cache/openfe
-            # osx cache locatio:
+            # osx cache location
             ~/Library/Caches/openfe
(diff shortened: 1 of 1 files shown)


mikemhenry:
We will need to make sure the cache clearing action doesn't end up deleting the pooch cache, but since it should be able to restore the cache from `main` it should be okay. I will keep an eye on it.

mikemhenry:
I tested this by running a test that uses pooch and cranking the log level up so I could see if things were downloaded and when there was a cache hit it didn't download the files and the tests passed so it worked!!

.github/workflows/clean_cache.yaml
mikemhenry, line 1: This update is mostly b/c the `gh` command now has built in cache support

atravitz (approved):

PR #1649, opened 2025-10-30
allow ci to run on all branches, not just main

atravitz:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,7 +1,6 @@
 name: "CI"
 on:
   pull_request:
-    branches:
    # Skip CI if changed files only affect the following folders
    # - docs: documentation changes don't need code validation
    # See here for more details: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths


.github/workflows/ci.yaml
```diff
@@ -2,12 +2,12 @@ name: "CI"
 on:
   pull_request:
     branches:
-      - main
    # Skip CI if changed files only affect the following folders
    # - docs: documentation changes don't need code validation
    # See here for more details: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths
     paths-ignore:
       - "docs/*"
+      - "news/*"
```
atravitz, line 9: this seems practical

.github/workflows/ci.yaml
```diff
@@ -2,12 +2,12 @@ name: "CI"
 on:
   pull_request:
```
atravitz, line 3: also I updated this to be more strict ( it was "Require approval for first-time contributors" before): 

Image

and then we can use: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks

atravitz:
@mikemhenry please take a look after this is merged.

.github/workflows/ci.yaml
mikemhenry: That's fine, I figured a trust on first use model (like the one we use for `ssh`) was okay, but I don't think we can be too careful. It does mean we can be quicker to let someone merge in a one line typo change since it then doesn't get them into the "run without approval" list

PR #1660, opened 2025-11-05
refactor: rename to WarehouseBaseClass

### LINKED ISSUE #1648: Rename ResultClient (aka Warehouse) to something more indicative of its purpose
This is in reference to a discussion between myself and @atravitz on whether Warehouse is the name that OpenFE wants to stick with when it comes to storage. Personally, I am okay with keeping it this way but wanted to ensure that we have a place for discussion.

Per the specification naming will be of the following nature for classes:
- `<Name>Base` or just `<Name>` for base classes
- `Default<Name>` or `FileSystem<Name>` for a file implementation class

ethanholz:
Closes #1648. 

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/storage/warehouse.py
+++ b/openfe/storage/warehouse.py
@@ -18,7 +18,7 @@ from gufe.storage.externalresource import ExternalStorage
 GUFEKEY_JSON_REGEX = re.compile('":gufe-key:": "(?P<token>[A-Za-z0-9_]+-[0-9a-f]+)"')
 
 
-class _Container(abc.ABC):
+class _DataContainer(abc.ABC):
     """
     Abstract class, represents all data under some level of the heirarchy.
     """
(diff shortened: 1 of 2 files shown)


openfe/storage/warehouse.py
```diff
@@ -18,7 +18,7 @@
 GUFEKEY_JSON_REGEX = re.compile('":gufe-key:": "(?P<token>[A-Za-z0-9_]+-[0-9a-f]+)"')
 
 
-class _ResultContainer(abc.ABC):
+class _Container(abc.ABC):
```
atravitz:
```suggestion
class _DataContainer(abc.ABC):
```
i'm being pedantic, but I know that we talked about "Container" being confusing because it has strong associations with Containerization. Does `DataContainer` make more sense?

ethanholz: I think that is a good option for us here. So I will go ahead and apply this across the board for places this is used.

atravitz (approved):

Issue #1582, opened 2025-10-15
feat: add a unified storage system for setup, result, and task storage


ethanholz:
This issue will be used to track the work done for warehouse which will contain any subissues for various integration steps.

ethanholz:
Important to note that I can't add sub-issues since I am not affiliated with the project. For more info see [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/adding-sub-issues).

For this request I would like to request a long-lived feature branch be added to this repo for my work that will be done. Going to tag @atravitz on this for completeness.

IAlibay:
@ethanholz I've added you to the org.

mikemhenry:
its_happening.gif

ethanholz:
It looks like I still can't create subissues. Should I just create them separately in the meantime and reference this issue in them? Pinging @atravitz for this since this was something her and I discussed regarding contributing guidelines.

atravitz:
@ethanholz you should have triage access now so that you can make sub-issues! also, all of this work can be merged into [`feat/warehouse`](https://github.com/OpenFreeEnergy/openfe/tree/feat/warehouse)

PR #1671, opened 2025-11-12
feat/refactor: add setup store

ethanholz:
Related to #1582.

This PR aims to add a setup store warehouse. In doing so, I ended up simplifying the base API since it was redundant abstraction. In this code, we remove the `_DataContainer` class since it is was largely used to handle cases where we wanted to mirror a filesystem API. However, we have discussed migrate to a more key-value-based approach. This refactor enables while making this codepath extensible in the future. 

I have included the start of the `FilesSystemWarehouse` to showcase how someone would use the Warehouse class. Next steps for implementing into `quickrun` will come in future issues. This code is ready to review but will probably need revision around naming and standards (and I anticipate that there will need to be some code coverage changes but I have not been able to run that locally without running all OpenFE tests). 

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfe/storage/warehouse.py
+++ b/openfe/storage/warehouse.py
@@ -106,7 +106,7 @@ class WarehouseBaseClass:
         """
         return self._load_gufe_tokenizable(gufe_key=obj)
 
-    def exists(self, key: GufeKey):
+    def exists(self, key: GufeKey) -> bool:
         """Check if an object with the given key exists in any store.
 
         Parameters


ethanholz:
Converting back to draft because I still have some failures.

openfe/storage/warehouse.py
```diff
@@ -3,171 +3,95 @@
...
         # probably should include repr of external store, too
-        return f"{self.__class__.__name__}({self.path})"
-
+        return f"{self.__class__.__name__}({self.stores})"
 
-class WarehouseBaseClass(_DataContainer):
-    def __init__(self, external_storage: ExternalStorage):
-        self._external_storage = external_storage
-        super().__init__(parent=self, path_component=None)
+    def delete(self, store_name: str, location: str):
```
atravitz:
```suggestion
    def delete(self, store_name: Literal["setup"], location: str):
```

`store_name` should be a `Literal` yes? Also, it's unclear to me what `location` should be - it looks to be a gufe key where I see it used.

openfe/storage/warehouse.py
```diff
@@ -3,171 +3,95 @@
...
-            object)
-        """
-        pref = prefix.split("/")  # remove this if we switch to tuples
-        cls, token = key.split("-")
-        tup = tuple(list(pref) + [cls, f"{token}.json"])
-        # right now we're using strings, but we've talked about switching
-        # that to tuples
-        return "/".join(tup)
+    def _get_store_for_key(self, key: GufeKey) -> ExternalStorage:
+        """Funciton to find the store in which a gufe key is stored in."""
```
atravitz:
```suggestion
        """Function to find the store in which a gufe key is stored in."""
```

openfe/storage/warehouse.py
ethanholz: That is correct. Much of this code is a bit strange because initially many parameters were untyped. I think that what we really want here is to explicitly state that we are accepting a `GufeKey` as the thing to be deleted. The problem is that as it exists, `delete` in `gufe.ExternalStorage` is actually accepting a `str`. My largest concern is that result objects might _not_ be of type `GufeKey`. For the time being, I am willing to change it and then update if we need to when we get there.

ethanholz:
This code now uses keyed_chain to store information instead of leveraging existing solutions. This is because keyed_chain information is portable. While our existing implementation, did actually work and mirrored the existing `gufe` documentation, using a keyed chain reduces the total amount of code reuse.

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
+    def delete(self, store_name: Literal["setup"], location: str):
+        """Delete an object from a specific store.
 
-    def __iter__(self):
-        for loc in self.external_storage:
-            if loc.startswith(self.path):
-                yield loc
+        Parameters
+        ----------
+        store_name : str
```
atravitz:
```suggestion
        store_name : Literal["setup"]
```

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
-        Parmeters
-        ---------
-        transformation: :class:`.Transformation`
-            the transformation to store
+        for name in self.stores:
+            if key in self.stores[name]:
+                return self.stores[name]
+        raise ValueError(f"GufeKey {key} is not stored")
+
+    def _store_gufe_tokenizable(self, store_name: str, obj: GufeTokenizable):
```
atravitz:
```suggestion
    def _store_gufe_tokenizable(self, store_name: Literal["setup"], obj: GufeTokenizable):
```

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
+        """Store a GufeTokenizable object with deduplication.
+
+        Parameters
+        ----------
+        store_name : str
```
atravitz:
```suggestion
        store_name : Literal["setup"]
```

openfe/storage/warehouse.py
```diff
@@ -182,103 +238,73 @@ def recursive_build_object_cache(gufe_key):
...
-
+            #
             registry[obj.key] = obj
             return obj
 
         return recursive_build_object_cache(gufe_key)
 
-    def load_transformation(self, key: str):
-        """Load a :class:`.Transformation` from its GufeKey
+    def _load_stream(self, store_name: str):
```
atravitz: this isn't used yet - are you anticipating using it later? I see you call `store.load_stream(key)` directly when loading gufe tokenizables.

openfe/storage/warehouse.py
```diff
@@ -182,103 +238,73 @@ def recursive_build_object_cache(gufe_key):
...
+        """Load a stream from the specified store.
 
         Parameters
         ----------
-        key: str
-            the gufe key for this object
+        store_name : str
```
atravitz:
```suggestion
        store_name : Literal["setup"]
```

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
+            Name of the store to delete from.
+        location : str
+            Location/path of the object to delete.
 
-    def load_stream(self, location):
-        return self.external_storage.load_stream(location)
+        Returns
+        -------
+        bool
+            True if deletion was successful, False otherwise.
```
atravitz: is this true? It looks like it doesn't return anything, but rather raises an error if it's not successful:
https://github.com/OpenFreeEnergy/gufe/blob/5c281ff4f611a48e2b6791c168bf8fe21431273c/gufe/storage/externalresource/base.py#L105

am I missing some intermediate handling?

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
+        ----------
+        obj : GufeTokenizable
+            The object to store.
+        """
+        self._store_gufe_tokenizable("setup", obj)
 
-    @property
-    def path(self):
-        return self.parent.path + "/" + self._path_component
+    def load_setup_tokenizable(self, obj: GufeKey):
```
atravitz:
```suggestion
    def load_setup_tokenizable(self, obj: GufeKey) -> GufeTokenizable:
```
I find return type hints very helpful.

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
-        # probably should include repr of external store, too
-        return f"{self.__class__.__name__}({self.path})"
+        Returns
+        -------
+        GufeTokenizable
+            The loaded object.
+        """
+        return self._load_gufe_tokenizable(gufe_key=obj)
 
+    def exists(self, key: GufeKey):
```
atravitz:
```suggestion
    def exists(self, key: GufeKey) -> bool:
```

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
-            the network to store
+        Returns
+        -------
+        bool
+            True if the key exists in any store, False otherwise.
         """
-        self._store_gufe_tokenizable("setup", network)
+        return any(key in store for store in self.stores.values())
+
+    # TODO: Fix this to be a little more concise
```
atravitz: clean up will be a follow-up PR?

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
+        Notes
+        -----
+        Uses depth-first search to rebuild object hierarchy and ensure
+        proper deduplication in memory.
+        """
         registry = {}
 
-        def recursive_build_object_cache(gufe_key):
-            """DFS to rebuild object heirarchy"""
+        def recursive_build_object_cache(key: GufeKey):
```
atravitz:
```suggestion
        def recursive_build_object_cache(key: GufeKey) -> GufeTokenizable:
```

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
+        Returns
+        -------
+        bool
+            True if the key exists in any store, False otherwise.
         """
-        self._store_gufe_tokenizable("setup", network)
+        return any(key in store for store in self.stores.values())
+
+    # TODO: Fix this to be a little more concise
+    def _load_gufe_tokenizable(self, gufe_key: GufeKey):
```
atravitz:
```suggestion
    def _load_gufe_tokenizable(self, gufe_key: GufeKey) -> GufeTokenizable:
```

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
+            The key to check for existence.
 
-    def delete(self, location: str):
-        self._external_storage.delete(location)
+        Returns
+        -------
+        bool
+            True if the object exists, False otherwise.
+        """
+        return self._key_exists(key)
```
atravitz: what's the motivation for having this as a private method instead of just having `exists`?

openfe/storage/warehouse.py
```diff
@@ -3,172 +3,228 @@
...
-
-    def store_transformation(self, transformation):
-        """Store a :class:`.Transformation`.
-
-        Parmeters
-        ---------
-        transformation: :class:`.Transformation`
-            the transformation to store
+        for name in self.stores:
+            if key in self.stores[name]:
```
atravitz, line 143: I know keys are deduplicated _within_ a given store, but is it possible for a given key to exist in multiple different stores?

atravitz (changes requested):

openfe/storage/warehouse.py
ethanholz: Honestly, no clue. Lets just move it all into `exists`.

openfe/storage/warehouse.py
ethanholz: Okay, so keys _should_ be deduplicated across stores to reduce space on disk. However, I see that we _aren't_ currently deduplicating across stores on insertion into the store. I can add that, good catch.

openfe/storage/warehouse.py
ethanholz: Actually, this is just a holdover and should be removed. The verbosity of this section should be kept, we can remove this TODO.

openfe/storage/warehouse.py
ethanholz: No, this was a consideration in an existing draft that I removed, throwing an error on deletion might be the way to go and to have this represented.

openfe/storage/warehouse.py
ethanholz: I think it might be in order to remove this. I hadn't quite determined if this was the way to go just yet, but I think removing and relying on `gufe` is better.

ethanholz:
@atravitz, this is in a good spot for re-review and merge if it looks agreeable!

PR #1686, opened 2025-11-19
feat: add abfe gathering support first draft

atravitz:
First pass at adding an MVP for supporting abfe results gathering, based on https://docs.openfree.energy/en/latest/tutorials/abfe_analysis_tutorial.html

Note that this is targeting `feat/cli-gather-improvements`, not `main`, and so will get a lot of code polish after this initial PR.

Checklist
* [ ] Added a ``news`` entry (I'll add a news entry once this is more polished)

### DIFF
--- a/openfecli/commands/gather_abfe.py
+++ b/openfecli/commands/gather_abfe.py
@@ -5,7 +5,6 @@ from typing import List, Literal
 import click
 import numpy as np
 import pandas as pd
-from cinnabar import FEMap, Measurement
 from openff.units import unit
 
 from openfecli import OFECommandPlugin


openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+    ----------
+    result_fns : list[pathlib.Path]
+        List of filepaths containing results formatted as JSON.
+    report : Literal["dg", "ddg", "raw"]
+        Type of report to generate.
+
+    Returns
+    -------
+    legs: dict[str, dict[str, list]]
+        Data extracted from the given result JSONs, organized by the leg's ligand names and simulation type.
```
hannahbaumann:
```suggestion
        Data extracted from the given result JSONs, organized by the leg's ligand name and simulation type.
```

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+    from collections import defaultdict
+
+    ddgs = defaultdict(lambda: defaultdict(list))
+
+    for result_fn in result_fns:
+        names, result = _load_valid_result_json(result_fn)
```
hannahbaumann: Technically, in ABFE's it's just one name, but may not matter. Just maybe to think about how to call this for the combined version some time later to not be confusing.

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+        ]  # fmt: skip
+        for p in proto_key:
+            if "unit_estimate" in result["unit_results"][p]["outputs"]:
+                simtype = result["unit_results"][p]["outputs"]["simtype"]
+                dg = result["unit_results"][p]["outputs"]["unit_estimate"]
+                dg_error = result["unit_results"][p]["outputs"]["unit_estimate_error"]
+
+                ddgs[names][simtype].append([dg, dg_error])
+            elif "standard_state_correction_A" in result["unit_results"][p]["outputs"]:
+                corr_A = result["unit_results"][p]["outputs"]["standard_state_correction_A"]
```
hannahbaumann: ABFE's only have one standard state correction, `"standard_state_correction"` since it's only one ligand. So it shouldn't be able to find these two keys `"standard_state_correction_A"` and `"standard_state_correction_B"`

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+        Type of report to generate.
+
+    Returns
+    -------
+    legs: dict[str, dict[str, list]]
+        Data extracted from the given result JSONs, organized by the leg's ligand names and simulation type.
+    """
+    from collections import defaultdict
+
+    ddgs = defaultdict(lambda: defaultdict(list))
```
hannahbaumann: Technically, these are dgs and not ddgs.

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+    complex_errors = [x[1].m for x in r["complex"]]
+    solvent_errors = [x[1].m for x in r["solvent"]]
+    return np.sqrt(np.mean(complex_errors) ** 2 + np.mean(solvent_errors) ** 2)
+
+
+def extract_results_dict(
+    results_files: list[os.PathLike | str],
+) -> dict[str, dict[str, list]]:
+    """
+    Get a dictionary of abfe results from a list of directories.
```
hannahbaumann:
```suggestion
    Get a dictionary of ABFE results from a list of directories.
```

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+
+    Parameters
+    ----------
+    results_files : list[ps.PathLike | str]
+        A list of directors with abfe result files to process.
```
hannahbaumann:
```suggestion
        A list of directors with ABFE result files to process.
```

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+    help=(
+        "Results that are output to stdout will be formatted as tab-separated, "
+        "identical to the formatting used when writing to file."
+        "By default, the output table will be formatted for human-readability."
+    ),
+)
+def gather_abfe(
+    results: List[os.PathLike | str],
+    output: os.PathLike | str,
+    report: Literal["dg", "ddg", "raw"],
```
hannahbaumann:
```suggestion
    report: Literal["dg", "raw"],
```

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+    tsv: bool,
+):
+    ddgs = extract_results_dict(results)
```
hannahbaumann: This would be dgs, but if code wise it would be better to be consistent with the other protocols, maybe ddgs is also fine, though might be a bit confusing for some, so maybe a comment here would be good?

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,315 @@
...
+            if k.startswith("ProtocolUnitResult")
+        ]  # fmt: skip
+        for p in proto_key:
+            if "unit_estimate" in result["unit_results"][p]["outputs"]:
+                simtype = result["unit_results"][p]["outputs"]["simtype"]
+                dg = result["unit_results"][p]["outputs"]["unit_estimate"]
+                dg_error = result["unit_results"][p]["outputs"]["unit_estimate_error"]
+
+                ddgs[names][simtype].append([dg, dg_error])
+            elif "standard_state_correction_A" in result["unit_results"][p]["outputs"]:
```
hannahbaumann: In ABFEs this may need to be an `if`. In the SepTop protocol we have a setup and a run unit, and the standard state correction is stored in the setup unit result, while the simulation results are stored in the run unit result. But in ABFEs, both are stored in the run unit, but only the run unit of the complex has the standard_state_correction.

openfecli/tests/commands/test_gather/test_abfe_full_results_raw_.csv
```diff
@@ -0,0 +1,7 @@
+,leg,ligand,DG (kcal/mol),uncertainty (kcal/mol)
+0,complex,1,36.869999999999997,0.35999999999999999
+1,complex,1,39.240000000000002,0.44
+2,complex,1,37.460000000000001,0.47999999999999998
+3,solvent,1,10.57,0.66000000000000003
+4,solvent,1,10.48,0.65000000000000002
+5,solvent,1,10.41,0.66000000000000003
```
hannahbaumann, line 7: The standard state correction seems to be missing here?

openfecli/commands/gather_abfe.py
atravitz: yes, i'll consider this for the refactor!

openfecli/commands/gather_abfe.py
atravitz: ah, thank you! I revisited the example notebook and made fixes accordingly (also for the comment below)

openfecli/commands/gather_abfe.py
atravitz: fixed! see above comment

openfecli/commands/gather_abfe.py
atravitz: thanks for catching this -i'll rename to be accurate!

openfecli/commands/gather_abfe.py
atravitz: fixed, along with other discrepancies I missed here.

openfecli/tests/commands/test_gather/test_abfe_full_results_raw_.csv
atravitz: I thought we considered excluding them at one point, but I'll at least include them here for this PR. and it's probably more rigorous to include them in a `raw` output.

hannahbaumann (approved):
Thanks @atravitz , lgtm!

PR #1690, opened 2025-11-20
Refactor/gather use pandas

atravitz:
The addition of this function will allow us to separate calculations from formatting - we will be able to consistently format results just before writing to output, rather than formatting within the actual calculations themselves.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/openfecli/commands/gather.py
+++ b/openfecli/commands/gather.py
@@ -111,15 +111,15 @@ def format_df_with_precision(
     >>> df
             ligand_i    ligand_j DDG(i->j) (kcal/mol) uncertainty (kcal/mol)
         0  lig_ejm_31  lig_ejm_42                Error                  Error
-        1  lig_ejm_31  lig_ejm_46                -0.89                   0.06
-        2  lig_ejm_31  lig_ejm_47                  0.0                    0.1
-        3  lig_ejm_31  lig_ejm_48                 0.61                   0.09
-        4  lig_ejm_31  lig_ejm_50                 1.00                   0.04
-        5  lig_ejm_42  lig_ejm_43                  1.4                    0.2
-        6  lig_ejm_46  lig_jmc_23                 0.29                   0.09
-        7  lig_ejm_46  lig_jmc_27                 -0.1                    0.1
+        1  lig_ejm_31  lig_ejm_46            -0.891077               0.064825
+        2  lig_ejm_31  lig_ejm_47             0.023341               0.145625
+        3  lig_ejm_31  lig_ejm_48             0.614103               0.088704
+        4  lig_ejm_31  lig_ejm_50             0.999904               0.044457
+        5  lig_ejm_42  lig_ejm_43             1.354348               0.156009
+        6  lig_ejm_46  lig_jmc_23             0.294761               0.086632
+        7  lig_ejm_46  lig_jmc_27            -0.101737               0.100997
         8  lig_ejm_46  lig_jmc_28                Error                  Error
-    >>> df_formatted = format_df_with_precision(df, "DG(i->j) (kcal/mol)", "uncertainty (kcal/mol)")
+    >>> df_out = format_df_with_precision(df, "DDG(i->j) (kcal/mol)", "uncertainty (kcal/mol)")
     >>> df_formatted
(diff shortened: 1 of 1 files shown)


openfecli/commands/gather.py
```diff
@@ -80,6 +80,79 @@ def format_estimate_uncertainty(
...
+            ligand_i    ligand_j DDG(i->j) (kcal/mol) uncertainty (kcal/mol)
+        0  lig_ejm_31  lig_ejm_42                Error                  Error
+        1  lig_ejm_31  lig_ejm_46                -0.89                   0.06
+        2  lig_ejm_31  lig_ejm_47                  0.0                    0.1
+        3  lig_ejm_31  lig_ejm_48                 0.61                   0.09
+        4  lig_ejm_31  lig_ejm_50                 1.00                   0.04
+        5  lig_ejm_42  lig_ejm_43                  1.4                    0.2
+        6  lig_ejm_46  lig_jmc_23                 0.29                   0.09
+        7  lig_ejm_46  lig_jmc_27                 -0.1                    0.1
+        8  lig_ejm_46  lig_jmc_28                Error                  Error
```
mikemhenry, line 121: I think this example is missing all the extra digits that `format_df_with_precision` is supposed to fix

openfecli/commands/gather.py
```diff
@@ -80,6 +80,79 @@ def format_estimate_uncertainty(
...
+
+    """
+    # we only want to round/format the floats (not any error strings getting passed through)
+    float_mask = df[est_col_name].apply(lambda x: isinstance(x, float))
```
mikemhenry: Do we also need to worry about `int` here? Would it be better to make a `string_mask` and then invert that selection so floats and ints get captured?

openfecli/commands/gather.py
mikemhenry, line 1: Surprised we didn't need to change any tests for this refactor

openfecli/commands/gather.py
```diff
@@ -316,10 +389,14 @@ def _get_ddgs(legs: dict, allow_partial=False) -> None:
...
+    df_ddg = pd.DataFrame(
+        DDGs,
+        columns=["ligand_i", "ligand_j", "DDG_bind", "bind_unc", "DDG_hyd", "hyd_unc"],
+    )
+    return df_ddg
 
 
-def _generate_ddg(legs: dict, allow_partial: bool) -> None:
+def _generate_ddg(legs: dict, allow_partial: bool) -> pd.DataFrame:
     """Compute and write out DDG values for the given legs.
```
mikemhenry, line 406: This doc string needs to be updated

mikemhenry (changes requested):
This is pretty clear! Just one spot where I think the doc strings need to be updated and a few notes

openfecli/commands/gather.py
atravitz: that was the goal! I can definitely write some specific unit tests for this functionality, but I think that's more in the scope of "make gather api public", since that's when we'll decide what points are user-facing.

openfecli/commands/gather.py
atravitz: oof thanks for catching - copy paste mistake by me

openfecli/commands/gather.py
atravitz: You're right, this will be more robust.  I updated it to filter by string, and also check for strings in *either* column to format.

openfecli/commands/gather.py
atravitz: does it? this function still computes and writes out the values because it calls `_get_ddgs` within it. or do you mean something else that I'm missing?

mikemhenry: I thought writes out implied writing and not returning but I missed that call within the function

openfecli/commands/gather.py
mikemhenry: Sounds good! I think it means we good integration tests but could improve our unit testing

mikemhenry (approved):

openfecli/commands/gather.py
atravitz: oh you're right - it formats but doesn't output. I'll revise this and related docstrings in further clean up.

openfecli/commands/gather.py
atravitz: exactly

Issue #1689, opened 2025-11-20
Issue not printing the proper name in the result.json


Oulfin:
Hi,

Very nice work with openFE is looking very good.

I am trying to generate inputs via notebook to run a RelativeHybridTopologyProtocol, when I run the simulation everything works fine but it does not get the names of the stateA and stateB. I get in the results.json (complex and solvent) with the variable name like below:

{"name": " to  repeat 0 generation 0",

I am just following the tutorial with my own sdf and protein pdb files.

This is how it look after I type the transformation_solvent, I get this information:

Transformation(stateA=ChemicalSystem(name=Compound1, components={'ligand': SmallMoleculeComponent(name=Compound1), 'solvent': SolventComponent(name=O, Na+, Cl-)}), stateB=ChemicalSystem(name=Compound2, components={'ligand': SmallMoleculeComponent(name=Compound2), 'solvent': SolventComponent(name=O, Na+, Cl-)}), protocol=<RelativeHybridTopologyProtocol-xxxxxxxxxxxx>, name=Compound1_Compound2_solvent)

Then I do this

from openfe.protocols.openmm_rfe import RelativeHybridTopologyProtocol
complex_dag = transformation_complex.create()
solvent_dag = transformation_solvent.create()

import pathlib
# first we create the directory
transformation_dir = pathlib.Path("Test")
transformation_dir.mkdir(exist_ok=True)
# then we write out the transformations
transformation_complex.to_json(transformation_dir / f"{transformation_complex.name}.json")
transformation_solvent.to_json(transformation_dir / f"{transformation_solvent.name}.json")

After I do the quickrun and gather I get the results, but with the wrong name due to the issue I wrote above.

ligand_i        ligand_j        DDG(i->j) (kcal/mol)    uncertainty (kcal/mol)
to      0       -1.4    0.0

I am using the last version of OpenFE 1.7 and Ubuntu 20.04.2 LTS.

I use the same files previously in older versions and I did not have any issue with the names.

atravitz:
Thanks for reporting this @Oulfin! A couple questions to help us debug:

1. When you say this worked in older versions, do you know which version you were running with before - was it openfe v1.6.1?
2. Are your names actually `Compound1` and `Compound2`, or are you using compound names with special characters in them (such as commas, hyphens, etc.)?
3. Could you share the output of `complex_dag.protocol_units[0].name` and `complex_dag.protocol_units[0].inputs`? With any propriety information redacted, of course.

[...]

PR #1691, opened 2025-11-20
Feat/improve names gathering

atravitz:
May fix https://github.com/OpenFreeEnergy/openfe/issues/1689, but we've been meaning to do this for a while anyway. 

This PR makes use of gufe `ChemicalSystem`s and the `contains` method (thanks @jthorton!) for gathering.
This should be much more robust than the string split and key name check we were doing before.

I suppose this has the potential to break some edge cases, but I think it also fixes some potential bugs.

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/news/gather_by_gufe.rst
+++ b/news/gather_by_gufe.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* ``openfe gather`` is now more rigorous in extracting ligand names and run types. These are now determined directly from component attributes, rather than relying on naming conventions. See `PR #1691 <https://github.com/OpenFreeEnergy/openfe/pull/1691>`_ for more details.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>


atravitz:
> LGTM! Much cleaner now, nice we didn't have to change any tests

tests just failed 😄  taking a look

openfecli/tests/commands/test_gather.py
```diff
@@ -68,283 +68,61 @@ def test_get_column(val, col):
...
+        with mock.patch("openfecli.commands.gather.load_json", return_value=min_result_json):
             result = _load_valid_result_json(fpath="")
             captured = capsys.readouterr()
-            assert result == ((("lig_ejm_31", "lig_ejm_42"), "solvent"), sim_result)
+            assert result == ((("lig_ejm_31", "lig_ejm_42"), "solvent"), min_result_json)
             assert captured.err == ""
 
-    def test_skip_missing_unit_result(self, capsys, sim_result):
-        sim_result["unit_results"] = {}
+    def test_skip_missing_unit_result(self, capsys, min_result_json):
```
atravitz, line 78: this is changed because before we were pulling the names from the `unit_results`. Now we pull directly from the gufe objects. 

Based on how `unit_results` are generated in quickrun ([here](https://github.com/OpenFreeEnergy/openfe/blob/fa30a6ba925d81f293a12a56851b29ce184e7a19/openfecli/commands/quickrun.py#L121)), I think it's appropriate to test for an empty dict, rather than missing the `unit_results` key.

openfecli/commands/gather.py
```diff
@@ -121,31 +122,28 @@ def _get_names(result: dict) -> tuple[str, str]:
...
-    try:
-        nm = list(result["unit_results"].values())[0]["name"]
-
-    except KeyError:
-        raise ValueError("Failed to guess names")
 
-    # TODO: make this more robust by pulling names from inputs.state[A/B].name
+    # TODO: I don't like this [0][0] indexing, but I can't think of a better way currently
+    protocol_data = list(result["protocol_result"]["data"].values())[0][0]
+    ligand_A = gufe.SmallMoleculeComponent.from_dict(
```
jthorton, line 207: Nice I like the fact we are building the objects now and using the API! Is there a reason we don't do that at a higher level in gather though like trying to load the protocolresult as the corresponding object and then working with the API throughout rather than pulling the dicts apart?

openfecli/commands/gather.py
```diff
@@ -121,31 +122,28 @@ def _get_names(result: dict) -> tuple[str, str]:
...
-    if toks[2] == "repeat":
-        return toks[0], toks[1]
-    else:
-        return toks[0], toks[2]
+    return ligand_A.name, ligand_B.name
 
 
-def _get_type(res: dict) -> Literal["vacuum", "solvent", "complex"]:
+def _get_type(result: dict) -> Literal["vacuum", "solvent", "complex"]:
     """Determine the simulation type based on the component names."""
```
jthorton:
```suggestion
    """Determine the simulation type based on the component types."""
```

openfecli/commands/gather.py
```diff
@@ -121,31 +122,28 @@ def _get_names(result: dict) -> tuple[str, str]:
...
-    # TODO: use component *types* instead here
-    list_of_pur = list(res["protocol_result"]["data"].values())[0]
-    pur = list_of_pur[0]
-    components = pur["inputs"]["stateA"]["components"]
 
-    if "solvent" not in components:
+    protocol_data = list(result["protocol_result"]["data"].values())[0][0]
+    chem_sys_A = gufe.tokenization.GufeTokenizable.from_dict(protocol_data["inputs"]["stateA"])
```
jthorton: [nit] Why not use the ChemicalSystem object directly?

openfecli/commands/gather.py
```diff
@@ -121,31 +122,28 @@ def _get_names(result: dict) -> tuple[str, str]:
...
+
+    if not chem_sys_A.contains(gufe.SolventComponent):
         return "vacuum"
-    elif "protein" in components:
+    elif chem_sys_A.contains(gufe.ProteinComponent):
```
jthorton, line 225: Yeeeey!

openfecli/commands/gather.py
atravitz: I think we want to move in that direction, but this is the minimally-invasive fix for now!

openfecli/commands/gather.py
atravitz: ah, we can! that was an artifact from me trying some different things during development. 
fixed!

PR #1694, opened 2025-11-24
Disable JAX acceleration by default

mikemhenry:
Checklist
* [ ] Ran pre-commit by making a comment with `pre-commit.ci autofix` before requesting review.

### DIFF
--- a/news/jax-warning.rst
+++ b/news/jax-warning.rst
@@ -3,6 +3,8 @@
 * Emit a clarifying log message when a user gets a warning from JAX (#1585).
   Fixes #1499.
 
+* Disable JAX acceleration by default, see https://docs.openfree.energy/en/latest/guide/troubleshooting.html#pymbar-disable-jax for more information (#1694).
+
 **Changed:**
 
 * <news item>


mikemhenry:
TODO: This needs a news item!

openfe/__init__.py
```diff
@@ -1,3 +1,12 @@
+# Before we do anything else, we want to disable JAX
+# acceleration by default but if a user has set
+# PYMBAR_DISABLE_JAX to some value, we want to keep
+# it
+
+import os
+
+os.environ.setdefault("PYMBAR_DISABLE_JAX", "TRUE")
```
IAlibay, line 17: Would it be safer to place this in the execute_DAG call or something?

atravitz: I think I prefer to have environment variables set at the highest level like this - it's less sneaky IMHO

IAlibay: The issue is that if you just have openfe imported you'll just disable Jax in pymbar. That can lead to cases where folks were successfully using it in a script where they might be using, something else than a protocol in openfe (e.g. a mapper) and suddenly their other simulation that doesn't use openfe but calls pymbar doesn't work anymore the way they wanted. 

Even worse, this doesn't unset the environment variable. So it's a headache even if it's across two scripts.

atravitz: ah you're right - setting then restoring to the original state is likely the safest move.

mikemhenry: If we don't do it this way, IMHO it is too risky that pymbar has already been imported. At import time this envar is checked https://github.com/choderalab/pymbar/blob/main/pymbar/mbar_solvers.py#L17-L23 

I do think we can print a warning BUT people don't like spamy warnings and this will happen at import.

mikemhenry:
We decided to add a warning that links to a document page that describes this behavior. 
If we have time, we will investigate a solution that re-sets the envar to what it was before we changed it/set it.

mikemhenry:
Actually I forgot that when you set an envar within python it only exists for the python process and doesn't stay around:
`$ echo "$PYMBAR_DISABLE_JAX"; python -c "import os; os.environ.setdefault('PYMBAR_DISABLE_JAX', 'TRUE')"; echo "$PYMBAR_DISABLE_JAX"`

this prints 2 blank lines for me, if it stayed around, it would print a blank line and a line that says `TRUE`

mikemhenry:
image
to save you a click (also TIL our theme makes github URLs look nice! I just used a raw URL)

mikemhenry:
pre-commit.ci autofix

docs/guide/troubleshooting.rst
```diff
@@ -73,3 +73,16 @@ If the necessary libraries for GPU acceleration are not installed and JAX detect
 
 This warning does not mean that the *molecular dynamics* simulation will fall back to using the CPU.
 The simulation will still use the computing platform specified in the settings.
+
+PYMBAR_DISABLE_JAX
+------------------
+
+Due to a suspected memory leak in the JAX acceleration code in ``pymbar`` we disable JAX acceleration by default.
+This memory leak may result in the simulation crashing, wasting compute time.
```
IAlibay, line 81: Can you put a snippet of the error being raised here? Like the LLVM out of memory snippet.

mikemhenry:
Good idea, will make searching the docs more useful -- let me know if this looks good @IAlibay @atravitz

atravitz (approved):

mikemhenry:
actually I need to add a news entry, I will edit the JAX warning one

PR #1696, opened 2025-11-25
gather: fix septop rounding

atravitz:
This PR fixes septop analysis behavior where values were being rounded within ddg calculation, then those rounded results were passed to the dg calculation. Now, values are only rounded at the very end.

This fix will need to be applied in https://github.com/OpenFreeEnergy/ExampleNotebooks/blob/main/openmm_septop/septop_analysis.ipynb as well.

Checklist

### DIFF
--- a/openfecli/commands/gather_septop.py
+++ b/openfecli/commands/gather_septop.py
@@ -206,8 +206,7 @@ def _get_ddgs(results_dict: dict[str, dict[str, list]]) -> pd.DataFrame:
     for ligpair, results in sorted(results_dict.items()):
         ddg = np.mean([v[0].m for v in results["overall"]])
         error = error_func(results)
-        m, u = (ddg, error)
-        data.append((ligpair[0], ligpair[1], m, u))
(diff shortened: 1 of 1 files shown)


openfecli/tests/commands/test_gather/test_septop_full_results_dg_.csv
```diff
@@ -1,5 +1,5 @@
 ,ligand,DG (kcal/mol),uncertainty (kcal/mol)
-0,1,-0.67500000000000027,0.66426651277932103
-1,25,1.3250000000000119,1.3119641763401924
-2,7a,-0.87500000000000566,0.9034655499796348
-3,7b,0.22499999999999401,0.90346554997963024
+0,1,-0.71703997416806253,0.66060368310736228
```
atravitz, line 2: @hannahbaumann this PR is to evaluate the actual value differences from fixing the rounding bug - a follow-up PR that checks the rounding precision on output is here: https://github.com/OpenFreeEnergy/openfe/pull/1697

PR #1698, opened 2025-11-25
gather: update names parsing (septop and abfe)

atravitz:
Checklist

### DIFF
--- a/openfecli/commands/gather_septop.py
+++ b/openfecli/commands/gather_septop.py
@@ -133,7 +133,6 @@ def _get_names(result: dict) -> tuple[str, str]:
         Ligand names corresponding to the results.
     """
 
-    # TODO: do we always want to treat "solvent" as the source of truth? should we verify that "complex" has the same name?
     # TODO: is it faster to only load the ligands, or is loading the entire pur cheap?
     solvent_data = list(result["protocol_result"]["data"]["solvent"].values())[0][0]
     pur_solvent = gufe.ProtocolUnitResult.from_dict(solvent_data)


openfecli/commands/gather_abfe.py
```diff
@@ -30,17 +31,15 @@ def _get_name(result: dict) -> str:
     str
         Ligand name corresponding to the results.
     """
-    try:
-        nm = list(result["unit_results"].values())[0]["name"]
 
-    except KeyError:
-        raise ValueError("Failed to guess name")
+    # TODO: do we always want to treat "solvent" as the source of truth? should we verify that "complex" has the same name?
```
IAlibay: for components - no need to verify this for ABFEs, it's explicitly tested

openfecli/commands/gather_abfe.py
```diff
@@ -30,17 +31,15 @@ def _get_name(result: dict) -> str:
...
+    # TODO: is it faster to only load the ligands, or is loading the entire pur cheap?
+    solvent_data = list(result["protocol_result"]["data"]["solvent"].values())[0][0]
+
+    pur_solvent = gufe.ProtocolUnitResult.from_dict(solvent_data)
+
+    name = pur_solvent.inputs["stateA"].components["ligand"].name
```
IAlibay: I _think_ this is a bad footgun! I don't think the "ligand" key is guaranteed.

Edit: instead the right thing to do is to take stateA and stateB and do a component_diff, the unique component is the one who's name you will want.

atravitz: ah, makes sense! A couple other ways we could do this:

1. `name = pur_solvent.inputs['alchemical_components']['stateA'][0].name`: assumes we want the name of the first entry of `stateA`
3. `name = pur_solvent.inputs['stateA'].get_components_of_type(gufe.SmallMoleculeComponent)[0].name`: explicitly pulls `SmallMoleculeComponent`: I see this is a restriction of ABFE's currently, but I'd prefer to stay more flexible.

Also, we can anticipate multiple entries by joining them - @IAlibay is this something that should be valid?
` name  = "_".join([i.name for i in pur_solvent.inputs['alchemical_components']['stateA']])`

IAlibay: Ah yes, looking at `alchemical_components` is the right way to do this! Until we do a major change in ABFEs (there's nothing in the plans for this to ever happen) this should always only ever contain a single entry.

IAlibay: Same thing for SepTop, you can just get name A and name B by looking at the entry for each alchemical component.

openfecli/commands/gather_abfe.py
atravitz: thanks! deleting this TODO

openfecli/commands/gather_abfe.py
atravitz: done! I assume I should use `alchemical_components` for septop as well?

openfecli/commands/gather_septop.py
```diff
@@ -131,16 +132,16 @@ def _get_names(result: dict) -> tuple[str, str]:
...
 
-    except KeyError:
-        raise ValueError("Failed to guess names")
+    # TODO: do we always want to treat "solvent" as the source of truth? should we verify that "complex" has the same name?
+    # TODO: is it faster to only load the ligands, or is loading the entire pur cheap?
+    solvent_data = list(result["protocol_result"]["data"]["solvent"].values())[0][0]
+    pur_solvent = gufe.ProtocolUnitResult.from_dict(solvent_data)
+
+    name_A = pur_solvent.inputs["alchemical_components"]["stateA"][0].name
+    name_B = pur_solvent.inputs["alchemical_components"]["stateB"][0].name
```
atravitz: soliciting @hannahbaumann's feedback on if this is the most robust way to pull names from septop protocol unit results

openfecli/commands/gather_abfe.py
IAlibay: I believe so, yes. It has the same kind of validation checks if I remember correctly.

jthorton (approved):
LGTM I think `alchemical_compoentns` is the most robust way to do this, the alternative of checking for small molecule componentns could be an issue for systems with cofactors.

openfecli/commands/gather_septop.py
hannahbaumann: Yes, this looks good!

PR #1707, opened 2025-12-01
Check user charges are not changed by espaloma

jthorton:
Add a test to make sure user charges are respected by the espaloma small molecule force field

Checklist
* [ ] Ran pre-commit by making a comment with `pre-commit.ci autofix` before requesting review.


jthorton:
pre-commit.ci autofix

openfe/tests/protocols/openmm_md/test_plain_md_protocol.py
```diff
@@ -188,28 +189,43 @@ def test_dry_run_gaff_vacuum(benzene_vacuum_system, vac_settings, tmpdir):
         stateB=benzene_vacuum_system,
         mapping=None,
     )
-    unit = list(dag.protocol_units)[0]
```
IAlibay, line 191: This is why we should slowly rename things to offunit and ommunit, but maybe not in this PR.

jthorton:
>Am I correct in understanding this is happening because "from-molecule" is the default argument if you don't pass any template kwargs?

Yes thats right, so the doc string seems to be incorrect, while its working in a our favour now if we ever pass some args to the generator it could break in future so this test should give us peace of mind for now. I'll open an issue upstream to see what needs to be fixed.

jthorton:
@IAlibay its slightly more complicated than we thought, I found that the section which [assigns the charges](https://github.com/openmm/openmmforcefields/blob/1c969a638f72e8814fd29c5fd86f1af37517a783/openmmforcefields/generators/template_generators.py#L1908-L1922) in the espaloma template generator which does another check for user charges and if they are not present changes the method back to `nn` charges.

IAlibay:
> @IAlibay its slightly more complicated than we thought, I found that the section which [assigns the charges](https://github.com/openmm/openmmforcefields/blob/1c969a638f72e8814fd29c5fd86f1af37517a783/openmmforcefields/generators/template_generators.py#L1908-L1922) in the espaloma template generator which does another check for user charges and if they are not present changes the method back to `nn` charges.

Could you expand on the complication here? I might be misunderstanding, but that bit of the code seems to be doing what I would expect. I.e. if it's got no charges it's going to have to generate some, might as well generate them with espaloma.

jthorton:
My thinking was that it adds another layer of confusion with regards to the docstring, it should should say by default checks for user charges and then if not defaults to espaloma.

PR #1710, opened 2025-12-01
feat: experimental `openfe gather-septop` & `openfe gather-abfe` support

atravitz:
Checklist

### DIFF
--- a/openfe/tests/protocols/openmm_abfe/test_abfe_slow.py
+++ b/openfe/tests/protocols/openmm_abfe/test_abfe_slow.py
@@ -7,12 +7,16 @@ from openff.units import unit as offunit
 
 import openfe
 from openfe.protocols.openmm_afe import AbsoluteBindingProtocol
-from openfe.protocols.openmm_utils.charge_generation import HAS_NAGL
+from openfe.protocols.openmm_utils.charge_generation import HAS_NAGL, HAS_OPENEYE
 
 
 @pytest.mark.integration
 @pytest.mark.flaky(reruns=3)  # pytest-rerunfailures; we can get bad minimisation
 @pytest.mark.skipif(not HAS_NAGL, reason="need NAGL")
+@pytest.mark.xfail(
+    HAS_OPENEYE and HAS_NAGL,
+    reason="NAGL/openeye incompatibility. See https://github.com/openforcefield/openff-nagl/issues/177",
+)
 @pytest.mark.parametrize("platform", ["CUDA"])
 def test_openmm_run_engine(
     platform,
--- a/openfecli/fetchables.py
+++ b/openfecli/fetchables.py
@@ -4,20 +4,6 @@ from openfecli.fetching import PkgResourceFetcher, URLFetcher
 
 _EXAMPLE_NB_BASE = "https://raw.githubusercontent.com/OpenFreeEnergy/ExampleNotebooks/main/"
 
-RBFE_SHOWCASE = URLFetcher(
-    resources=[
-        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "181L_mod_capped_protonated.pdb"),
-        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "Jnk1_ligands.sdf"),
-        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "benzene.sdf"),
-        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "ligands.sdf"),
-        (_EXAMPLE_NB_BASE + "openmm_rbfe/inputs/", "styrene.sdf"),
-    ],
-    short_name="rbfe-showcase",
-    short_help="Inputs needed for the RBFE Showcase Notebook",
-    section="hidden",
-    requires_ofe=(0, 9, 1),
-).plugin
-
 RBFE_TUTORIAL = URLFetcher(
     resources=[
         (_EXAMPLE_NB_BASE + "rbfe_tutorial/", "tyk2_ligands.sdf"),
--- a/openfecli/tests/test_fetchables.py
+++ b/openfecli/tests/test_fetchables.py
@@ -3,7 +3,7 @@ import pathlib
 import pytest
 from click.testing import CliRunner
 
-from openfecli.fetchables import RBFE_SHOWCASE, RBFE_TUTORIAL, RBFE_TUTORIAL_RESULTS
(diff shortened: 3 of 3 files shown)


openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,301 @@
...
+    Returns
+    -------
+    str
+        Ligand name corresponding to the results.
+    """
+
+    # TODO: is it faster to only load the ligands, or is loading the entire pur cheap?
+    solvent_data = list(result["protocol_result"]["data"]["solvent"].values())[0][0]
+    pur_solvent = gufe.ProtocolUnitResult.from_dict(solvent_data)
+    name = pur_solvent.inputs["alchemical_components"]["stateA"][0].name
```
IAlibay: Rather than having to go through the cost of loading the protocol unit result, how about:

```suggestion
    name = solvent_data["inputs"]["alchemical_components"]["stateA"][0]["molprops"]["ofe-name"]
```
In a similar way to how you fetch the name from the ligand atom mapping?

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,301 @@
...
+        dg = np.mean([v[0].m for v in results["overall"]])
+        error = error_func(results)
+        data.append((lig, dg, error))
+
+    df = pd.DataFrame(
+        data,
+        columns=[
+            "ligand",
+            "DG (kcal/mol)",
+            "uncertainty (kcal/mol)",
```
IAlibay: Can you dynamically set the label for uncertainty here? I.e. either MBAR or stdev?

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+def _error_std(r):
+    """
+    Calculate the error of the estimate as the std of the repeats
+    """
+    return np.std([v[0].m for v in r["overall"]])
+
+
+def _error_mbar(r):
```
IAlibay, line 137: I'll be honest my brain isn't working right now and I'm not 100% sure here.

Checking - did you account for the fact that the mbar errors are standard deviations in themselves? I.e. you're averaging the standard deviations properly?

On paper, I _think_ this works, but I would love a double check.

openfecli/commands/gather_abfe.py
```diff
@@ -0,0 +1,301 @@
...
+                    measurement, uncertainty = (repeat[0].m, repeat[1].m)
+                    data.append((simtype, lig, measurement, uncertainty))
+
+    df = pd.DataFrame(
+        data,
+        columns=[
+            "leg",
+            "ligand",
+            "DG (kcal/mol)",
+            "uncertainty (kcal/mol)",
```
IAlibay: make it clear it's the MBAR error here?

openfecli/commands/gather_septop.py
```diff
@@ -0,0 +1,377 @@
...
+    tuple[str, str]
+        Ligand names corresponding to the results.
+    """
+
+    # TODO: is it faster to only load the ligands, or is loading the entire pur cheap?
+    solvent_data = list(result["protocol_result"]["data"]["solvent"].values())[0][0]
+    pur_solvent = gufe.ProtocolUnitResult.from_dict(solvent_data)
+
+    name_A = pur_solvent.inputs["alchemical_components"]["stateA"][0].name
+    name_B = pur_solvent.inputs["alchemical_components"]["stateB"][0].name
```
IAlibay: same thing as above re: faster name access?

openfecli/commands/gather_septop.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+def _error_std(r):
+    """
+    Calculate the error of the estimate as the std of the repeats
+    """
+    return np.std([v[0].m for v in r["overall"]])
+
+
+def _error_mbar(r):
```
IAlibay, line 151: as above re: double checking

openfecli/commands/gather_abfe.py
atravitz: done! and made me realize we need to be casting to a string in the rbfe gather as well: https://github.com/OpenFreeEnergy/openfe/pull/1711

IAlibay: nice catch!

openfecli/commands/gather_abfe.py
atravitz: https://github.com/OpenFreeEnergy/openfe/pull/1712/files

openfecli/commands/gather_abfe.py
atravitz: i'm taking a look, but also pinging @hannahbaumann since she's the original author of the code

openfecli/commands/gather_abfe.py
atravitz: see https://github.com/OpenFreeEnergy/openfe/pull/1712

openfecli/commands/gather_abfe.py
atravitz: My understanding is that this is correct if complex and solvent have the same number of entries, but if the sample sizes differ, we would need to weight accordingly: 

image

https://en.wikipedia.org/wiki/Pooled_variance

I believe we want to account for differing sample sizes, since users may want to look at incomplete results.

hannahbaumann: @IAlibay you might be right, would one maybe have to propagate the RMS instead of the mean error?
instead of `np.mean(complex_errors)` maybe  `np.sqrt(np.mean(complex_errors**2))`?

hannahbaumann: So in total, could it be `np.sqrt(np.mean(complex_errors**2) + np.mean(solvent_errors**2))`?

IAlibay: Brief offline discussion with Alyssa - we agreed with @hannahbaumann's latest version making the most sense. Also the numbers didn't change.

openfecli/commands/gather_septop.py
atravitz: addressed same as above.

news/gather_abfe_septop.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* Added experimental features ``openfe gather-septop`` and ``openfe gather-abfe``, which are analogous to ``openfe gather`` and allow for gathering results generated by the Separated Toplogies and Absolute Binding Free Energy protocols, respectively.  These commands are experimental and are liable to be changed in a future release.
```
IAlibay, line 3: TODO post release - we should have a bit of docs on this.

PR #1773, opened 2026-01-03
Create setup, run, and analysis units for HybridTop Protocol

IAlibay:
Bocked by #1771 and #1772 
Fixes #1724 

Checklist
* [x] All new code is appropriately documented (user-facing code _must_ have complete docstrings).
* [ ] Added a ``news`` entry, or the changes are not user-facing.
* [x] Ran pre-commit by making a comment with `pre-commit.ci autofix` before requesting review.

Manual Tests: these are slow so don't need to be run every commit, only before merging and when relevant changes are made (generally at reviewer-discretion). 
* [x] [GPU integration tests](https://github.com/OpenFreeEnergy/openfe/actions/workflows/gpu-integration-tests.yaml)
* [ ] [example notebook testing](https://github.com/OpenFreeEnergy/openfe/actions/workflows/test-example-notebooks.yaml)
* [ ] [packaging tests](https://github.com/OpenFreeEnergy/openfe/actions/workflows/test-feedstock-pkg-build.yaml): run this for any large feature PRs or PRs that add test data.

## Developers certificate of origin
- [x] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

PR #1774, opened 2026-01-03
Add resume support to HybridTop simulation unit

### LINKED ISSUE #1722: Add resume suppport to HybridTopologyProtocol run unit

IAlibay:
Blocked by #1773 
Fixes #1722 

TODO:
  - Add a test where we resume the unit with an existing partially complete set of files.

Checklist


IAlibay:
pre-commit.ci autofix

IAlibay:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -984,51 +1042,56 @@ def _get_sampler(
...
+            sampler = _SAMPLERS[sampler_method].from_storage(reporter)  # type: ignore[attr-defined]
+
+            # We also do some lightweight checks to make sure we are
+            # running the right system.
+            sampler_system = sampler._thermodynamic_states[0].get_system(remove_thermostat=True)
+            if (
+                (simulation_settings.n_replicas != sampler.n_states != sampler.n_replicas)
+                or (system.getNumForces() != sampler_system.getNumForces())
+                or (system.getNumParticles() != sampler_system.getNumParticles())
+                or (system.getNumConstraints() != sampler_system.getNumConstraints())
```
jthorton: I like this check but could we go even further and check more things, like comparing the seralised xmls of the systems that should compare every force and parameter in the system, assuming ordering is consistent?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -815,6 +815,32 @@ class HybridTopologyMultiStateSimulationUnit(gufe.ProtocolUnit, HybridTopologyUn
...
+        Notes
+        -----
+        For now this just checks if the netcdf files are present in the
+        shared directory but in the future this may expand depending on
+        how warehouse works.
+        """
+        trajectory = shared_path / output_settings.output_filename
+        checkpoint = shared_path / output_settings.checkpoint_storage_filename
+
+        if trajectory.is_file() and checkpoint.is_file():
```
jthorton, line 876: Do we need to check if there is anything in the files, what happens if they were made but are empty?

IAlibay: That's already handled by multisate reporter.

IAlibay: To explain this a little bit more - in that case the reporter/sampler will fail saying it can't read the file.

One option would be to recover by turning off the restart and trying from scratch again in the unit, in practice the only way you could do that is by deleting files - which could be dangerous. Instead, given how rare this situation should be, the way I think it should recover is by executing a fresh version of the unit again, that will give you fresh new inputs that aren't likely corrupted without the problematic case of deleting files.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
IAlibay: Ordering is not consistent unfortunately - the thermodynamic state object adds/pops forces, which reorders things.

IAlibay: Do you have a particular edge case in mind here? I went for a lightweight check as purely as set of guardrails for our users - i.e. the likelihood of a clash like this is rather low.

jthorton: If the order changes, could we zip over the matched forces and compare the seralised versions of just that force? If we are doing this, why not check everything if its possible to be sure? 

One case that this might miss is if the openmm version is different between restarts should we be worried about this?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
jthorton: Okay as long as that is covered somewhere and an error is raised thats fine, even if the user has to delete the files and run again in the meantime before we have an automated solution.

IAlibay: I'll add a specific test for this - it's easy enough to make it fail on an empty file.

hannahbaumann: Can there be any cases where one of the two files is corrupted , leading to a failure/things being out of sync, or is that also handled by the multistatereporter?

IAlibay: > files are empty

If the files are actually empty, then it will fail with an OSError

> checkpoint file is missing

This is actually a case that shouldn't happen - we shouldn't be in a situation where either file is missing, I have added a check in `_check_restart` to cover that case and raise an IOError.

> one of the two files is corrupted

If either of the files are corrupted & unreadable then you will get an OSError (i.e. NetCDF won't be able to recognise the file).

There's no real way to look for bit flips (we don't store checksums or anything).

> checkpoint is not associated with the trajectory

That's checked by multistatereporter (I won't add a test for it).

IAlibay:
pre-commit.ci autofix

IAlibay:
pre-commit.ci autofix

IAlibay:
pre-commit.ci autofix

IAlibay:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -143,6 +145,22 @@ def _get_settings(
...
+        Check that the Python environment hasn't changed based on the
+        relevant Python library versions stored in the setup outputs.
+        """
+        if (
+            (gufe.__version__ != setup_outputs["gufe_version"])
+            or (openfe.__version__ != setup_outputs["openfe_version"])
+            or (openmm.__version__ != setup_outputs["openmm_version"])
+        ):
+            errmsg = "Python environment has changed, cannot continue Protocol execution."
+            raise ProtocolUnitExecutionError(errmsg)
```
jthorton, line 166: Nice check and great to see these errors being used in protocols!

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -972,10 +1042,31 @@ def _get_sampler(
...
           The requested sampler.
         """
+        _SAMPLERS = {
+            "repex": _rfe_utils.multistate.HybridRepexSampler,
+            "sams": _rfe_utils.multistate.HybridSAMSSampler,
+            "independent": _rfe_utils.multistate.HybridMultiStateSampler,
+        }
+
+        sampler_method = simulation_settings.sampler_method.lower()
+
```
jthorton, line 1064: Should we be failing early here if the sampler method is not in the dict or do we no longer need the attribute error that was raised in the old code? I was thinking that the settings validation should have caught the error way before here but is it worth checking?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
jthorton: Should we be catching the keyerror here?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -1273,7 +1382,10 @@ def _execute(
         **inputs,
     ) -> dict[str, Any]:
         log_system_probe(logging.INFO, paths=[ctx.scratch])
-        # Get the relevant inputs
+        # Ensure that we the environment hasn't changed
```
jthorton:
```suggestion
        # Ensure that the environment hasn't changed
```

src/openfe/protocols/openmm_rfe/hybridtop_units.py
IAlibay: done

src/openfe/protocols/openmm_rfe/hybridtop_units.py
IAlibay: I think the attribute error from before was me being overly cautious in the past (a past where we didn't have settings probably).

That being said, it doesn't hurt to move it up and wrap around the key error.

IAlibay:
pre-commit.ci autofix

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_resume.py
```diff
@@ -0,0 +1,462 @@
...
+        # netcdf files properly, so we can't do any follow-up operations
+        # Once openfe-analysis is released, add tests for this.
+        # # Finally we analyze the results
+        # analysis_results = analysis_unit.run(
+        #     pdb_file=setup_results["pdb_structure"],
+        #     trajectory=sim_results["nc"],
+        #     checkpoint=sim_results["checkpoint"],
+        #     scratch_basepath=cwd,
+        #     shared_basepath=cwd,
+        # )
```
jthorton: Is this now fixed in the latest release should we be testing this now?

IAlibay: Yes, it works with the latest release. Added back and a slight check that the npz file is generated.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -984,51 +1087,57 @@ def _get_sampler(
...
+            sampler = _SAMPLERS[sampler_method].from_storage(reporter)  # type: ignore[attr-defined]
+
+            # We do some checks to make sure we are running the same system
+            system_validation.assert_multistate_system_equality(
+                ref_system=system,
+                stored_system=sampler._thermodynamic_states[0].get_system(remove_thermostat=True),
             )
 
+            if (
+                (simulation_settings.n_replicas != sampler.n_states != sampler.n_replicas)
```
jthorton: Can we split this into two explicit comparisons, I think the settings keeps this in sync but just to be sure!

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -1089,23 +1198,28 @@ def _run_simulation(
...
-            # production
+            # At this point we are ready for production
             if self.verbose:
                 self.logger.info("running production phase")
 
-            sampler.extend(int(prod_steps / mc_steps))
+            # We use `run` so that we're limited by the number of iterations
+            # we passed when we built the sampler.
+            # TODO: I'm being extra prudent by passing in n_iterations here - remove?
+            sampler.run(n_iterations=int(prod_steps / mc_steps) - sampler._iteration)
```
jthorton, line 1236: Could this be 0 or negative if some settings have been changed before the restart or if you try to restart a finished simulation?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -1273,7 +1394,10 @@ def _execute(
         **inputs,
     ) -> dict[str, Any]:
         log_system_probe(logging.INFO, paths=[ctx.scratch])
-        # Get the relevant inputs
+        # Ensure that the environment hasn't changed
+        self._verify_execution_environment(setup_results.outputs)
```
jthorton, line 1412: Should we move this check into run? For now you can skip this if you only can run rather than execute, I think.

IAlibay: That's intentional. `run` is a convenience so that we can manually execute units for testing, so any checking from the results of the previous outputs gets done in `_execute` that way `run` can just be fed python objects directly.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -984,51 +1087,57 @@ def _get_sampler(
...
-                hybrid_positions=positions,
-                online_analysis_interval=rta_its,
-                online_analysis_target_error=early_termination_target_error,
-                online_analysis_minimum_iterations=rta_min_its,
+        # Restarting doesn't need any setup, we just rebuild from storage.
+        if restart:
+            sampler = _SAMPLERS[sampler_method].from_storage(reporter)  # type: ignore[attr-defined]
+
+            # We do some checks to make sure we are running the same system
+            system_validation.assert_multistate_system_equality(
```
jthorton, line 1114: For a restart, we are also not checking if the temperature, preasure or lambda values have been changed we just silently use the ones in the checkpoint file is this worth checking? I am not sure how many people would change this stuff its more a just incase.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
IAlibay: I'm not fully sure I understand this one, since n_states should be n_replicas. I've split it into two lines.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
IAlibay: You can't change settings before a restart (it's tied to the DAG and the settings are frozen).
I don't think there's a way for this to be negative.

If it is zero, then it will just not run and pass as True here: https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/multistatesampler.py#L1722

jthorton (approved):
This is great, love all the gaurdrails excited to see what we missed once this hits production!

src/openfe/protocols/openmm_rfe/hybridtop_units.py
IAlibay: Took me longer than I want to admit, but done - looping over the thermodynamic states and checking temperature, pressure and lambda values.

PR #1814, opened 2026-01-28
add --download-only option to openfe test

atravitz:
this will allow us to:
- pull all the test data up-front, so that we fail fast if there's a zenodo-fetching issue
- enable users to pull data on a instance with internet access, then run tests on air-gapped compute
- both cli and openfe zenodo info is stored in a registry - this should make it easier to migrate to cloudflare in the future, plus we can maybe diff that file to trigger a cache update for CI

Checklist


src/openfe/tests/protocols/conftest.py
```diff
@@ -324,18 +328,13 @@ def t4_lysozyme_trajectory_universe():
     return universe
 
 
-RFE_OUTPUT = pooch.create(
```
atravitz, line 324: note: we moved to from figshare to zenodo here: https://github.com/OpenFreeEnergy/openfe/blob/f7561cda2249e5fc477d51c4db42134b1ed6d399/src/openfe/tests/protocols/test_openmmutils.py#L1037-L1038

but this was left as a stragger no-op.

mikemhenry (approved):
LGTM, a hash of the `_registry.py` files can now be used as the cache key, this will also make it easy to migrate things somewhere else since we only have one file to track.

PR #1846, opened 2026-02-14
Safely close reporter and clear GPU contexts to avoid UnboundLocalErrors

### LINKED ISSUE #1845: UnboundLocalErrors popping up when Protocols fail
This is something I actually fixed in #1774 but the fix accidentally didn't make it into the v1.9 release.

Our protocols try/except around the reporter & sampler creation as well as the simulation running now. The idea is that in the `finally` step, we clean up any leftover objects, since those tend to be slow to be garbage collected (if at all) and can lead to memory bloating in executors like Alchemiscale.

Historically this was only around the simulation running, but now it's been extended to the sampler & reporter creation. However, because of this, there's a non-zero way that you can crash early (e.g. during sampler or integrator creation) and hit the `finally` block. This will then lead to an UnboundLocalError, since e.g. the sampler hasn't been created yet.

IAlibay:
Fixes #1845

Added try/except block to safely close things in case there's an unbound variable.

Checklist

* [x] [GPU integration tests](https://github.com/OpenFreeEnergy/openfe/actions/workflows/gpu-integration-tests.yaml)

### DIFF
--- a/news/issue-1845.rst
+++ b/news/issue-1845.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed a bug in Protocol termination for the HybridTop and AFE Protocols
+  which would unecessary declare an UnboundLocalError.
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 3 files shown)


IAlibay:
pre-commit.ci autofix

news/issue-1845.rst
```diff
@@ -0,0 +1,24 @@
...
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed a bug in Protocol termination for the HybridTop and AFE Protocols
+  which would unecessary declare an UnboundLocalError.
```
atravitz:
```suggestion
* Fixed a bug in Protocol termination for the HybridTop and AFE Protocols
  which would unnecessarily declare an ``UnboundLocalError``.
```

atravitz (approved):
lgtm!

PR #1848, opened 2026-02-16
Modify quickrun to allow resuming

IAlibay:
Minimal changes to quickrun in order to allow resuming by serializing ProtocolDAGs.
Related to & requires https://github.com/OpenFreeEnergy/gufe/pull/738/

Checklist


IAlibay:
pre-commit.ci autofix

src/openfe/tests/protocols/openmm_abfe/test_abfe_protocol_results.py
```diff
@@ -79,12 +79,12 @@ def patcher():
         yield
 
 
-def test_gather(benzene_complex_dag, patcher, tmpdir):
+def test_gather(benzene_complex_dag, patcher, tmp_path):
```
IAlibay, line 86: Here and everywhere else I'm doing a tmpdir -> tmp_path change, turns out tmpdir's implementation in pytest doesn't implement all the pathlib.Path features, so it's better to use tmp_path.

As far as I'm aware the slow tests aren't affected by this because we wrap them around `Path` calls.

IAlibay:
I don't know what's up with the tyk2 CLI test, but locally the restart is working!

src/openfe/tests/protocols/openmm_abfe/test_abfe_protocol_results.py
ethanholz: As a note, I updated some of the harnesses in `gufe`  to use this as well.

atravitz:
question: do we want this to be the default behavior, or require ``openfe quickrun --resume``?

src/openfecli/commands/quickrun.py
```diff
@@ -94,13 +96,28 @@ def quickrun(transformation, work_dir, output):
...
-    dag = trans.create()
+    # Attempt to either deserialize or freshly create DAG
+    trans_DAG_json = work_dir / f"{trans.key}-protocolDAG.json"
+
+    if trans_DAG_json.is_file():
+        write(f"Attempting to resume execution using existing edges from '{trans_DAG_json}'")
+        try:
+            dag = ProtocolDAG.from_json(trans_DAG_json)
+        except JSONDecodeError:
+            errmsg = f"Recovery failed, please remove {trans_DAG_json} and any results from your working directory before continuing to create a new protocol."
```
mikemhenry: I like that we are not using a flag to control if we want to resume a simulation or not and instead just check if something already exists

atravitz: my only concern with this is that it changes default behavior, instead of it being opt-in behavior. I'd like @IAlibay's thoughts on what would be least disruptive to users.

IAlibay: I think we should have a call to discuss it more.  I originally fell on a "no flag" thought, but @atravitz raised a good point about avoiding accidental "double restarts" or "restarts whilst you're already running something". And then there's also the consideration that folks might be calling `quickrun` with the same output folder directory multiple times but writing to different output files. I'm not too sure what happens in that case.

mikemhenry: I would be happy if it was opt in, but a `--resume` flag would be a no-op if there was nothing to resume. That would make it easy to use the same command to run and re-run failed jobs.

atravitz: recapping today's conversation with myself, @IAlibay, and @mikemhenry: 

- remove the protocol_dag checkpoint when the job has completed 
- add a `--resume` flag that means quickrun will look for a checkpoint. if the checkpoint doesn't exist, raise a warning and run anyway. 
    - *not discussed today but my addition*: if the checkpoint is broken (corrupted json or something), I think we should throw an error.
- if `--resume` is NOT used, but the checkpoint file exists:  error out with a "did you mean to resume?" error message.

mikemhenry: > not discussed today but my addition: if the checkpoint is broken (corrupted json or something), I think we should throw an error.

Yes I think we should error in that case. Also your summary is correct!

news/quickrun_resume.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* Added ``--resume`` flag to ``openfe quickrun``.
```
IAlibay, line 3: We probably need to add some docs in the user guide for it. I'll open a separate issue / PR.

src/openfecli/commands/quickrun.py
```diff
@@ -94,13 +103,37 @@ def quickrun(transformation, work_dir, output):
...
+        if resume:
+            write(f"Attempting to resume execution using existing edges from '{trans_DAG_json}'")
+            try:
+                dag = ProtocolDAG.from_json(trans_DAG_json)
+            except JSONDecodeError:
+                errmsg = f"Recovery failed, please remove {trans_DAG_json} and any results from your working directory before continuing to create a new protocol, or run without `--resume`."
+                raise click.ClickException(errmsg)
+        else:
+            errmsg = f"Transformation has been started but is incomplete. Please remove {trans_DAG_json} and rerun, or resume execution using the ``--resume`` flag."
+            raise RuntimeError(errmsg)
```
IAlibay: You use click Exception above. Is there a difference between using it and not? (i.e. it's not clear to me if the intent is to always try to use the click Exception or if it's more context dependent).

IAlibay (commented):
Just the one question but otherwise lgtm. I'll retest on a real FEP job today too.

news/quickrun_resume.rst
atravitz: linking here: https://github.com/OpenFreeEnergy/openfe/pull/1878

src/openfecli/commands/quickrun.py
atravitz: I originally used a ClickException but then decided to use a RuntimeError - now that I'm looking at it, I'll make both RuntimeError.

My rationale is that ClickExceptions should be for things Click or CLI related (see examples here https://click.palletsprojects.com/en/stable/api/#click.ClickException)

whereas this is really an openfe execution exception.

atravitz: follow-up: after playing around with it, the key difference to users is that `RuntimeError` will show a traceback and look scarier, whereas a `ClickException`  will just print the message with no traceback.

Do you have an opinion on our noisiness here? I'm slightly leaning toward `ClickException` now, since these are both cases that should be easily resolvable and not require the user to report the error.

IAlibay: Agreed on preferring ClickException.

atravitz (approved):

src/openfecli/commands/quickrun.py
```diff
@@ -94,13 +103,40 @@ def quickrun(transformation, work_dir, output):
     else:
         output.parent.mkdir(exist_ok=True, parents=True)
 
-    write("Planning simulations for this edge...")
-    dag = trans.create()
+    # Attempt to either deserialize or freshly create DAG
+    cache_basedir = work_dir / "quickrun_cache"
+    trans_DAG_json = cache_basedir / f"{trans.key}-ProtocolDAG.json"
```
atravitz: right now the directory structure (with https://github.com/OpenFreeEnergy/gufe/pull/764/files) looks like:
```
quickrun_cache/
├── ProtocolDAG-8090c89c950e976b33829a15e662ed89-results_cache/
└── Transformation-dbea03c534737749bb413e01a382f3af-protocolDAG.json
```

is there a better way to indicate that `Transformation-dbea03c534737749bb413e01a382f3af-ProtocolDAG.json` is the `ProtocolDAG` _corresponding to_ `Transformation-dbea03c534737749bb413e01a382f3af`?

IAlibay: You could move the dag creation up a little bit, and just use the dag key as part of the name?

I.e. of you define `trans_DAG_json` after you define dag, it could be f"ProtocolDAG-{dag.key}.json" instead.

IAlibay:
Testing locally with multiple interrupts / resumes is working as expected.

src/openfecli/commands/quickrun.py
atravitz: in offline conversation, @IAlibay and I agree that moving dag creation up will _not_ work, because each time .create() is called, a new unique ProtocolDAG is created.

We also realize that, as of right now, the cache is only unique based on the transformation key and the `-d` argument. This means that if a user called the following two commands in succession:

`openfe quickrun transformation.json -o result1.json -d results/`
`openfe quickrun transformation.json -o result2.json -d results/`

openfe will treat these as a re-execution and raise an error telling the user that they should pass in `--resume` or delete the file and restart.

There are two possible solutions to this:

1. Enforce user behavior to have separate `-d` values for each repeat (essentially what is implemented now)
2. Build the hash based on the uniqueness of all 3 1. transformation key 2. `-o`, and 3. `-d`. Since -o can be any arbitrary filepath we may want to hash `-o` and store it as:
`[-d arg]/quickrun_cache/dagcache-[hash(transformation.key, -o arg)]`.

atravitz: option 2 has been implemented in https://github.com/OpenFreeEnergy/openfe/pull/1890

IAlibay (commented):
lgtm, thanks!

PR #1856, opened 2026-02-23
Fix 1-4 interpolation for core-dummy interactions

jthorton:
This fixes the treatment of 1-4 interactions between core and dummy atoms and extensive tests to the hybrid topology factory to ensure they are handled correctly. 

Checklist
* [x] All new code is appropriately documented (user-facing code _must_ have complete docstrings).
* [x] Added a ``news`` entry, or the changes are not user-facing.
* [x] Ran pre-commit: you can run [pre-commit](https://pre-commit.com) locally or comment on this PR with `pre-commit.ci autofix`.

Manual Tests: these are slow so don't need to be run every commit, only before merging and when relevant changes are made (generally at reviewer-discretion). 
* [ ] [GPU integration tests](https://github.com/OpenFreeEnergy/openfe/actions/workflows/gpu-integration-tests.yaml)
* [ ] [example notebook testing](https://github.com/OpenFreeEnergy/openfe/actions/workflows/test-example-notebooks.yaml)
* [ ] [packaging tests](https://github.com/OpenFreeEnergy/openfe/actions/workflows/test-feedstock-pkg-build.yaml): run this for any large feature PRs or PRs that add test data.

## Developers certificate of origin
- [x] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

PR #1858, opened 2026-02-25
Speed up HTF creation

jthorton:
Try to speed up the HTF creation, targeting the torsion force and nonbonded exceptions as they scale terribly with system size. Local testing on a small protein-ligand system shows good improvements

Main
> systems built starting test
Handled harmonic bonds in 0.36160308588296175 seconds
Handled harmonic angles in 1.0089957918971777 seconds
Handled periodic torsions in 7.016903054900467 seconds
Handled nonbonded interactions in 25.400564597919583 seconds
HybridTopologyFactory took 37.79149968735874 seconds

This branch
> systems built starting test
Handled harmonic bonds in 0.3190057808533311 seconds
Handled harmonic angles in 0.9470380647107959 seconds
Handled periodic torsions in 0.24904045276343822 seconds
Time to handle original exceptions 0.51 seconds
Handled nonbonded interactions in 2.244331883266568 seconds
Handled nonbonded exceptions in 0.04826593864709139 seconds
HybridTopologyFactory took 7.013191535137594 seconds

We probably need the tests in #1856 to confirm that nothing is broken with these changes.

## TODO
- test on membrane system 

Checklist


jthorton:
pre-commit.ci autofix

mikemhenry:
Oh this is very promising -- how stable are these timings?  Do you think it would be worth making a test that ensures (for this test case) `HybridTopologyFactory` takes less than 10 seconds to build (might need to adjust depending on how long it takes on CI) but it would be nice to have some checks we don't lose these speed-ups later?

jthorton:
> how stable are these timings?

Locally, they are within a 3 second window so we could make a timing test, though I am not sure how common that is?We could also provide a timing script to be run whenever someone touches the HTF, which should not be often. Any preference @atravitz ?

jthorton:
With https://github.com/OpenFreeEnergy/openfe/pull/1858/commits/bccc53c5e17493b505cbd8a99b4aea559edc0745 the timings are now 
> systems built starting test
Handled harmonic bonds in 0.04707380570471287 seconds
Handled harmonic angles in 0.1307398546487093 seconds
Handled periodic torsions in 0.2686290815472603 seconds
Handled nonbonded interactions in 2.285492484457791 seconds
Handled nonbonded exceptions in 0.04942901339381933 seconds
HybridTopologyFactory took 6.055246475152671 seconds

mikemhenry:
There are a few ways to track regressions, one cool way is to make sure the algo doesn't change the way it scales: https://pythonspeed.com/articles/big-o-tests/ but we don't really have things set up to make that easy (but it would be good to know how the HTF scales with system size, so it might be worth doing)

atravitz:
> > how stable are these timings?
> 
> Locally, they are within a 3 second window so we could make a timing test, though I am not sure how common that is?We could also provide a timing script to be run whenever someone touches the HTF, which should not be often. Any preference @atravitz ?

I like the idea of a timing script for now. I want us to add performance benchmarking like this, but structure is TBD, so pop it in the `devtools/` directory for now.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1225,18 +1235,18 @@ def _handle_harmonic_bonds(self):
...
-                    # TODO - why is this being recalculated?
-                    [index1, index2, r0_new, k_new] = self._find_bond_parameters(
-                        new_system_bond_force, index1_new, index2_new)
+                    r0_new, k_new = new_bond_parameters
+
                 self._hybrid_system_forces['core_bond_force'].addBond(
                     index1_hybrid, index2_hybrid,
                     [r0_old, k_old, r0_new, k_new])
+                # track that we've added this bond
+                hybrid_bonds[self._pair_key(index1_hybrid, index2_hybrid)] = True
```
IAlibay: Is there any reason for this to be a dictionary? Could keeping a set or list of pair key indices be suitable?

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1159,33 +1173,25 @@ def _add_nonbonded_force_terms(self):
...
-        # Loop through all the bonds:
+        bond_lookup = {}
         for bond_index in range(bond_force.getNumBonds()):
-            parms = bond_force.getBondParameters(bond_index)
-            if index_set == {parms[0], parms[1]}:
-                return parms
-
-        return []
+            index1, index2, r0, k = bond_force.getBondParameters(bond_index)
+            bond_lookup[HybridTopologyFactory._pair_key(index1, index2)] = (r0, k)
```
IAlibay: Could we ever be in a situation where there are multiple bonds acting on the same pair of indices? It would be odd, ,but I don't think there's anything in HarmonicBondForce that prevents that.

Given the likely low cost, I would probably suggest capturing that case with a quick "if pairkey in bond_lookup" check.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1322,36 +1334,46 @@ def _handle_harmonic_bonds(self):
...
-        return []  # Return empty if no matching angle found
+        Returns
+        -------
+        dict
+            A dictionary mapping a sorted tuple of atom indices to their angle parameters.
+        """
+        angle_lookup = {}
+        for angle_index in range(angle_force.getNumAngles()):
+            index1, index2, index3, theta0, k = angle_force.getAngleParameters(angle_index)
+            angle_lookup[HybridTopologyFactory._triplet_key(index1, index2, index3)] = (theta0, k)
```
IAlibay: Same as above re: bonds with the same keys, should we guard against a case where you could have the same key but with a different angle term?

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1401,26 +1427,27 @@ def _handle_harmonic_angles(self):
...
                     old_angle_parameters[3], old_angle_parameters[4],
-                    new_angle_parameters[3], new_angle_parameters[4]
+                    new_theta0, new_k
                 ]
                 self._hybrid_system_forces['core_angle_force'].addAngle(
                     hybrid_index_list[0], hybrid_index_list[1],
                     hybrid_index_list[2], hybrid_force_parameters
                 )
+                # track that we have added this angle for the hybrid system
+                hybrid_angles[self._triplet_key(*hybrid_index_list)] = True
```
IAlibay: Same as above - how about using a list or a set?

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1578,9 +1574,13 @@ def _handle_periodic_torsion_force(self):
         """
         old_system_torsion_force = self._old_system_forces['PeriodicTorsionForce']
         new_system_torsion_force = self._new_system_forces['PeriodicTorsionForce']
+        # local variables for speed while doing many lookups
+        unique_old_atoms = self._atom_classes["unique_old_atoms"]
+        unique_new_atoms = self._atom_classes["unique_new_atoms"]
 
-        auxiliary_custom_torsion_force = []
-        old_custom_torsions_to_standard = []
+        # use sets to keep membership checks quick as systems have many torsions
```
IAlibay: Wouldn't sets drop torsions with equal indices? My understanding is that this is more common / acceptable.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1613,13 +1613,13 @@ def _handle_periodic_torsion_force(self):
                 # core/env term; in any case, it goes to the core torsion_force
                 # TODO - why are we even adding the 0.0, 0.0, 0.0 section?
                 hybrid_force_parameters = [
-                    torsion_parameters[4], torsion_parameters[5],
-                    torsion_parameters[6], 0.0, 0.0, 0.0
+                    torsion_parameters[4], torsion_parameters[5].value_in_unit(unit.radian),
```
IAlibay: It's not immediately clear to me, why the `value_in_unit` change?

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1641,16 +1641,15 @@ def _handle_periodic_torsion_force(self):
             else:
                 hybrid_force_parameters = [
                     0.0, 0.0, 0.0, torsion_parameters[4],
-                    torsion_parameters[5], torsion_parameters[6]]
+                    torsion_parameters[5].value_in_unit(unit.radian), torsion_parameters[6].value_in_unit(unit.kilojoule_per_mole)]
 
                 # Check to see if this term is in the olds...
-                term = [hybrid_index_list[0], hybrid_index_list[1],
+                term = (hybrid_index_list[0], hybrid_index_list[1],
```
IAlibay: Exact quality on converted floats seems like it would increase the likelyhood of a miss here - it would be better if the comparisons was done to a certain float level instead, e.g. truncating the floats to 5 d.p. or something.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1641,16 +1641,15 @@ def _handle_periodic_torsion_force(self):
...
                         hybrid_index_list[2], hybrid_index_list[3],
-                        hybrid_force_parameters[3:]]
+                        *hybrid_force_parameters[3:])
                 if term in auxiliary_custom_torsion_force:
                     # Then this terms has to go to standard and be deleted...
-                    old_index = auxiliary_custom_torsion_force.index(term)
-                    old_custom_torsions_to_standard.append(old_index)
+                    old_custom_torsions_to_standard.add(term)
```
IAlibay: [nit] (I know it's a holdover from the previous code) maybe we should rename this variable - it's confusing

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
IAlibay: If we're looking for other performance improvements - you could probably cache this?

IAlibay (changes requested):
Overall looks great, but there's a few things - especially key collisions, that probably should be further discussed / handled.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Nope good catch, changed this to a set.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Yeah this should be a set as well changed it over.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: We store the full torsion in the set, including the parameters so torsions with different phases and periodicities covering the same atoms should be unique, the full system energy tests should also pick this up if we drop torsions.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: I think this was to get the set hashing to work.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Good point I think it is allowed, added a catch for it and a test for bonds and angles.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Added a test and a catch for this case.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Even though the floats are converted, we do request the value in the OpenMM internal units, so it should be safe, as it's just the value of the quantity provided, both systems have been parameterised using the same procedure and the same precision. We could add rounding for the hash and change to a dict where the full precision version is stored as the key if we think this is an issue, but this copies the current implimentation.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Yeah good idea updated it.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Good find added this as well.

jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
IAlibay: Let's keep it as-is for now and revisit if someone reports issues.

IAlibay (changes requested):
Overall this looks good - the only thing is the potential float equality mismatch we discussed offline. Your proposed approach of having a dictionary with the rounded values could work.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1578,105 +1611,134 @@ def _handle_periodic_torsion_force(self):
         """
         old_system_torsion_force = self._old_system_forces['PeriodicTorsionForce']
         new_system_torsion_force = self._new_system_forces['PeriodicTorsionForce']
+        # local variables for speed while doing many lookups
```
jthorton, line 1614: This section was rewritten with GPT-5.3-codex to optimise performance and move to the rounded key comparison for torsions. I confirmed that dropping torsions from the shared and unique sets caused the energy tests to fail locally.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1578,105 +1611,134 @@ def _handle_periodic_torsion_force(self):
...
+            -------
+            unique_terms: list[tuple[int, int, int, int, int, Quantity, Quantity]]
+                A list of torsion terms that are unique to this end state. Each term is a tuple of (hybrid_indices, periodicity, phase_value, k_value)
+            interpolated_terms: dict[tuple[int, int, int, int, int, float, float], tuple[int, int, int, int, int, Quantity, Quantity]]
+                A dictionary of torsion terms that are shared between the two end states. The keys are tuples using the rounded
+                keys while the values are the same at full precision. Each value is a tuple of (hybrid_indices, periodicity, phase_value, k_value)
+            """
+            unique_terms = []
+            interpolated_terms = {}
+            get_torsion_parameters = torsion_force.getTorsionParameters
```
IAlibay, line 1662: [nit] not sure if this was a codex suggestion or yours. If it's the former, the difference in the length really doesn't make it worth it, i.e. I would just use the full torsion_force.getTorsionParameters, especially given that you're only using it once.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1578,105 +1611,134 @@ def _handle_periodic_torsion_force(self):
...
+                            torsion_parameters[6],
+                        )
+                    )
+                    continue
+
+                periodicity = torsion_parameters[4]
+                phase_value = torsion_parameters[5].value_in_unit(unit.radian)
+                k_value = torsion_parameters[6].value_in_unit(unit.kilojoule_per_mole)
+                term_key = _rounded_term_key(hybrid_indices, periodicity, phase_value, k_value)
+                interpolated_terms[term_key] = (
```
IAlibay, line 1683: Just double checking - we're confident that it's unreasonable for someone to have a case where they have the exact same force acting on the same atoms twice?

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
IAlibay: I don't think these are being used anymore, should they be removed?

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -2044,6 +2049,8 @@ def _handle_original_exceptions(self):
         new_system_nonbonded_force = self._new_system_forces['NonbondedForce']
```
IAlibay: Are `new_system_nonbonded_force` and `old_system_nonbonded_force` here also now redundant?

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
IAlibay: I think these are also ok to remove.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1225,18 +1269,18 @@ def _handle_harmonic_bonds(self):
             if index_set.issubset(self._atom_classes['core_atoms']):
                 index1_new = self._old_to_new_map[index1_old]
                 index2_new = self._old_to_new_map[index2_old]
-                new_bond_parameters = self._find_bond_parameters(
-                    new_system_bond_force, index1_new, index2_new)
-                if not new_bond_parameters:
+                new_bond_parameters = new_bond_lookup.get(self._pair_key(index1_new, index2_new), None)
```
IAlibay, line 1272: I really don't think it matters, so consider this just a comment to be discarded. Technically this is set up so that there can be duplicate bonds in the old force and not in the new one.

In practice.. I don't think it makes sense that someone would have 2+ bond entries for the same two atoms.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
```diff
@@ -1578,105 +1611,134 @@ def _handle_periodic_torsion_force(self):
...
         old_system_torsion_force = self._old_system_forces['PeriodicTorsionForce']
         new_system_torsion_force = self._new_system_forces['PeriodicTorsionForce']
+        # local variables for speed while doing many lookups
+        unique_old_atoms = self._atom_classes["unique_old_atoms"]
+        unique_new_atoms = self._atom_classes["unique_new_atoms"]
+        old_to_hybrid_map = self._old_to_hybrid_map
+        new_to_hybrid_map = self._new_to_hybrid_map
+
+        unique_torsion_force = self._hybrid_system_forces['unique_atom_torsion_force']
+        unique_torsion_add = unique_torsion_force.addTorsion
```
IAlibay, line 1621: [nit] to me, this is less readable than just doing the full thing.

IAlibay (approved):
Overall this looks good, couple of small things but nothing I'd consider blocking, so I'll approve early.

IAlibay:
> With [bccc53c](https://github.com/OpenFreeEnergy/openfe/commit/bccc53c5e17493b505cbd8a99b4aea559edc0745) the timings are now
> 
> > systems built starting test
> > Handled harmonic bonds in 0.04707380570471287 seconds
> > Handled harmonic angles in 0.1307398546487093 seconds
> > Handled periodic torsions in 0.2686290815472603 seconds
> > Handled nonbonded interactions in 2.285492484457791 seconds
> > Handled nonbonded exceptions in 0.04942901339381933 seconds
> > HybridTopologyFactory took 6.055246475152671 seconds

Can you do another one of these with the final state of this PR? I suspect with the nonbonded force stuff, this should be faster too.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Good point though I don't think amber or the openff force fields allow this currently but something to keep in mind.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Yeah this is a codex speed trick in hot loops, the assignment stops the function lookup on the torsion force and should be slightly faster.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: That would be strange - while its possible, if we take amber and openff as a reference I don't think this should happen in practice.

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Good catch!

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Removed thanks!

jthorton: Removed thanks!

src/openfe/protocols/openmm_rfe/_rfe_utils/relative.py
jthorton: Yeah its not perfect but its a hot loop speed thing apprently so I'll keep it for now.

jthorton:
With all the extra changes we get a small improvment in the nonbonded section 
> ready to handle forces, took 2.576 seconds
Handled harmonic bonds in 0.046 seconds
Handled harmonic angles in 0.129 seconds
Periodic torsion force in 0.323 seconds
Handled nonbonded force in 2.149 seconds
HybridTopologyFactory took 6.020 seconds

I noticed that it we remove the deepcopies we can get another speedup, what do we think about removing it?
> ready to handle forces, took 1.843 seconds
Handled harmonic bonds in 0.046 seconds
Handled harmonic angles in 0.131 seconds
Periodic torsion force in 0.320 seconds
Handled nonbonded force in 2.143 seconds
HybridTopologyFactory took 5.252 seconds

jthorton:
pre-commit.ci autofix

IAlibay (approved):

Issue #1869, opened 2026-03-10
conda cron searching for dev version


atravitz:
I think we need to update how we define the "latest" version:
https://github.com/OpenFreeEnergy/openfe/actions/runs/22888859553

mikemhenry:
Good point:
`LATEST_TAG=$(git describe --tags $(git rev-list --tags --max-count=1))`
Just grabs the latest tag, which might not be the latest version, how about this?
```
curl -s https://api.github.com/repos/OpenFreeEnergy/openfe/releases/latest | jq -r '.tag_name | ltrimstr("v")
```

That will get the latest version we have tagged as "latest", get the tag name, then trim off the v.

PR #1870, opened 2026-03-10
Use GitHub API to fetch latest release/version

mikemhenry:
Checklist

fixes issue #1869

### DIFF
--- a/.github/workflows/conda_cron.yaml
+++ b/.github/workflows/conda_cron.yaml
@@ -41,7 +41,8 @@ jobs:
         working-directory: openfe_repo
         run: |
           REPO="${{ github.repository }}"
-          VERSION=$(curl -s "https://api.github.com/repos/$REPO/releases/latest" | jq -r '.tag_name | ltrimstr("v")')
+          VERSION=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            "https://api.github.com/repos/$REPO/releases/latest" | jq -r '.tag_name | ltrimstr("v")')
           echo $VERSION
           echo "VERSION=$VERSION" >> $GITHUB_OUTPUT
 
--- a/.github/workflows/conda_cron.yaml
+++ b/.github/workflows/conda_cron.yaml
@@ -40,9 +40,7 @@ jobs:
         id: latest-version
         working-directory: openfe_repo
         run: |
-          LATEST_TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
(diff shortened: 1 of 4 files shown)


mikemhenry:
I could use `$REPO` to make it easier to copy and paste into other repositories

mikemhenry:
If it works I will actually do that

mikemhenry:
testing here testing here https://github.com/OpenFreeEnergy/openfe/actions/runs/22920034257

mikemhenry:
testing update version here: https://github.com/OpenFreeEnergy/openfe/actions/runs/22920555651

mikemhenry:
testing version with auth 
https://github.com/OpenFreeEnergy/openfe/actions/runs/22920833146

.github/workflows/conda_cron.yaml
```diff
@@ -40,9 +40,9 @@ jobs:
         id: latest-version
         working-directory: openfe_repo
         run: |
-          LATEST_TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
-          # slice off the v, ie v0.7.2 -> 0.7.2
-          VERSION=${LATEST_TAG:1}
+          REPO="${{ github.repository }}"
+          VERSION=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
```
atravitz, line 44: are you using this instead of `gh release list` to avoid having to use gh?

atravitz (approved):
one non-blocking question

.github/workflows/conda_cron.yaml
mikemhenry: I wasn't sure if `gh` would be on the runners. I just looked at using `gh release list` and it would require a bit more work since it doesn't look like there is a way to just fetch the latest, but instead `isLatest` so it would take some extra filtering:

```
$ gh release list --json isLatest,tagName
[
  {
    "isLatest": false,
    "tagName": "v1.9.1dev0"
  },
  {
    "isLatest": true,
    "tagName": "v1.9.1"
  },
  {
    "isLatest": false,
    "tagName": "v1.9.0"
  }
]  
```
Instead, here we are leveraging the `latest` api end point. Also if you ever want to use this, you don't need the auth headers, just included that to avoid rate limiting:
`curl -s https://api.github.com/repos/OpenFreeEnergy/openfe/releases/latest | jq -r '.tag_name | ltrimstr("v")'`

atravitz: for the sake of mental exercise, `gh`, does have the `--jq` option! but I think the approach you did here is good.

PR #1876, opened 2026-03-17
Minor fix to support ABFE gathering with openfe v1.9+

IAlibay:
Small update to gather machinery to avoid picking results from UnitResults that aren't the Analysis unit.
It's not the nicest fix, but it works in a backwards compatible manner.

### DIFF
--- a/src/openfecli/commands/gather_abfe.py
+++ b/src/openfecli/commands/gather_abfe.py
@@ -116,8 +116,14 @@ def _get_legs_from_result_jsons(
             continue
 
         dgs[name]["overall"].append([result["estimate"], result["uncertainty"]])
-        proto_key = [k for k in result["unit_results"].keys() if k.startswith("ProtocolUnitResult")]
+        proto_key = [
+            k for k in result["unit_results"].keys()
+            if k.startswith("ProtocolUnitResult")
+        ]
         for p in proto_key:
+            if ("Setup" in result["unit_results"][p]["source_key"]
+                or "Simulation" in result["unit_results"][p]["source_key"]):
+                continue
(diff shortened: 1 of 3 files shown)


IAlibay:
pre-commit.ci autofix

atravitz:
@IAlibay from what I can tell, this only affects the output for `--report=raw`, not the computed dg or ddg values at all.

I'm generally in favor of having `raw` be as unprocessed and transparent as possible, but if this is genuinely just extra information that isn't ever useful, I can be convinced that this is a useful patch.

IAlibay:
> @IAlibay from what I can tell, this only affects the output for `--report=raw`, not the computed dg or ddg values at all.
> 
> I'm generally in favor of having `raw` be as unprocessed and transparent as possible, but if this is genuinely just extra information that isn't ever useful, I can be convinced that this is a useful patch.

It's both unuseful and also somewhat dangerous (in the case of picking up results from intermediary units). Personally I would prefer we try to keep behaviour the same as pre-1.9 because it's "what someone doing this experiment would expect".

atravitz:
> > @IAlibay from what I can tell, this only affects the output for `--report=raw`, not the computed dg or ddg values at all.
> > I'm generally in favor of having `raw` be as unprocessed and transparent as possible, but if this is genuinely just extra information that isn't ever useful, I can be convinced that this is a useful patch.
> 
> It's both unuseful and also somewhat dangerous (in the case of picking up results from intermediary units). Personally I would prefer we try to keep behaviour the same as pre-1.9 because it's "what someone doing this experiment would expect".

makes sense to me, I'll go ahead and merge!

PR #1896, opened 2026-03-24
membrane CLI support

### LINKED ISSUE #1955: update CLI docs to include membrane option

atravitz:
Checklist


src/openfecli/commands/plan_rbfe_network.py
```diff
@@ -44,8 +45,9 @@ def plan_rbfe_network_main(
         ligands of the system
     solvent : SolventComponent
         Solvent component used for solvation
-    protein : ProteinComponent
-        protein component for complex simulations, to which the ligands are bound
+    protein : ProteinComponent | ProteinMembraneComponent
+        protein component for complex simulations, to which the ligands are bound.
+        If a ProteinMembraneComponent is used, the
```
hannahbaumann: Just leaving a comment here to remember to finish this sentence.

src/openfecli/parameters/protein.py
```diff
@@ -42,5 +42,30 @@ def _load_protein_from_pdbx(user_input, context):
...
+    ):  # this silences some stderr spam
+        return NOT_PARSED
+
+    from gufe import ProteinMembraneComponent
+
+    try:
+        return ProteinMembraneComponent.from_pdb_file(user_input)
+    except ValueError:
+        try:
+            return ProteinMembraneComponent.from_pdbx_file(user_input)
```
hannahbaumann: This can give confusing error messages. I just tried out loading a protein-only PDB file as a `--protein-membrane`. It gives a `ValueError` because it cannot find the box vectors, but then it tries to read it as pdbx and gives an `IndexError` because it tried to read a pdb as pdbx. Maybe it's better to handle the cases separately?

src/openfecli/commands/plan_rbfe_network.py
```diff
@@ -222,6 +235,7 @@ def plan_rbfe_network(
     write("")
     write(f"\t{n_protocol_repeats=} ({n_protocol_repeats} simulation repeat(s) per transformation)\n")  # fmt: skip
 
+    # protein.validate()
```
hannahbaumann: Yes, a validation is definitely necessary. One could do a validation of the Protocol in the `RBFEAlchemicalNetworkPlanner` after the protocol is created, e.g.
`transformation_protocol.validate(stateA=stateA, stateB=stateB, mapping=ligand_mapping_edge)`.

A test for this would be great as well. Currently, if you pass a system without a membrane that does have a box vector (e.g. the `eg5_protein.pdb` in the eg5 folder) as a ``--protein-membrane``, it would create an AlchemicalNetwork without complaining. This however should fail. Adding the protocol validate will lead to a `ComponentValidationError`.

hannahbaumann (changes requested):
Thanks @atravitz ! I added some comments regarding validation and error handling. 
One thing that's currently missing is that the network planner always adds a `SolventComponent` to the `ChemicalSystem`, however the membrane case does not support a SolventComponent in the complex.
Once you add the `protocol.validate` to the network planner, this would also raise an Error.

src/openfecli/parameters/protein.py
atravitz: yes! apologies, I was working on this locally but hadn't pushed the code yet.

src/openfecli/commands/plan_rbfe_network.py
atravitz: ah, thank you! i wasn't sure where was best to add validation before. I added protein component validation [here](https://github.com/OpenFreeEnergy/openfe/pull/1896/commits/239e3ac4104bd88cc0e67d70bf58917ea3fcd983) as well as  transformation protocol validation [here](https://github.com/OpenFreeEnergy/openfe/pull/1896/commits/5fd87f2e5cb338b7105ca83da4852cbfea76fad1).

src/openfecli/parameters/protein.py
atravitz: I've updated this code now, please let me know if you see any cases I've missed!

docs/mambaf9t4jj8g5ei
atravitz, line 1: (this was an erroneous commit months ago)

src/openfecli/commands/plan_rbfe_network.py
```diff
@@ -216,6 +226,9 @@ def plan_rbfe_network(
     # TODO:  write nice parameter
     write("\tNetwork Generation: " + str(ligand_network_planner))
 
+    write(f"Validating {type(protein_component).__name__}...")
```
hannahbaumann: This is not necessary since the protocol.validate method validates all components.

src/openfecli/tests/commands/test_plan_rbfe_network.py
```diff
@@ -463,3 +463,15 @@ def test_custom_yaml_plan_radial_smoke_test(custom_yaml_radial, eg5_files, tmp_p
...
+def test_plan_rbfe_network_invalid_membrane(eg5_files):
+    """eg5_protein has box vectors but no membrane. ProteinMembraneComponent validation should catch this."""
+
+    runner = CliRunner()
+    args = ["--protein-membrane", eg5_files[0], "-M", eg5_files[1]]
+
+    with runner.isolated_filesystem():
+        result = runner.invoke(plan_rbfe_network, args)
+
+        assert result.exit_code == 1
```
hannahbaumann: Would it make sense to directly test for the `ComponentValidationError` here?

src/openfecli/parameters/protein.py
```diff
@@ -1,46 +1,60 @@
...
+        )
     try:
-        return ProteinComponent.from_pdbx_file(user_input)
-    except ValueError:  # any exception should try other strategies
-        return NOT_PARSED
+        if _contains_any_substring(input_file, _PDB_EXT):
+            return protein_class.from_pdb_file(input_file)
+        elif _contains_any_substring(input_file, _PDBX_EXT):
+            return protein_class.from_pdbx_file(input_file)
+    except ValueError:
```
hannahbaumann: I think here I would catch and print the actual error. E.g. in the protein membrane case where no box vector is provided, the user wouldn't know why it was unable to load the file. The error would have said:
`
Error: Could not determine box_vectors. Please provide them explicitly via the ``box_vectors`` argument or enable ``infer_box_vectors``
`
So maybe something like this?
```
except ValueError as e:
    click.secho(
        f"Unable to load a {protein_class.__name__} from {input_file}.\nError: {e}",
        err=True
    )
```

src/openfe/setup/chemicalsystem_generator/easy_chemicalsystem_generator.py
```diff
@@ -20,9 +21,9 @@
 class EasyChemicalSystemGenerator(AbstractChemicalSystemGenerator):
     def __init__(
         self,
-        solvent: Optional[SolventComponent] = None,
-        protein: Optional[ProteinComponent] = None,
-        cofactors: Optional[Iterable[SmallMoleculeComponent]] = None,
+        solvent: SolventComponent | None = None,
+        protein: ProteinComponent | None = None,
```
hannahbaumann, line 25: In the doc string below you have both ProteinComponent and ProteinMembraneComponent as type hint. Since ProteinMembraneComponent is a subclass of ProteinComponent, you probably don't need to list both here, but it should be consistent with the doc string.

src/openfecli/parameters/protein.py
```diff
@@ -1,46 +1,60 @@
...
+_PDBX_EXT = [".cif", ".pdbx"]
 
 
-def _load_protein_from_pdbx(user_input, context):
-    if not any(
-        [ext in str(user_input) for ext in [".pdb", ".cif", ".pdbx"]]
-    ):  # this silences some stderr spam
-        return NOT_PARSED
+def _contains_any_substring(input: str, substrings: Iterable[str]) -> bool:
+    return any([substring in input for substring in substrings])
```
hannahbaumann, line 16: Would it be safer to check for the file ending? Though of course it could also be gziped, so I'm not sure. I was just thinking in case .pdb could just be part of the file name, without it being a pdb file?

src/openfecli/tests/parameters/test_protein.py
```diff
@@ -1,24 +1,66 @@
...
+
+def test_load_membrane_with_protein_only(capsys):
+    with resources.as_file(resources.files("gufe.tests.data")) as d:
+        filename = str(d / "181l.pdb")
+
+    with pytest.raises(SystemExit) as e:
+        _ = _get_protein_membrane(filename, None)
+
+    captured = capsys.readouterr()
+    assert "Unable to load a ProteinMembraneComponent" in captured.err
```
hannahbaumann: If you add the actual Error to this, I would here also check for the box vector thing in the error.

hannahbaumann (changes requested):
Thanks @atravitz , this looks great! Left some small comments, the only important one is the catching of the actual error which would be very helpful to users I think.

src/openfe/setup/chemicalsystem_generator/easy_chemicalsystem_generator.py
atravitz: I was thinking that explicitly including `ProteinMembraneComponent` would make it clearer what the code is doing. Now that I revisit it, I agree that we should drop it. Done!

src/openfecli/parameters/protein.py
```diff
@@ -1,46 +1,60 @@
...
-        return ProteinComponent.from_pdbx_file(user_input)
-    except ValueError:  # any exception should try other strategies
-        return NOT_PARSED
+        if _contains_any_substring(input_file, _PDB_EXT):
+            return protein_class.from_pdb_file(input_file)
+        elif _contains_any_substring(input_file, _PDBX_EXT):
+            return protein_class.from_pdbx_file(input_file)
+    except ValueError:
+        click.secho(f"Unable to load a {protein_class.__name__} from {input_file}.", err=True)
+        sys.exit(1)
```
mikemhenry: If you use a `click.BadParameter` exception here it will be formatted nicely for a user and also then handle giving us a non-zero exit code
```suggestion
    except ValueError as e:
        raise click.BadParameter(
            f"unable to parse {protein_class.__name__} from {click.format_filename(path)}: {e}",
            param_hint=param_name,
        ) from e
```

src/openfecli/commands/plan_rbfe_network.py
```diff
@@ -187,9 +188,18 @@ def plan_rbfe_network(
...
+    elif protein:
+        protein_component = PROTEIN.get(protein)
+        write("\t\tProteinComponent: " + str(protein_component))
+    elif protein_membrane:
+        protein_component = PROTEIN_MEMBRANE.get(protein_membrane)
+        write("\t\tProteinMembraneComponent: " + str(protein_component))
+    else:
+        raise ValueError(
+            "Either --protein or --protein-membrane must be provided. Run ``openfe plan-rbfe-network -h`` for more information."
+        )
```
mikemhenry: If you use a `click.UsageError` it will automatically include a hint to check `--help`
```suggestion
        raise click.UsageError(
            "Either --protein or --protein-membrane must be provided.
        )
```

mikemhenry:
We can wait for the dust to settle but I think we should also include a news item here so it shows up in the change log

src/openfecli/parameters/protein.py
atravitz: notes from conversation with @hannahbaumann and @IAlibay offline - we'll leave the checks as-is for now, but improve the error message.

src/openfecli/tests/commands/test_plan_rbfe_network.py
atravitz: it will be raised as a BadParameterError by click, but pass through the message:
 

> Could not determine box_vectors. Please provide them explicitly via the ``box_vectors`` argument or enable ``infer_box_vectors``

which I now check for!

src/openfecli/parameters/protein.py
atravitz: I took mike's suggestion to use click's BadParameter error, which achieves this.

src/openfecli/tests/parameters/test_protein.py
atravitz: done!

mikemhenry:
Maybe add a link in the news item to the membrane prep tutorial?

src/openfecli/tests/commands/test_plan_rbfe_network.py
```diff
@@ -463,3 +464,34 @@ def test_custom_yaml_plan_radial_smoke_test(custom_yaml_radial, eg5_files, tmp_p
...
+    args = ["-M", eg5_files[1]]
+
+    runner = CliRunner(catch_exceptions=False)
+    result = runner.invoke(plan_rbfe_network, args)
+    assert result.exit_code == 2
+    assert "Either --protein or --protein-membrane must be provided." in result.stderr
+
+
+def test_plan_rbfe_too_many_protein_error(eg5_files):
+    """eg5_protein has box vectors but no membrane. ProteinMembraneComponent validation should catch this."""
```
hannahbaumann: Doc string needs updating.

hannahbaumann (commented):
Thanks @atravitz , this looks good!
The only thing I just realized is that if the user provides a fully solvated protein-membrane system using the `--protein` flag, it will create the transformations without an error. This is also not handled in the API, but I wonder if it's even more dangerous on the CLI level. One could e.g. also add a density check on the `ProteinComponent` and if it's above a threshold, but not a `SolvatePDBComponent`, at least add a warning? Though people probably don't look at warnings since there are so many. This is probably something that should be handled elsewhere but still wanted to raise it here. Do you have any thoughts @IAlibay ?

IAlibay:
> The only thing I just realized is that if the user provides a fully solvated protein-membrane system using the `--protein` flag, it will create the transformations without an error. This is also not handled in the API, but I wonder if it's even more dangerous on the CLI level. One could e.g. also add a density check on the `ProteinComponent` and if it's above a threshold, but not a `SolvatePDBComponent`, at least add a warning? Though people probably don't look at warnings since there are so many. This is probably something that should be handled elsewhere but still wanted to raise it here. Do you have any thoughts @IAlibay ?

Couple of takes here (already communicated in earlier meeting, just writing it down here):
- It's probably ok to go with this release without that check.
- Adding some check for the component that sees if you have both box vectors and high density would be good. My initial take is that a warning would be better, but I do agree that we just have too many warnings. We could either: 1) give validate an optional "error on warning" that is used by more sensitive tooling like the CLI, or 2) just go with the error and see how many folks it affects.

hannahbaumann (approved):
Approving this PR to not be blocking the release. I think only the doc string needs an update but since it's in the test it's also not that important.

atravitz:
thanks @hannahbaumann! You make a good point, so I opened an issue [here](https://github.com/OpenFreeEnergy/openfe/issues/1964) to improve validation.

PR #1911, opened 2026-04-01
Use set operations for unique atom detection in HybridTopologyFactory

kokishimada:
## Summary
- Replace linear-scan loops with set difference operations in `HybridTopologyFactory._set_mappings` for finding unique atoms
- Use dict comprehension for `_env_old_to_new_map` construction

## Test plan
- [ ] Existing tests for `HybridTopologyFactory` pass
- [ ] Verify `_unique_old_atoms` and `_unique_new_atoms` produce the same sorted results as before


IAlibay:
Hi @kokishimada - thanks for opening this PR, this is very interesting.

That being said, there is already some work towards speeding up the HTF being done here: https://github.com/OpenFreeEnergy/openfe/pull/1858

It's possible that we would go for that PR instead, but I will let @jthorton weigh in.

PR #1948, opened 2026-04-20
Switch repeat id assignment from leg to repeat in AFE Protocols

IAlibay:
This PR changes things so that a single uuid is used for each repeat in a Protocol, rather than each leg. This more strictly adheres to the idea of being a "repeat_id". It is also more intuitive since it allows us to assume that the uuid represents a single DAG execution.

We may have to revisit this formalism if/when we have a case where we have multiple indentical units being executed in a single DAG.

Checklist

* [x] [GPU integration tests](https://github.com/OpenFreeEnergy/openfe/actions/workflows/aws-gpu-integration-tests.yaml) - ran locally


IAlibay:
pre-commit.ci autofix

atravitz:
what is the motivation for making this change now? Is the idea that more robust bookkeeping is possible if we can match up legs with identical uuids?

IAlibay:
> what is the motivation for making this change now? Is the idea that more robust bookkeeping is possible if we can match up legs with identical uuids?

Couple of things:

1) The old behaviour is a bug - all our protocols have a "repeat_id" but we've been using it inconsistently and here in AFE Protocols as a "leg_id". 
2) I've made SepTop do a similar behaviour where it's one id per repeat rather than per leg. So it would be good to align the two protocols.
3) Yes it would generally be good to be able to rely on the repeat id consistently for bookkeeping. Right now it's inconsistent and can't be used.

atravitz (approved):

PR #1978, opened 2026-05-18
Support for charge changes in the HybTop protocol for explicitly solvated systems

### LINKED ISSUE #1976: HybridTopologyProtocol does not allow for net charge transformation when running with a membrane
Originally reported by @hannahbaumann 

We unfortunately forgot to update the validation checks to account for not having a SolventComponent but some other kind of BaseSolventComponent. This leads to the validation checks failing if you have a net charge with a ProteinMembraneComponent.

hannahbaumann:
This is just a first idea (and the `_get_ion_resnames_from_topology` was written together with Claude). This is mostly just to show the different things we have to consider, but I don't like the dictionary idea of known ions yet. Maybe it would also be ok to always use Cl and Na ions for explicitly solvated systems, no matter if other ion types were used as counterions?

Checklist

### DIFF
--- a/src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
+++ b/src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
@@ -166,7 +166,7 @@ def _get_ion_parameters(
             continue
         charge, _, _ = nbf.getParticleParameters(atom.index)
         charge_val = charge.value_in_unit(omm_unit.elementary_charge)
-        if (abs(abs(charge_val) - 1.0) < 0.01 and np.sign(charge_val) == desired_sign):
+        if np.isclose(abs(charge_val), 1.0, atol=0.01) and np.sign(charge_val) == desired_sign:
             ion_counts[residue.name] += 1
             if residue.name not in ion_atom_indices:
                 ion_atom_indices[residue.name] = atom.index
--- a/src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
+++ b/src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
@@ -62,90 +62,55 @@ def _get_ion_parameters_from_forcefield(
 
     return nbf.getParticleParameters(0)
 
-
-def _get_ion_and_water_parameters(
+def _get_water_parameters(
     topology: app.Topology,
     system: System,
-    forcefield: app.ForceField,
-    charge_difference: int,
     water_resname: str = 'HOH',
 ) -> tuple:
     """
-    Get ion NonbondedForce parameters and water charges.
-
-    Scans the topology for the most abundant monovalent ion of the appropriate
-    charge sign, identified by single-atom residues with a partial charge
-    close to +/- 1.
-    If no matching ion is found, falls back to NA or CL.
+    Get water oxygen and hydrogen partial charges from the system.
 
     Parameters
     ----------
     topology : app.Topology
-        The topology to search.
-    system : openmm.System
-        The system associated with the topology.
-    forcefield : app.ForceField
-        Used as a fallback to obtain ion parameters when no matching ion is
-        present in the topology.
-    charge_difference : int
-        Charge difference between state A and state B, used to determine the
-        required ion charge sign.
+      The topology to search for water atoms.
+    system : app.System
+      The system associated with the input topology object.
     water_resname : str
-        Residue name of water. Default 'HOH'.
+      The residue name of the water. Default 'HOH'.
 
     Returns
     -------
-    ion_charge, ion_sigma, ion_epsilon : openmm.unit.Quantity
-        NonbondedForce parameters for the ion.
-    o_charge, h_charge : openmm.unit.Quantity
-        Partial charges of the water oxygen and hydrogen.
+    o_charge : openmm.unit.Quantity
+      The partial charge of the water oxygen.
+    h_charge : openmm.unit.Quantity
+      The partial charge of the water hydrogen.
+
+    Raises
+    ------
+    ValueError
+      If no water residue named ``water_resname`` with O and H atoms
+      is found in the topology.
     """
     nbf = [i for i in system.getForces() if isinstance(i, NonbondedForce)][0]
-
-    desired_sign = np.sign(charge_difference)
-
-    ion_counts = Counter()
-    ion_atom_indices = {}
 
     o_charge = None
     h_charge = None
 
     for residue in topology.residues():
-        atoms = list(residue.atoms())
-
-        # water parameters
-        if o_charge is None and residue.name == water_resname:
-            for atom in atoms:
-                charge, _, _ = nbf.getParticleParameters(atom.index)
-                if atom.element.symbol == 'O':
-                    o_charge = charge
-                elif atom.element.symbol == 'H':
-                    h_charge = charge
+        if residue.name != water_resname:
             continue
-
-        # ion candidates
-        if len(atoms) != 1:
-            continue
-
-        atom = atoms[0]
-        if atom.element is None:
-            continue
-
-        charge, _, _ = nbf.getParticleParameters(atom.index)
-        charge_val = charge.value_in_unit(omm_unit.elementary_charge)
-
-        # Only monovalent ions for now
-        if not np.isclose(abs(charge_val), 1.0, atol=0.2):
-            continue
-
-        if np.sign(charge_val) != desired_sign:
-            continue
-
-        ion_counts[residue.name] += 1
-        ion_atom_indices.setdefault(
-            residue.name,
-            atom.index,
-        )
+        for atom in residue.atoms():
+            if atom.element is None:
+                continue
+            charge, _, _ = nbf.getParticleParameters(atom.index)
+            if atom.element.symbol == 'O':
+                o_charge = charge
+            elif atom.element.symbol == 'H':
+                h_charge = charge
+        # Stop as soon as we have both charges
+        if o_charge is not None and h_charge is not None:
+            break
 
     if o_charge is None or h_charge is None:
         raise ValueError(
@@ -154,36 +119,82 @@ def _get_ion_and_water_parameters(
             "alchemical charge correction."
         )
 
+    return o_charge, h_charge
+
+def _get_ion_parameters(
+    topology: app.Topology,
+    system: System,
+    charge_difference: int,
+    forcefield: app.ForceField,
+) -> tuple:
+    """
+    Get NonbondedForce parameters for the most abundant monovalent ion of
+    the appropriate charge sign found in the topology. Falls back to
+    'NA' or 'CL' parameters if no ion is found in the topology.
+
+    Parameters
+    ----------
+    topology : app.Topology
+      The topology to search for the most abundant ion type.
+    system : app.System
+      The system associated with the input topology object.
+    charge_difference : int
+      The charge difference between state A and state B. Determines
+      whether to look for a positive or negative ion.
+    forcefield : app.ForceField
+      The force field to use for parameterization if no ion is found
+      in the topology.
+
+    Returns
+    -------
+    ion_charge : openmm.unit.Quantity
+    ion_sigma : openmm.unit.Quantity
+    ion_epsilon : openmm.unit.Quantity
+    """
+    nbf = [i for i in system.getForces() if isinstance(i, NonbondedForce)][0]
+
+    desired_sign = np.sign(charge_difference)
+    ion_counts: Counter = Counter()
+    ion_atom_indices: dict[str, int] = {}
+
+    for residue in topology.residues():
+        atoms = list(residue.atoms())
+        if len(atoms) != 1:
+            continue
+        atom = atoms[0]
+        if atom.element is None:
+            continue
+        charge, _, _ = nbf.getParticleParameters(atom.index)
+        charge_val = charge.value_in_unit(omm_unit.elementary_charge)
+        if (abs(abs(charge_val) - 1.0) < 0.01 and np.sign(charge_val) == desired_sign):
+            ion_counts[residue.name] += 1
+            if residue.name not in ion_atom_indices:
+                ion_atom_indices[residue.name] = atom.index
+
     if ion_counts:
         # Use the most abundant matching ion found in the topology
         best_resname = ion_counts.most_common(1)[0][0]
-        ion_atom_index = ion_atom_indices[best_resname]
-        ion_charge, ion_sigma, ion_epsilon = nbf.getParticleParameters(
-            ion_atom_index
-        )
+        return nbf.getParticleParameters(ion_atom_indices[best_resname])
+
+    # No matching ion in topology: fall back to NA/CL from forcefield
+    if charge_difference > 0:
+        fallback_resname = 'NA'
+        fallback_element = app.Element.getByAtomicNumber(11)  # Na
     else:
-        # No matching ion in topology — fall back to NA/CL from forcefield
-        if charge_difference > 0:
-            fallback_resname = 'NA'
-            fallback_element = app.Element.getByAtomicNumber(11)  # Na
-        else:
-            fallback_resname = 'CL'
-            fallback_element = app.Element.getByAtomicNumber(17)  # Cl
+        fallback_resname = 'CL'
+        fallback_element = app.Element.getByAtomicNumber(17)  # Cl
 
-        wmsg = (
-            f"No monovalent ion with the appropriate charge sign found in "
-            f"the topology. Defaulting to '{fallback_resname}' and obtaining "
-            f"parameters from the forcefield directly."
-        )
-        warnings.warn(wmsg)
-        logger.warning(wmsg)
+    wmsg = (
+        f"No monovalent ion with the appropriate charge sign found in "
+        f"the topology. Defaulting to '{fallback_resname}' and obtaining "
+        f"parameters from the forcefield directly."
+    )
+    warnings.warn(wmsg)
+    logger.warning(wmsg)
 
-        ion_charge, ion_sigma, ion_epsilon = _get_ion_parameters_from_forcefield(
-            forcefield, fallback_resname, fallback_element
-        )
-
-    return ion_charge, ion_sigma, ion_epsilon, o_charge, h_charge
-
+    return _get_ion_parameters_from_forcefield(
+        forcefield, fallback_resname, fallback_element,
+    )
 
 def _fix_alchemical_water_atom_mapping(
     system_mapping: dict[str, Union[dict[int, int], list[int]]],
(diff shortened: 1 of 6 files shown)


src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -720,3 +724,60 @@ def set_and_check_new_positions(mapping, old_topology, new_topology,
             logging.warning(wmsg)
 
     return new_pos_array * omm_unit.angstrom
+
+
+def _get_ion_resnames_from_topology(topology: app.Topology) -> tuple[str, str]:
+    """
+    Infer positive and negative ion residue names from a topology by
+    finding the most common monovalent ion of each charge type.
+    Falls back to NA/CL if none are found.
```
hannahbaumann: I'm doing this mostly because e.g. maestro doesn't always add ions (in the a2a example there were no negative ions in the system).

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -123,10 +124,14 @@ def _fix_alchemical_water_atom_mapping(
...
 def handle_alchemical_waters(
-    water_resids: list[int], topology: app.Topology,
-    system: System, system_mapping: dict,
+    water_resids: list[int],
+    topology: app.Topology,
+    system: System,
+    system_mapping: dict,
     charge_difference: int,
-    solvent_component: SolventComponent,
+    positive_ion_resname: str,
```
hannahbaumann: use element or formal charge instead?

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
```diff
@@ -435,13 +437,22 @@ def _validate_charge_difference(
...
             raise ValueError(errmsg)
 
-        ion = {-1: solvent_component.positive_ion, 1: solvent_component.negative_ion}[difference]
-
-        wmsg = (
-            f"A charge difference of {difference} is observed "
-            "between the end states. This will be addressed by "
-            f"transforming a water into a {ion} ion"
-        )
+        # resolve ion names from SolventComponent or topology
```
IAlibay: TBD: It's probably ok squash this into a single check that just says we'll convert to an ion of opposite charge.

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -21,77 +22,179 @@
...
+            ion_counts[residue.name] += 1
+            if residue.name not in ion_atom_indices:
+                ion_atom_indices[residue.name] = atom.index
+
+    if ion_counts:
+        # Use the most abundant matching ion found in the topology
+        best_resname = ion_counts.most_common(1)[0][0]
+        return nbf.getParticleParameters(ion_atom_indices[best_resname])
+
+    # No matching ion in topology: fall back to NA/CL from forcefield
```
IAlibay, line 126: TBD: how does this end up happening? I.e. it is purely a case where your user didn't add any counterions or did a strict neutralization?

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -21,77 +22,179 @@
...
+    for residue in topology.residues():
+        atoms = list(residue.atoms())
+        if len(atoms) != 1:
+            continue
+        atom = atoms[0]
+        if atom.element is None:
+            continue
+        charge, _, _ = nbf.getParticleParameters(atom.index)
+        charge_val = charge.value_in_unit(omm_unit.elementary_charge)
+        if (abs(abs(charge_val) - 1.0) < 0.01 and np.sign(charge_val) == desired_sign):
```
IAlibay: Maybe np.isclose would be simpler.

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: Yes, in the a2a case this was what the maestro prepped membrane from Ross et al. looked like, no counter ions, just a structural Na in the protein, but no ions in the solvent.

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -171,18 +273,13 @@ def handle_alchemical_waters(
...
+    if charge_difference == 0:
         return None
 
-    ion_charge, ion_sigma, ion_epsilon, o_charge, h_charge = _get_ion_and_water_parameters(
-        topology, system, ion_resname,
-        'HOH',  # Modeller always adds HOH waters
+    ion_charge, ion_sigma, ion_epsilon = _get_ion_parameters(
+        topology, system, charge_difference, forcefield,
     )
+    o_charge, h_charge = _get_water_parameters(topology, system, water_resname)
```
jthorton: Could we just use the single ion system trick to get the parameters for the water now that we have the forcefield? This would avoid the need to know the water resname and looping over the full system to find a water molecule?

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -143,14 +249,10 @@ def handle_alchemical_waters(
       A dictionary of system mappings between the stateA and stateB systems
     charge_difference : int
       The charge difference between state A and state B.
```
jthorton: I would make it clear in all of the doc strings for charge_difference how this should be calculated, stateA - stateB

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -21,77 +22,179 @@
...
+    Returns
+    -------
+    ion_charge : openmm.unit.Quantity
+        The partial charge of the ion.
+    ion_sigma : openmm.unit.Quantity
+        The NonbondedForce sigma parameter of the ion.
+    ion_epsilon : openmm.unit.Quantity
+        The NonbondedForce epsilon parameter of the ion.
+    """
+    dummy_top = app.Topology()
```
IAlibay: This could help it make this re-usable for both ions and water
```
def get_topology(smiles):
    from openff.toolkit import Molecule, Topology
    offmol = Molecule.from_smiles(smiles)
    omm_top = m.to_topology().to_openmm()
```

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: I changed this to also use the ff to get the water parameters. But I had to move up the ValueError to check for the 3point water model since I was otherwise getting an error when assigning the ff from the smiles. Not sure if there's a way around that.

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: Added this!

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: I ended up putting this into the `_get_ff_parameters` that can be used for both waters and ions, please let me know what you think!

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
jthorton: Is the error due to not matching a template if you have an extra site water model force field? If so we just need to make a modeller from the topology and add the extra particles, though, since we don't currently support using waters with extra sites in charge change transforms we can probably leave this as it is and tackle it later.

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
```diff
@@ -2073,7 +2128,7 @@ def test_handle_alchemical_wats(
         system=system,
         system_mapping=benzene_self_system_mapping,
         charge_difference=1,
```
jthorton: maybe add a parameterise flag and check for positive and negative charge differences and check that the charge is adjusted accordingly?

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
```diff
@@ -1932,28 +1962,53 @@ def benzene_self_system_mapping(benzene_solvent_openmm_system):
...
-@pytest.mark.parametrize(
-    "ion, water",
-    [
-        ["NA", "SOL"],
-        ["NX", "WAT"],
-    ],
-)
-def test_get_ion_water_parameters_unknownresname(ion, water, benzene_solvent_openmm_system):
-    system, topology, positions = benzene_solvent_openmm_system
+def test_get_ion_parameters_fallback_to_forcefield(benzene_modifications):
```
jthorton: Check for both +/- charge differences and that the correct fallback ion is chosen.

jthorton (commented):
Two minor things to add to the tests then I think this is good to go!

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
```diff
@@ -1215,6 +1215,36 @@ def test_dry_run_membrane_complex(
...
+    mapping = openfe.LigandAtomMapping(
+        componentA=ligA,
+        componentB=ligB,
+        componentA_to_componentB={i: i for i in range(36)},
+    )
+
+    settings = openmm_rfe.RelativeHybridTopologyProtocol.default_settings()
+
+    # Mock get_alchemical_charge_difference to return non-zero
+    with mock.patch.object(
```
IAlibay: rather than patching - how about just using any 2 ligands with different net charge directly?

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -21,30 +22,68 @@
...
 logger = logging.getLogger(__name__)
 
 
-def _get_ion_and_water_parameters(
+def _get_ff_parameters(
+    forcefield: app.ForceField,
+    smiles: str,
+) -> dict[str, tuple]:
+    """
+    Get NonbondedForce parameters for all atoms in a molecule by creating
```
IAlibay, line 33: > all atoms

The way your list comprehension works, it willl only get you the unique elements in the molecule. So for example, the two hydrogens in the water will get squashed to a single entry. That's fine for a symmetric molecule, but worth noting.

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -21,30 +22,68 @@
...
-from openfe import SolventComponent
-
 logger = logging.getLogger(__name__)
 
 
-def _get_ion_and_water_parameters(
+def _get_ff_parameters(
+    forcefield: app.ForceField,
+    smiles: str,
+) -> dict[str, tuple]:
```
IAlibay:
```suggestion
) -> dict[str, tuple[float, float, float]]:
```
?

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -21,30 +22,68 @@
...
+    ----------
+    forcefield : app.ForceField
+      The force field to use for parameterisation.
+    smiles : str
+      The SMILES string of the molecule to parameterise
+      (e.g. '[Na+]', '[Cl-]', 'O').
+
+    Returns
+    -------
+    dict[str, tuple]
```
IAlibay: As above re: being more explicit with the tuple return.

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -54,44 +93,48 @@ def _get_ion_and_water_parameters(
...
+    desired_sign = np.sign(charge_difference)
+    ion_counts: Counter = Counter()
+    ion_atom_indices: dict[str, int] = {}
+
+    for residue in topology.residues():
+        atoms = list(residue.atoms())
+        if len(atoms) != 1:
+            continue
+        atom = atoms[0]
+        if atom.element is None:
```
IAlibay: Is this necessary? I don't see you use the element anywhere after.

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
```diff
@@ -435,12 +437,10 @@ def _validate_charge_difference(
             )
             raise ValueError(errmsg)
 
-        ion = {-1: solvent_component.positive_ion, 1: solvent_component.negative_ion}[difference]
-
         wmsg = (
             f"A charge difference of {difference} is observed "
             "between the end states. This will be addressed by "
-            f"transforming a water into a {ion} ion"
+            "transforming a water into an ion of opposite charge."
```
IAlibay:
```suggestion
            "transforming a water into an ion of the same charge."
```
It's counterintuitive on paper, but if you have a difference of -1 you add a Cl- in the other state, etc...

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -54,44 +93,48 @@ def _get_ion_and_water_parameters(
...
+    else:
+        fallback_smiles = '[Cl-]'
 
-    return ion_charge, ion_sigma, ion_epsilon, o_charge, h_charge
+    wmsg = (
+        f"No monovalent ion with the appropriate charge sign found in "
+        f"the topology. Defaulting to '{fallback_smiles}' and obtaining "
+        f"parameters from the forcefield directly."
+    )
+    warnings.warn(wmsg)
```
IAlibay: [nit] I know we're not consistent with this, but it's probably ok to just do logger.warning given it's something that happens during execution.

IAlibay (changes requested):
Just going to put a request changes on the main warning being emitted in the protocol, everything else is small things. Overall looks great though!

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: Yes, I think maybe we can tackle this later, or what do you think @IAlibay ?

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
hannahbaumann: Added this!

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
hannahbaumann: Added this!

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: Added a note in the doc string!

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
hannahbaumann: True, this was left over from a previous approach where I was using the elements. removed this.

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
IAlibay: This is definitely a "later" problem - we'll need to rework a bunch of this when we have to consider doing net charge transformations with non-3 site waters.

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -21,30 +22,73 @@
...
+    offmol = Molecule.from_smiles(smiles)
+    dummy_top = offmol.to_topology().to_openmm()
+    dummy_system = forcefield.createSystem(dummy_top)
+    nbf = [f for f in dummy_system.getForces() if isinstance(f, NonbondedForce)][0]
+
+    return {
+        atom.element.symbol: nbf.getParticleParameters(atom.index)
+        for atom in dummy_top.atoms()
+    }
+
```
IAlibay, line 66:
```suggestion


```
not sure why linters aren't picking this up

src/openfe/protocols/openmm_rfe/_rfe_utils/topologyhelpers.py
```diff
@@ -21,30 +22,73 @@
...
+def _get_ion_parameters(
     topology: app.Topology,
     system: System,
-    ion_resname: str,
-    water_resname: str = 'HOH',
-):
+    charge_difference: int,
+    forcefield: app.ForceField,
+) -> tuple:
```
IAlibay:
```suggestion
) -> tuple[float, float, float]:
```
[nit] Where possible, it's better to be as explicit as possible about type returns.

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_validation.py
```diff
@@ -441,11 +441,10 @@ def test_get_charge_difference(mapping_name, result, request, caplog):
     mapping = request.getfixturevalue(mapping_name)
     caplog.set_level(logging.INFO)
 
-    ion = r"Na+" if result == -1 else r"Cl-"
     msg = (
         f"A charge difference of {result} is observed "
         "between the end states. This will be addressed by "
-        f"transforming a water into a {ion} ion"
+        f"transforming a water into an ion of opposite charge"
```
IAlibay:
```suggestion
        f"transforming a water into an ion of the same charge"
```

IAlibay (approved):
LGTM! Only a few tiny things (one fix for the failling test).

jthorton (approved):
LGTM great job!

hannahbaumann:
@IAlibay Do we want to wait with merging this until we have benchmarked this on a real system, or do we already want to put it in main?

IAlibay:
> @IAlibay Do we want to wait with merging this until we have benchmarked this on a real system, or do we already want to put it in main?

It's fine to merge now, we just don't want to release until we've done some tests.

PR #1982, opened 2026-05-28
Add structural analysis to the SepTop protocol

hannahbaumann:
Checklist


hannahbaumann:
pre-commit.ci autofix

hannahbaumann:
This is what it would look like for the complex right now.

protein_2D_RMSD

ligand_B_RMSD

ligand_B_COM_drift
ligand_A_RMSD
ligand_A_COM_drift

hannahbaumann:
pre-commit.ci autofix

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1579,14 +1904,31 @@ def _execute(
         trajectory = simulation.outputs["trajectory"]
         checkpoint = simulation.outputs["checkpoint"]
 
+        alchem_comps = self._inputs["alchemical_components"]
+        rdmol_A = alchem_comps["stateA"][0].to_rdkit()
+        rdmol_B = alchem_comps["stateB"][0].to_rdkit()
```
IAlibay: In the spirit of "Protocols taking in components" - it might be good to do that inside of `run`.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1579,14 +1904,31 @@ def _execute(
...
+            pdb_file=setup.outputs["subsampled_pdb_structure"],
+            ligand_A_indices=ligand_A_indices,
+            ligand_B_indices=ligand_B_indices,
+            rdmol_A=rdmol_A,
+            rdmol_B=rdmol_B,
             scratch_basepath=ctx.scratch,
             shared_basepath=ctx.shared,
         )
 
-        # We re-include things here to make life easier when gathering results
```
IAlibay, line 1589: Might be good to keep this around.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1560,7 +1869,23 @@ def run(
...
 
-        return energy_analysis
+        if self.verbose:
+            self.logger.info("Analyzing structural outputs")
+
+        structural_analysis = self._structural_analysis(
+            pdb_file=pdb_file,
+            trj_file=trajectory,
+            output_directory=self.shared_basepath,
+            dry=dry,
```
IAlibay: [nit] Purely a preference - it's usually good to order `dry` towards the end or the start of the argument list (I tend to go with the end).

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,302 @@ def _analyze_multistate_energies(
...
+
+                if data.get(f"{label}_COM_drift"):
+                    fig = plotting.plot_ligand_COM_drift(time_ps, data[f"{label}_COM_drift"])
+                    fig.savefig(output_directory / f"{label}_COM_drift.png")
+                    plt.close(fig)
+
+        # Write out an NPZ with all the relevant analysis data
+        npz_file = output_directory / "structural_analysis.npz"
+        if simtype == "complex":
+            np.savez_compressed(
```
IAlibay: to avoid lots of branches, could this be done higher above, just after where you do the `run_X_analysis`?

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,302 @@ def _analyze_multistate_energies(
...
+                        ds,
+                        pdb_file,
+                        skip,
+                        ligand_A_indices,
+                        ligand_B_indices,
+                        rdmol_A,
+                        rdmol_B,
+                    )
+
+        except Exception as e:
```
IAlibay, line 1769: What Exception are we covering here? Would it be better if we did this in the `run_X_analysis` methods directly?

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,302 @@ def _analyze_multistate_energies(
...
+            "ligand_B_RMSD": [],
+            "ligand_A_COM_drift": [],
+            "ligand_B_COM_drift": [],
+            "protein_2D_RMSD": [],
+            "time_ps": None,
+        }
+        u_top = mda.Universe(pdb_file)
+        for state_idx in range(n_lambda):
+            universe = create_universe_single_state(u_top._topology, ds, state=state_idx)
+            prot = universe.select_atoms(protein_selection)
```
jthorton: Not sure if its possible or faster but could we do this selection once on the u_top to get the indices of the atoms and then just indexing to get the atoms in each loop?

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,302 @@ def _analyze_multistate_energies(
...
+        return data
+
+    @classmethod
+    def _structural_analysis(
+        cls,
+        pdb_file: pathlib.Path,
+        trj_file: pathlib.Path,
+        output_directory: pathlib.Path,
+        dry: bool,
+        simtype: str,
```
jthorton: Use a literal type here with `complex` and `solvent`

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,302 @@ def _analyze_multistate_energies(
...
+        import netCDF4 as nc
+        from openfe_analysis.utils import plotting
+
+        try:
+            with nc.Dataset(trj_file) as ds:
+                if hasattr(ds, "PositionInterval"):
+                    n_frames = len(range(0, ds.dimensions["iteration"].size, ds.PositionInterval))
+                else:
+                    n_frames = ds.dimensions["iteration"].size
+                skip = max(n_frames // 500, 1)
```
jthorton: Is there a reason why we want to keep the analysis to only 500 frames, should this be exposed?

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,302 @@ def _analyze_multistate_energies(
...
+                protein_2D_RMSD=np.asarray(data["protein_2D_RMSD"], dtype=np.float32),
+                time_ps=np.asarray(data["time_ps"], dtype=np.float32),
+            )
+        else:
+            np.savez_compressed(
+                npz_file,
+                ligand_A_RMSD=np.asarray(data["ligand_A_RMSD"], dtype=np.float32),
+                ligand_B_RMSD=np.asarray(data["ligand_B_RMSD"], dtype=np.float32),
+                time_ps=np.asarray(data["time_ps"], dtype=np.float32),
+            )
```
jthorton: What about building a dict of shared data that will be saved in both cases so the `time_ps` and `ligand_A/B_RMSD` data and then if the `simtype=="complex"` then add the extra data to the dict then you can have single `np.savez_compressed` call?

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,302 @@ def _analyze_multistate_energies(
...
         self,
         *,
         trajectory: pathlib.Path,
         checkpoint: pathlib.Path,
+        pdb_file: pathlib.Path,
+        ligand_A_indices: list[int],
+        ligand_B_indices: list[int],
+        rdmol_A: Chem.Mol,
+        rdmol_B: Chem.Mol,
+        protein_selection: str = "protein and name CA",
```
jthorton: Do you want to expose this to users, and does it make sense to remove the `protein_selection` default from the private functions so that if you do update the default, you only have to do it in a single place, i.e single source of truth on the public function.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1579,14 +1904,31 @@ def _execute(
...
+        rdmol_B = alchem_comps["stateB"][0].to_rdkit()
+
+        # Remap ligand indices from full system to subsampled system
+        selection_indices = np.array(setup.outputs["selection_indices"])
+        ligand_A_indices = np.where(np.isin(selection_indices, setup.outputs["ligand_A_indices"]))[
+            0
+        ].tolist()
+        ligand_B_indices = np.where(np.isin(selection_indices, setup.outputs["ligand_B_indices"]))[
+            0
+        ].tolist()
```
jthorton: I got lost trying to follow this through but could this go wrong if the user accidentally changes the settings to only save the protein and water by mistake?

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Removed the default from the private functions.

I like the idea of exposing it to the user, but I'm not sure where yet. Would it make sense to add it to the `MultiStateOutputSettings`? Or should we create a new `MultistateAnalysisSettings` class? I think this would also be nice for the HybTop protocol, in case someone runs host-guest systems or similar.

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: So I think in that case it would return an empty list, np.isin would return only false, so np.where wouldn't find anything.
I added a warning at the beginning of _structural_analysis that would give a meaningful structural_analysis_error. Does that make sense?

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: That's a great idea, updated this!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Good point, I exposed it!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Thanks, fixed!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: I updated this, not sure if it's faster, but definitely makes sense!

IAlibay (commented):
Some initial review comments from last week (sorry about the delay, I'll try to complete the review today).

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
...
+            "ligand_B_RMSD": [],
+            "ligand_A_COM_drift": [],
+            "ligand_B_COM_drift": [],
+            "protein_2D_RMSD": [],
+            "time_ps": None,
+        }
+        u_top = mda.Universe(pdb_file)
+        prot_indices = u_top.select_atoms(protein_selection).indices
+        for state_idx in range(n_lambda):
+            universe = create_universe_single_state(u_top._topology, ds, state=state_idx)
```
IAlibay, line 1579: Not for this PR but a reminder to myself - I should make a proposal in MDAnalysis to make it easier to do this without having to use the private attribute.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
         reporter.close()
         return analyzer.unit_results_dict
 
+    @classmethod
```
IAlibay: Should this be a classmethod? It doesn't look like you're using `cls` anywhere.
Should it not be a staticmethod?

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
...
+        """
+        from openfe_analysis.rmsd import (
+            LigandCOMDrift,
+            Protein2DRMSD,
+            SymmetryCorrectedLigandRMSD,
+        )
+        from openfe_analysis.utils.apply_transformations import (
+            apply_complex_alignment_transformations,
+        )
+        from openfe_analysis.utils.universe_utils import create_universe_single_state
```
IAlibay: It's probably ok to have these imports be at the top level, we're not exactly worried about import cost at this point.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
...
+
+        n_lambda = ds.dimensions["state"].size
+        data: dict[str, Any] = {
```
IAlibay: [nit] typing here could be tigether since we know it's `list[np.ndarray] | float | None`. Ok to leave as-is if we expect it to grow though.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
...
+            "ligand_A_RMSD": [],
+            "ligand_B_RMSD": [],
+            "ligand_A_COM_drift": [],
+            "ligand_B_COM_drift": [],
+            "protein_2D_RMSD": [],
+            "time_ps": None,
+        }
+        u_top = mda.Universe(pdb_file)
```
IAlibay, line 1576: Can you add a comment to mention what `u_top` is intended for? You and I know what it is right now, but in 6 months that won't be as obvious.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
...
+
+        if len(ligand_A_indices) == 0 or len(ligand_B_indices) == 0:
+            errmsg = (
+                "No ligand atoms found in the subsampled trajectory. "
+                "This likely means the output_indices selection does not include "
+                "the ligands. Check the output_indices setting in your protocol "
+                "settings to ensure ligand atoms are included in the trajectory output."
+            )
+            logger.warning(errmsg)
+            return {"structural_analysis_error": errmsg}
```
jthorton, line 1729: Yeah this is great!

src/openfe/protocols/openmm_septop/base_units.py
jthorton: I think the `MultistateAnalysisSettings` might make more sense if we follow the design for the rest of the settings objects, though that might be best done in another PR if we want it to work in HybTop as well.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
...
+                data[f"{label}_COM_drift"].append(lig_drift.results.com_drift)
+
+            if data["time_ps"] is None:
+                data["time_ps"] = (
+                    np.arange(len(universe.trajectory))[::skip] * universe.trajectory.dt
+                )
+
+        return data
+
+    @classmethod
```
IAlibay: As above on classmethod vs staticmethod

IAlibay (changes requested):
Still working on this, couple more things

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
...
+                if simtype == "complex":
+                    data = cls._run_complex_analysis(
+                        ds,
+                        pdb_file,
+                        skip,
+                        ligand_A_indices,
+                        ligand_B_indices,
+                        rdmol_A,
+                        rdmol_B,
+                        protein_selection=protein_selection,
```
IAlibay, line 1752: If possible, I would explicitly define the kwarg arguments for everything.

src/openfe/protocols/openmm_septop/base_units.py
IAlibay: I would agree - definitely prefer this not be an addition to the `run` signature but a setting instead.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1507,323 @@ def _analyze_multistate_energies(
...
         *,
         trajectory: pathlib.Path,
         checkpoint: pathlib.Path,
+        pdb_file: pathlib.Path,
+        ligand_A_indices: list[int],
+        ligand_B_indices: list[int],
+        rdmol_A: Chem.Mol,
+        rdmol_B: Chem.Mol,
+        protein_selection: str = "protein and name CA",
+        skip: int | None = None,
```
IAlibay: Same with this being a setting.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1579,14 +1930,33 @@ def _execute(
         trajectory = simulation.outputs["trajectory"]
         checkpoint = simulation.outputs["checkpoint"]
 
+        alchem_comps = self._inputs["alchemical_components"]
```
IAlibay, line 1904: [nit] It won't matter long term when we serialize things better but for now, rather than adding this as an additional input to the unit, you can also just read it from the inputs of the setup unit.

src/openfe/protocols/openmm_septop/equil_septop_method.py
```diff
@@ -577,6 +577,9 @@ def _create(
 
                 analysis = unit_classes[phase]["analysis"](
                     protocol=self,
+                    stateA=stateA,
+                    stateB=stateB,
```
IAlibay: Are you using these anywhere?

src/openfe/protocols/openmm_septop/base_units.py
jthorton: Yeah that makes sense and the test is great!

src/openfe/tests/protocols/openmm_septop/test_septop_protocol_analysis.py
```diff
@@ -0,0 +1,246 @@
...
+
+import numpy as np
+import pooch
+import pytest
+from rdkit.Chem import SDMolSupplier
+
+from openfe.data._registry import POOCH_CACHE
+from openfe.protocols.openmm_septop.base_units import BaseSepTopAnalysisUnit
+
+pooch_septop_structural = pooch.create(
```
IAlibay: I believe this should be in `_registry.py` under `openfe.data`

src/openfe/tests/protocols/openmm_septop/test_septop_protocol_analysis.py
```diff
@@ -0,0 +1,246 @@
...
+        "nc": base / "solvent.nc",
+        "ligand_A_indices": np.load(base / "ligand_A_indices.npy").tolist(),
+        "ligand_B_indices": np.load(base / "ligand_B_indices.npy").tolist(),
+        "rdmol_A": SDMolSupplier(str(base / "ligand_A.sdf"), removeHs=False)[0],
+        "rdmol_B": SDMolSupplier(str(base / "ligand_B.sdf"), removeHs=False)[0],
+    }
+
+
+@pytest.fixture(scope="session")
+def complex_structural_analysis_result(septop_complex_data, tmp_path_factory):
```
IAlibay: Can you add the word septop in the name here? That way we don't clash with another session scoped fixture in the future.

IAlibay (changes requested):
Sorry it took so long to fully review - here's a final set of things.

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Here I was just following what the hybrid topology protocol does, but I just saw that there's a todo there to change it, so I could also be more specific in the exceptions here
```
# TODO: eventually change this to more specific exception types
        except Exception as e:
            return {"structural_analysis_error": str(e)}
```
I'm not fully sure what types of errors to expect; probably FileNotFoundError if the files somehow were deleted, ValueError, some others?

Regarding whether to put it in the run_X_analysis function, I could also put it there, but I think we may still need it in this function to catch potential erros when opening the netcdf and determining the number of frames?

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Added a comment!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Changed this!

src/openfe/tests/protocols/openmm_septop/test_septop_protocol_analysis.py
hannahbaumann: Changed this!

src/openfe/tests/protocols/openmm_septop/test_septop_protocol_analysis.py
hannahbaumann: Put this into this registry!

src/openfe/protocols/openmm_septop/equil_septop_method.py
hannahbaumann: no, good spot, removed this!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Thanks, changed this!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Moved all imports to the top.

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Move it there!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Moved it to the end!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: actually changing this to `list[np.ndarray] | float | None` and `list[np.ndarray] | np.ndarray | None` still failed with mypy I think because of the append later.
It had actually been cunfusing to me that the per lambda data and the time data are in the same dict but with different shapes, so I decided to separate them instead, I hope that's more clear.

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Actually, I just reverted this since CI was failing. I couldn't figure out how to correctly access the inputs of the other units, setup.inputs[...] was empty.

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: I added a `MultiStateAnalysisSettings` object which is then passed to run.

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Put this into the settings object.

hannahbaumann:
As discussed offline, CI is currently breaking since the prev. openfe 1.11 results files for gather do not have the new `analysis_settings`. Should we turn them on by default to keep backward compatibility or remove the old test files and break backward compatibility? @IAlibay

src/openfe/tests/protocols/openmm_septop/test_septop_protocol_analysis.py
```diff
@@ -0,0 +1,226 @@
...
+        trj_file=d["nc"],
+        output_directory=tmp,
+        dry=False,
+        simtype="solvent",
+        ligand_A_indices=d["ligand_A_indices"],
+        ligand_B_indices=d["ligand_B_indices"],
+        rdmol_A=d["rdmol_A"],
+        rdmol_B=d["rdmol_B"],
+        protein_selection="protein and name CA",
+        skip=None,
```
jthorton, line 68: How about a test which checks the length of the output analysis changes in response to the skip value?

docs/guide/protocols/septop.rst
```diff
@@ -113,6 +113,42 @@ which is different compared to the results in the :class:`.RelativeHybridTopolog
...
+  and ligand B in every lambda state.
+
+* **Symmetry-corrected RMSD.** A symmetry-corrected RMSD is used to account for equivalent
+  atom orderings in symmetric molecules (e.g. a flipping phenyl
+  ring) instead of a standard mass-weighted RMSD.
+
+The following metrics are computed per lambda state for both the complex and solvent legs:
+
+* **Ligand A and B RMSD** — symmetry-corrected RMSD relative to the first production frame.
+* **Ligand A and B COM drift** — centre-of-mass displacement from the initial position
```
jthorton:
```suggestion
* **Ligand A and B COM drift** — centre-of-mass displacement from the initial production simulation position
```
should it be from the first production positions as well?

docs/guide/protocols/septop.rst
```diff
@@ -113,6 +113,42 @@ which is different compared to the results in the :class:`.RelativeHybridTopolog
...
+
+* **Two ligands per simulation.** Since both ligands are present simultaneously throughout
+  the SepTop simulation, structural metrics are computed independently for both ligand A
+  and ligand B in every lambda state.
+
+* **Symmetry-corrected RMSD.** A symmetry-corrected RMSD is used to account for equivalent
+  atom orderings in symmetric molecules (e.g. a flipping phenyl
+  ring) instead of a standard mass-weighted RMSD.
+
+The following metrics are computed per lambda state for both the complex and solvent legs:
```
jthorton:
```suggestion
The following metrics are computed per lambda state for both the complex and solvent leg production simulations only:
```

src/openfe/protocols/openmm_septop/equil_septop_settings.py
```diff
@@ -361,3 +362,7 @@ def must_be_positive(cls, v):
     """
     Settings for the Boresch restraints in the complex
     """
+    analysis_settings: MultiStateAnalysisSettings
```
IAlibay: I believe the fix here would be to have pydantic set a default value.

Something like:

```suggestion
    analysis_settings: MultiStateAnalysisSettings = Field(default_factory=MultiStateAnalysisSettings())
```

I *think* should work - @jthorton might have more experience with this.

jthorton: Yeah I think this should work!

hannahbaumann:
pre-commit.ci autofix

docs/guide/protocols/septop.rst
```diff
@@ -113,6 +113,42 @@ which is different compared to the results in the :class:`.RelativeHybridTopolog
...
+  ring) instead of a standard mass-weighted RMSD.
+
+The following metrics are computed per lambda state for both the complex and solvent leg production simulations only:
+
+* **Ligand A and B RMSD** — symmetry-corrected RMSD relative to the first production frame.
+* **Ligand A and B COM drift** — centre-of-mass displacement from the initial production simulation position
+  (complex leg only).
+* **Protein 2D RMSD** — pairwise RMSD matrix between all analyzed frames (complex leg only).
+
+Results are saved as an NPZ file (``structural_analysis.npz``) and plots are generated
```
IAlibay: [nit] can you link to https://numpy.org/doc/stable/reference/generated/numpy.savez.html in case someone doessn't know what NPZ is.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -57,6 +73,7 @@
     ThermoSettings,
 )
 from openfe.protocols.openmm_md.plain_md_methods import PlainMDSimulationUnit
+from openfe.protocols.openmm_septop.equil_septop_settings import MultiStateAnalysisSettings
```
IAlibay: Can you import directly from omm_settings or does it cause a circular import?

src/openfe/tests/protocols/openmm_septop/test_septop_protocol_analysis.py
hannahbaumann: Added a test here: https://github.com/OpenFreeEnergy/openfe/pull/1982/commits/396b35312adfbd1183c845fe2a27a64f02e82503

src/openfe/protocols/openmm_septop/base_units.py
IAlibay: That's fair - I'd open an issue and put a todo here for now then.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1503,11 +1520,290 @@ def _analyze_multistate_energies(
...
         self,
         *,
         trajectory: pathlib.Path,
         checkpoint: pathlib.Path,
+        pdb_file: pathlib.Path,
+        ligand_A_indices: list[int],
+        ligand_B_indices: list[int],
+        smc_A: SmallMoleculeComponent,
+        smc_B: SmallMoleculeComponent,
+        analysis_settings: MultiStateAnalysisSettings,
```
IAlibay: Rather than passing it directly, can you just access it via `_get_settings`? That way you won't have two sets of settings around.

src/openfe/protocols/openmm_septop/base_units.py
IAlibay: That's odd - something we can work out in a separate issue / PR - could you raise an issue about setup.inputs being empty? That seems unintuitive.

src/openfe/protocols/openmm_septop/base_units.py
```diff
@@ -1578,10 +1903,34 @@ def _execute(
         # Get the relevant inputs for running the unit
         trajectory = simulation.outputs["trajectory"]
         checkpoint = simulation.outputs["checkpoint"]
+        alchem_comps = self._inputs["alchemical_components"]
+        smc_A = alchem_comps["stateA"][0]
+        smc_B = alchem_comps["stateB"][0]
+
+        # Get the analysis settings
+        settings = self._get_settings()
+        analysis_settings = settings["analysis_settings"]
```
IAlibay: Can we do this in `run` to match the other units / general pattern across the protocols?

src/openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -741,3 +741,30 @@ class MDOutputSettings(OutputSettings):
...
+    protein_selection: str = "protein and name CA"
+    """
+    MDAnalysis selection string for the guest atoms used for
+    alignment and RMSD calculations in the complex phase.
+    Ignored for the solvent phase. Default "protein and name CA".
+    """
+    skip: int | None = None
+    """
+    Frame stride for structural analysis. If ``None``, a stride is
+    chosen such that approximately (max.) 500 frames are analyzed per state.
```
IAlibay, line 768: Might be good to open an issue to work out if this is something we want to keep around in the future - at least it seems rather unintuitive to not just use all the data we have on hand.

IAlibay (changes requested):
overall looks good - just the one blocker for me re: settings

src/openfe/protocols/openmm_utils/omm_settings.py
hannahbaumann: Raised an issue here: https://github.com/OpenFreeEnergy/openfe/issues/1993

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Moved this into run!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Yes, changed this!

docs/guide/protocols/septop.rst
hannahbaumann: Added this!

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Raised an issue here: https://github.com/OpenFreeEnergy/openfe/issues/1994

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Changed this!

src/openfe/protocols/openmm_septop/base_units.py
hannahbaumann: Raised an issue here: https://github.com/OpenFreeEnergy/openfe/issues/1999

src/openfe/tests/protocols/openmm_septop/test_septop_protocol_analysis.py
jthorton: Fantastic!

IAlibay (approved):
lgtm, thanks!

PR #1984, opened 2026-06-04
return partial convergence analysis if mbar fails for individual fraction

aqemia-benedict-tan:
When running ABFE calculations in OpenFE, MBAR sometimes fails to produce an estimate at low fractions of uncorrelated samples (~0.1-0.4). It would be helpful to still return the convergence trace where MBAR can generate a meaningful result (e.g. >0.7) This PR modifies `get_forward_and_reverse_analysis` under `openmm_utils` so that `NaN` values are appended to the forwards/reverse estimate lists where MBAR fails and the successful estimates are still returned.

Checklist

- [x]  All new code is appropriately documented (user-facing code *must* have complete docstrings).
- [x]  Ran [pre-commit](https://pre-commit.com/): you can run pre-commit locally or comment on this PR with `pre-commit.ci autofix`.

- [x]  I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at https://developercertificate.org/.


src/openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -339,30 +344,52 @@ def get_forward_and_reverse_analysis(
...
-                    self.analyzer,
-                    u_ln[:, :samples],
-                    new_N_l,
-                    0,
-                    self.units,
-                )
+                # MBAR can fail to converge at low fractions of uncorrelated
+                # samples. If this happens, append NaN and carry on so that
+                # estimates at higher fractions are still reported.
+                try:
```
IAlibay: [nit] Given the similar-ish duplicated code, it might be good to just move this to a helper method that takes in the sub-sampled `u_ln`, `fraction`, and either "reverse" or "forward" for the warning string.

IAlibay (changes requested):
Thanks @aqemia-benedict-tan, this looks great - just the one comment on the code as-is.

The main thing we would need (and why I have to request changes) is that we need to add tests to cover the new behaviour

Ideally we would want to:

1. Test that the check the array of forward and backwards energies returned make sense (similar to https://github.com/OpenFreeEnergy/openfe/blob/53470827a4489bc4f69a72d37cdcb2089b409c42/src/openfe/tests/protocols/test_openmmutils.py#L360)
2. Something in `tests/analysis/test_plotting.py` that checks that passing in an array with NaN's works fine and probably a manual check that the plot looks ok.

aqemia-benedict-tan:
Thanks for the feedback @IAlibay, I have refactored to include the helper function `_get_fraction_free_energy` for the purpose of appending `NaN` in the event of low fraction failure. I have also added unit tests to verify that clean arrays are returned in the event of the lowest fraction in the forwards estimate failing and to assert that the convergence plot is returned successfully if an array containing `NaN` estimates is passed to it. The resulting plot looks like this: 

image

IAlibay:
Thanks @aqemia-benedict-tan ! I'lll have another look at this early next week.

src/openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -339,30 +406,23 @@ def get_forward_and_reverse_analysis(
             for chunk in chunks:
                 new_N_l = np.array([chunk for _ in range(n_states)])
                 samples = chunk * n_states
+                fraction = chunk / N_l[0]
 
                 # Forward
-                DG, dDG = self._get_free_energy(
```
IAlibay, line 344: I've been thinking about this a little bit more and I've come to the realization that we probably should NaN both the forward and the reverse chunk if either fails.

My reasoning is that whilst a failure is directly caused by a lack of convergence in the estimator, it's more broadly caused by too few data points (generally too high a decorrelation time). So whilst one direction could pass (e.g. the reduced potentials happen to be easier to deal with), it's probably not reasonable to include either.

src/openfe/tests/analysis/test_plotting.py
```diff
@@ -167,3 +169,32 @@ def test_plot_2D_rmsd(num):
...
+    }
+
+
+def test_plot_convergence_with_nan():
+    """
+    NaN entries (e.g. from MBAR failing to converge at low fractions of
+    uncorrelated samples) should be rendered as gaps without raising.
+    """
+    forward_reverse = _make_forward_reverse(
+        forward=[np.nan, np.nan, -9.2, -9.4, -9.5, -9.6, -9.65, -9.7, -9.72, -9.75],
```
IAlibay, line 194: What if the nan happens towards the end? It's unlikely but could happen.

My understanding is that the `fill_between` call would just do nothing. In that case - it might be better to just return None for the whole thing.

src/openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -311,6 +373,11 @@ def get_forward_and_reverse_analysis(
         * This method does not currently use bootstrap uncertainties due to
           issues with the solver when using low amounts of data points. All
           uncertainties are MBAR analytical errors.
+        * If MBAR fails to obtain an estimate for a given sample fraction (a
```
IAlibay, line 387: [ok for separate issue / PR] This new behaviour should be reflected in the docstring for the relevant results classes for the ABFE, SepTop & HybridTop protocols.

For example here: https://github.com/OpenFreeEnergy/openfe/blob/53470827a4489bc4f69a72d37cdcb2089b409c42/src/openfe/protocols/openmm_afe/afe_protocol_results.py#L72

src/openfe/tests/protocols/test_openmmutils.py
```diff
@@ -408,6 +408,50 @@ def test_free_energies(self, analyzer):
             rtol=5e-01,
         )  # fmt: skip
 
+    def test_forward_and_reverse_nan_on_mbar_failure(self, analyzer):
```
IAlibay: Thanks for doing this - this test looks great!

IAlibay (changes requested):
Sorry - I somewhat changed my mind on what the behaviour should look like on failure, that will require some code changes.

Could you also:
- Add a news entry under `openfe/news`.
- Add yourself to the CITATION.cff author list?

Thanks!

src/openfe/protocols/openmm_utils/multistate_analysis.py
aqemia-benedict-tan: Yes this makes sense - commits 7b7dba3248756018926286febad977f724c27aa6 and dbf9b71d39ad2acd37b08c94bef7f1edc643dbd3 refactor the `_get_fraction_free_energy` helper to return `NaN` for both directions, if either the forwards or reverse estimate fails. 
Commit 42ee8389103700467eb7df72a8dca2fb9cbcd1b2 updates the tests, the test at `src/openfe/tests/protocols/test_openmmutils.py` is adapted to ensure `NaN` is returned for both directions in the case that the lowest fraction of samples failes in either the forwards or reverse direction.
The new graph from the plotting test looks like this: 

image

src/openfe/tests/analysis/test_plotting.py
aqemia-benedict-tan: Good point! Commit ce66327c85556e58452ab185011cd1666527515e fixes this by returning`None` if the final element of the reverse and forwards estimate (i.e. fraction 1.0) is`NaN`. A unit test is also added for this scenario.

aqemia-benedict-tan:
> Sorry - I somewhat changed my mind on what the behaviour should look like on failure, that will require some code changes.
> 
> Could you also:
> 
> * Add a news entry under `openfe/news`.
> * Add yourself to the CITATION.cff author list?
> 
> Thanks!

Added in commits 4ce6e8ac7cab01544211fb966506808634c08f38 and e5d52daedee0bf653dba7739a18ffe9894a65230

src/openfe/protocols/openmm_utils/multistate_analysis.py
aqemia-benedict-tan: Sounds good - easiest to open a new PR once the changes have been merged?

src/openfe/protocols/openmm_utils/multistate_analysis.py
IAlibay: Amazing, thanks!

src/openfe/protocols/openmm_utils/multistate_analysis.py
IAlibay: Sounds good, I've opened https://github.com/OpenFreeEnergy/openfe/issues/1998 so we don't lose track of it.

src/openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -339,30 +421,20 @@ def get_forward_and_reverse_analysis(
...
-                reverse_DGs.append(DG)
-                reverse_dDGs.append(dDG)
+                forward_DGs.append(forward_DG)
+                forward_dDGs.append(forward_dDG)
+                reverse_DGs.append(reverse_DG)
+                reverse_dDGs.append(reverse_dDG)
 
-                fractions.append(chunk / N_l[0])
+                fractions.append(fraction)
         except ParameterError:
```
IAlibay: I think this outer try/except is now dead. Can you remove it (alongisde the associated ParameterError import)?

src/openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -373,6 +445,22 @@ def get_forward_and_reverse_analysis(
...
+        # analysis is not meaningful and we discard it by returning None.
+        final_forward = forward_reverse["forward_DGs"][-1].m
+        final_reverse = forward_reverse["reverse_DGs"][-1].m
+        if np.isnan(final_forward) or np.isnan(final_reverse):
+            wmsg = (
+                "MBAR could not obtain a free energy estimate using the full set "
+                "of uncorrelated samples (fraction 1.0); discarding the forward "
+                "and reverse convergence analysis."
+            )
+            warnings.warn(wmsg)
```
IAlibay: Can you set the stacklevel parameter to 2 here? We're not great at being consistent about it elsewhere, but it's a small QoL thing that will be useful for users.

IAlibay (changes requested):
A few more things and then I think we're there!

src/openfe/protocols/openmm_utils/multistate_analysis.py
aqemia-benedict-tan: Fixed in commit 8cd3275d312e5abce352e30d415f0b0417e3b709

src/openfe/protocols/openmm_utils/multistate_analysis.py
aqemia-benedict-tan: Fixed in commit 92324532e5a7e6b46aa99de0a42418b3c6427f3d

IAlibay:
Sorry this slipped off my todo list - I'll prioritise this tomorrow.

src/openfe/protocols/openmm_utils/multistate_analysis.py
```diff
@@ -282,6 +282,79 @@ def get_equil_free_energy(self) -> tuple[Quantity, Quantity]:
...
+                0,
+                self.units,
+            )
+        except ParameterError:
+            wmsg = (
+                f"Could not obtain a free energy estimate at fraction "
+                f"{fraction:.2f} of the uncorrelated samples; recording NaN "
+                "for both the forward and reverse estimates."
+            )
+            warnings.warn(wmsg)
```
IAlibay:
```suggestion
            warnings.warn(wmsg, stacklevel=2)
```

IAlibay (changes requested):
Just the one thing (will just directly merge the suggestion), otherwise it looks great to me!

Because we also need to fix the ProtocolResult docstrings etc.. we can't include it in the release coming out (hopefully) today, I'm going to put a "request changes" block to avoid us merging in too early. We can merge once the release goes out.

IAlibay (approved):
Sorry about the delay, we're ready to merge this now, thanks so much @aqemia-benedict-tan !

PR #2000, opened 2026-06-12
Use unbiased std for uncertainty estimates

jthorton:
Updates the uncertainty functions on protocol results to return the unbiased estimate of the standard deviation from the sample population.

# Questions
- Would it be better to return the `SE = std / sqrt(n)` which is the uncertainty in the estimate of the average rather than the spread of the data? Where std is the unbiased estimate? 
- What do we do when `n=1` the uncertainty will now return Nan? 

Checklist


IAlibay (approved):
Just needs a news item, otherwise looks good.

hannahbaumann:
> Would it be better to return the SE = std / sqrt(n) which is the uncertainty in the estimate of the average rather than the spread of the data? Where std is the unbiased estimate?

I'm also not so sure, I think both options make sense, but of course answer a different question. Is there something in the best practices paper about this?

IAlibay:
I would prefer the standard deviation rather than the standard error as the main measure of uncertainty. 

We can discuss this more offline.

IAlibay (approved):

PR #2018, opened 2026-06-16
Updated the CLI to use unbiased std for errors 

jthorton:
Follow on from #2000 updates the CLI error functions to report the unbiased std as the error.

# Questions
- [ ] Do we want to update the gather CLIs to use the `get_estimate` and `get_uncertainty` protocol result functions to keep them in sync now or is this a 2.0 issue?
- [ ] If not can we have a single util function which calculates the error and is used in the protocols and the CLI, isolating the behaviour to a single place? 

Checklist

### DIFF
--- a/src/openfecli/commands/gather_abfe.py
+++ b/src/openfecli/commands/gather_abfe.py
@@ -150,9 +150,12 @@ def _get_legs_from_result_jsons(
 
 def _error_std(r):
     """
-    Calculate the error of the estimate as the std of the repeats
+    Calculate the error of the estimate as the unbiased std of the repeats
     """
-    return np.std([v[0].m for v in r["overall"]])
+    std = np.std([v[0].m for v in r["overall"]], ddof=1)
+    if np.isnan(std):
+        std = 0.0
+    return std
 
 
 def _error_mbar(r):
--- a/src/openfecli/commands/gather_septop.py
+++ b/src/openfecli/commands/gather_septop.py
@@ -163,9 +163,12 @@ def _get_names(result: dict) -> tuple[str, str]:
 
 def _error_std(r):
     """
-    Calculate the error of the estimate as the std of the repeats
+    Calculate the error of the estimate as the unbiased std of the repeats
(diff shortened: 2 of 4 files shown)


IAlibay (approved):
lgtm, thanks!

IAlibay:
> Do we want to update the gather CLIs to use the get_estimate and get_uncertainty protocol result functions to keep them in sync now or is this a 2.0 issue?

I think to do this you will need to load in the full Results object, my understanding is that we aren't doing this right now because there's some performance issues with it based on how it's currently stored in the quickrun outputs.

This is to say, I think it's a "new CLI issue", rather than something that can be dealt with today, but not necessary an 2.0 issue.

IAlibay:
>  If not can we have a single util function which calculates the error and is used in the protocols and the CLI, isolating the behaviour to a single place?

This one I'm not sure I fully understand. Do you mean something like `_get_stdev` that does something like:

```python
   std = np.std([v[0].m for v in r["overall"]], ddof=1)
    if np.isnan(std):
        std = 0.0
    return std
```

?

If so, I wouldn't be against having that as a small reusable method, as long as we're ok with it maybe being refactored in the near future (as we move towards squashing the protocols together).

jthorton:
> Do you mean something like _get_stdev that does something like:

Yesa single function which is called by the CLI and the protocols, I'll make an issue for that.

atravitz:
> I think to do this you will need to load in the full Results object, my understanding is that we aren't doing this right now because there's some performance issues with it based on how it's currently stored in the quickrun outputs.
> 
> This is to say, I think it's a "new CLI issue", rather than something that can be dealt with today, but not necessary an 2.0 issue.

yes, this will require light breaking changes to quickrun - to be discussed in ResultsNetwork conversation tomorrow

atravitz:
Splitting off into a separate function is nice but not necessary - there's so much duplicated code I'm going to consolidate when merging everything into a single gather that this is the least of my worries.

PR #2027, opened 2026-06-18
testing: catch konnektor warnings 

atravitz:
before we bump to the latest version of konnektor, I'd like to make sure we're catching warnings.

Checklist

### DIFF
--- a/src/openfe/tests/setup/test_network_planning.py
+++ b/src/openfe/tests/setup/test_network_planning.py
@@ -280,11 +280,15 @@ class TestRadialNetworkGenerator:
         toluene, others = toluene_vs_others
         mappers = [BadMapper(), lomap_old_mapper]
 
-        network = openfe.setup.ligand_network_planning.generate_radial_network(
-            ligands=others,
-            central_ligand=toluene,
-            mappers=mappers,
-        )
+        with pytest.warns(
+            UserWarning,
+            match="Multiple mappers were provided, but no scorer. Only the first valid mapper will be used",
+        ):
+            network = openfe.setup.ligand_network_planning.generate_radial_network(
+                ligands=others,
+                central_ligand=toluene,
+                mappers=mappers,
+            )
 
         expected_names = set([c.name for c in others] + ["toluene"])
 
@@ -327,6 +331,7 @@ def test_generate_maximal_network(
     extra_mapper,
     lomap_old_mapper,
     simple_scorer,
+    recwarn,
 ):
     toluene, others = toluene_vs_others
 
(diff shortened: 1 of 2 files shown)


mikemhenry:
do we expect the tests to take ~42 min?

src/openfe/tests/setup/test_network_planning.py
```diff
@@ -344,6 +349,14 @@ def test_generate_maximal_network(
...
     )
 
+    if extra_mapper and scorer is None:
+        assert len(recwarn) == 1
+        w = recwarn.pop(UserWarning)
+        assert issubclass(w.category, UserWarning)
+        assert "Only the first valid mapper will be used: <LomapAtomMapper" in str(w.message)
+    else:
+        assert len(recwarn) == 0
+
```
mikemhenry, line 359:
```suggestion
    if extra_mapper and scorer is None:
        with pytest.warns(
            UserWarning,
            match="..",
        ):
            network = openfe.setup.ligand_network_planning.generate_maximal_network(
        ligands=others + [toluene],
        mappers=mappers,
        scorer=scorer,
        progress=with_progress,
        )
    else:
        network = openfe.setup.ligand_network_planning.generate_maximal_network(
        ligands=others + [toluene],
        mappers=mappers,
        scorer=scorer,
        progress=with_progress,
    )

```

src/openfe/tests/setup/test_network_planning.py
```diff
@@ -344,6 +349,14 @@ def test_generate_maximal_network(
         progress=with_progress,
     )
```
mikemhenry, line 350:
```suggestion
```

mikemhenry (changes requested):
I think I prefer using `with pytest.warns` we don't use `recwarn` in other tests (I had to look up that fixture (we do use it actually once in a gufe test but I have the same comment for that one now)). 

Do you have some reason you prefer using `rcwarn`? I could be convinced but I like that you don't need to do extra bookkeeping to check the warning.

src/openfe/tests/setup/test_network_planning.py
mikemhenry: idk what happened, the preview was correct but I was trying to say "delete this block" 

```
    network = openfe.setup.ligand_network_planning.generate_maximal_network(
        ligands=others + [toluene],
        mappers=mappers,
        scorer=scorer,
        progress=with_progress,
    )
```

src/openfe/tests/setup/test_network_planning.py
atravitz: my motivation for using recwarn here is that I didn't want to have the call duplicated in the `if/else`.

 The way I have it written, I think it's clearer that the call is identical - the duplication is confusing IMO.

mikemhenry: Sounds good!

mikemhenry (approved):

src/openfe/tests/setup/test_network_planning.py
mikemhenry: I think since this is so we can get stuff tested it is good, but I think this speaks to an issue with the testing, we conditionally expect a warning so we should really be using that. I was thinking that this sounded like using a context manager, but one thing that is different with your version is that you are asserting zero warnings are rasied, not just the one from lomap we care about (or in the case of us expecting a warning, that we only get one warning and no other warning). 

If that level of strictness is fine/wanted, then the `recwarn` is the correct approach.

atravitz: Ah that's a good point about other warnings being caught by this as well. I'll merge this as-is but I'm willing to revisit it.

PR #2031, opened 2026-06-22
Test kartograf 2.0

mikemhenry:
Checklist


mikemhenry:
So python 3.11 and 3.12 had issues solving the env, but it looks like if we drop the `numpy ~=2.0` from the kartograf feedstock, then everything will work. See https://github.com/OpenFreeEnergy/kartograf/pull/158

mikemhenry:
We can't install kartograf from conda-forge until we merge in https://github.com/conda-forge/kartograf-feedstock/pull/25 BUT I believe this is a good test that things are working in openfe with kartograf 2.0.

mikemhenry:
These are the two spots we actually call kartograf:
(this is technically a test)
https://github.com/OpenFreeEnergy/openfe/blob/a5d671404e229eba9c1807e898522758f4c8cfc4/src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py#L772
and here:
https://github.com/OpenFreeEnergy/openfe/blob/a5d671404e229eba9c1807e898522758f4c8cfc4/src/openfecli/parameters/plan_network_options.py#L149-L158


The main breaking change in kartograf is `map_hydrogens_on_hydrogens_only` option is now True by default. Previously it was False.

mikemhenry:
So since we specify that option (see above) we should be fine in theory to use kartograf 2.0.

atravitz:
example notebooks failing with kartograf-related things here: https://github.com/OpenFreeEnergy/openfe/actions/runs/27986771574/job/82830005369

IAlibay (approved):

environment.yml
```diff
@@ -54,6 +54,7 @@ dependencies:
   - threadpoolctl
   - pip:
     - git+https://github.com/OpenFreeEnergy/gufe@main
+    - git+https://github.com/OpenFreeEnergy/kartograf@main
```
mikemhenry:
```suggestion
    - git+https://github.com/OpenFreeEnergy/kartograf@v2.0.0
```

environment.yml
```diff
@@ -54,6 +54,7 @@ dependencies:
   - threadpoolctl
   - pip:
     - git+https://github.com/OpenFreeEnergy/gufe@main
+    - git+https://github.com/OpenFreeEnergy/kartograf@v2.0.0
```
atravitz: why not install from conda-forge?

atravitz (changes requested):

environment.yml
mikemhenry: Good point! I tried to back in the [start of the PR](https://github.com/OpenFreeEnergy/openfe/pull/2031/commits/cf28293ec2f3f95c775654803af5021ddd7a6f46), but ran into env issues (I think it was related to the numpy pin) but now that we fixed that, we should be able to install from conda-forge.

atravitz (approved):

PR #2042, opened 2026-06-25
Fix residue naming

### LINKED ISSUE #2050: Change ligand and cofactor residue names in the topology

hannahbaumann:
Checklist


hannahbaumann:
pre-commit.ci autofix

hannahbaumann:
pre-commit.ci autofix

news/fix_unk_resnames.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Small molecules in RelativeHybridTopology topologies (including the output PDB) are now named LIG (alchemical ligand) and CF1, CF2… (cofactors) instead of UNK.
```
IAlibay: What about if a residue name was already assigned?

src/openfe/protocols/openmm_utils/offmolecule_utils.py
IAlibay, line 1: I would call this "offmolecule_utils" so that we know it's utilities for openff molecules and not rdkit molecules

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
```diff
@@ -2401,3 +2408,74 @@ def test_dry_run_vacuum_write_frequency(
...
+
+    u = mda.Universe(out["pdb_structure"])
+    lig = u.select_atoms("resname LIG")
+    cof = u.select_atoms("resname CF1")
+    assert lig.n_atoms > 0
+    assert cof.n_atoms > 0
+    assert set(lig.indices).isdisjoint(cof.indices)
+
+
+def test_get_metadata_inconsistent_warns(caplog):
```
IAlibay: Can you move this and the next one to test_openmmutils instead?

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
```diff
@@ -2401,3 +2408,74 @@ def test_dry_run_vacuum_write_frequency(
...
+    stateA = openfe.ChemicalSystem({"ligand": ligA, "cofactor": cof})
+    stateB = openfe.ChemicalSystem({"ligand": ligB, "cofactor": cof})
+
+    out = _run_setup_dry(stateA, stateB, mapping, vac_settings, tmp_path)
+
+    u = mda.Universe(out["pdb_structure"])
+    resnames = {r.resname for r in u.residues}
+    assert "MYC" in resnames
+    assert "CF1" not in resnames
+    assert "LIG" in resnames
```
IAlibay: Resnames is mostly cached at the Topology level in the MDAnalysis Universe, so you don't need to create the set.

```suggestion
    assert "MYC" in u.residues.resnames
    assert "CF1" not in u.residues.resnames
    assert "LIG" in u.residues.resnames
```

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -1535,7 +1547,7 @@ def _structural_analysis(
         from openfe_analysis import rmsd
 
         try:
-            data = rmsd.gather_rms_data(pdb_file, trj_file)
+            data = rmsd.gather_rms_data(pdb_file, trj_file, ligand_selection="resname LIG")
```
IAlibay: This will fail is someone provides their own residue names ahead of time (which is something we should encourage - i.e. power users should be rewarded not punished).

There are two potential solutions here:

1. (lowest cost / temporary fix) We just store the alchemical residue names in the output of the setup unit and then fetch them in the analysis unit.
2. (higher cost / more complicated) We pass in the alchemical components & the comp resids through all the way to the analysis unit and then extract the relevant residue names from that.


For now, I would suggest we just pass the residue names and have the residue name kwarg be optional and/or have a default of LIG, that way we don't break backwards compatibility.

We can clean up after the fact.

IAlibay: Note that for hybrid toplogy, it's technically possible for both endstate molecules to have different names, so we should check that it works properly.

IAlibay (changes requested):
This looks good thanks! - couple of things to fix but otherwise should be fine.

After this PR, will you be able to do the same for PlainMD, AFE, and SepTop?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
jthorton: I think a 3rd option might be best, where we do away with tracking the residue names and just keep the alchemical atom indices around. This would mean moving away from the `gather_rms_data` function and using the new openfe-analaysis API but we plan on doing that anyway, so this could be a good time to do it? 

This would fix the case pointed out above and the fact that if a user sets the alchemical ligands and the cofactors to have the same residue name (LIG), we will still have the same issue in the analysis unit.

jthorton:
Not blocking: Does a user have a way of setting the residue_name via a CLI input? It seems like the atom metadata fields are not preserved on writing an openff molecule to file?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
jthorton: I wouldn't consider the above blocking though we could have this as a temp fix and move to indices when we change the analysis for this protocol?

src/openfe/protocols/openmm_utils/offmolecule_utils.py
hannahbaumann: Done!

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: Yes, I like the 3rd option as the long term fix, for now I went ahead with option 1, but happy to change it to a longer term fix.

Re two ligands different residue names: I tested this out and it assigns the residue name from the stateA ligand to the hybrid topology (in the HTF). We could still check for both residue names since it wouldn't hurt, but may also confuse the user?

IAlibay: > I think a 3rd option might be best, where we do away with tracking the residue names and just keep the alchemical atom indices around. 

This somewhat fits within the longer term idea of switching comp_resids to just comp_indices (where you track the indices of each component in the system). However, that's a lot bigger lift.

I don't mind just tracking the alchemical indices in the setup results - that's something we nearly already do in the AFE Protocol. But yeah, for the sake of incremental PRs = faster velocity, I would vote for the quick fix now, then the medium/long term fix after.

IAlibay:
> Not blocking: Does a user have a way of setting the residue_name via a CLI input? It seems like the atom metadata fields are not preserved on writing an openff molecule to file?

I had forgotten that this was still an issue...

The problem is that residue name is a per atom PDBResidueInfo property in rdkit molecules, and a) gufe SMCs don't serialize that, and b) it doesn't get written out when you write out an SDF file. See: https://github.com/OpenFreeEnergy/gufe/issues/327

news/fix_unk_resnames.rst
hannahbaumann: Added something!

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
hannahbaumann: Done!

hannahbaumann:
> After this PR, will you be able to do the same for PlainMD, AFE, and SepTop?
Yes, after having decided on the way how to address this in this PR, I can do the same for the other protocols in a follow-up PR!

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
```diff
@@ -50,6 +51,12 @@
     HAS_NAGL,
     HAS_OPENEYE,
 )
+from openfe.protocols.openmm_utils.offmolecule_utils import (
+    _get_offmol_metadata,
+    _get_offmol_resname,
+    _set_offmol_metadata,
+    _set_offmol_resname,
+)
```
IAlibay, line 57:
```suggestion
from openfe.protocols.openmm_utils.offmolecule_utils import (
    _get_offmol_resname,
    _set_offmol_resname,
)
```
I think these two imports weren't getting used?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -753,6 +754,20 @@ def run(
...
+        for smc, offmol in small_mols.items():
+            if _get_offmol_resname(offmol) is not None:
+                continue
+            if smc in alchemical:
+                _set_offmol_resname(offmol, "LIG")
+            else:
+                _set_offmol_resname(offmol, f"CF{cofactors.index(smc) + 1}")
+
+        # Record the resname the alchemical ligand actually ended up with
+        alchem_resname = _get_offmol_resname(small_mols[mapping.componentA])
```
IAlibay: I think you'll need to make this a list and include componentB.

IAlibay: Also, we're not accounting for the edge case where this clashes with any other residue names... It's rare but we probably should defend against it.

IAlibay (changes requested):
I think this needs to be a list of residue names so that you can account for the case where you have a different residue name between component A and component B.

Also, we'll probably want to protect against clashes in cofactor names with the ligand names.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: I changed this to be a list instead. 
I also added a check for existing residue names to make sure that those can never clash. (both for e.g. cofactors being named LIG or ligands names CF1).

hannahbaumann:
pre-commit.ci autofix

hannahbaumann:
@IAlibay I changed this now such that the cofactors would always get the same COF residue name, but different residue numbers. The two alchemical ligands would be LG1 and LG2 by default, would that make sense?

hannahbaumann:
pre-commit.ci autofix

news/fix_unk_resnames.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Small molecules in RelativeHybridTopology topologies (including the output PDB) are now named LIG (alchemical ligand) and CF1, CF2… (cofactors) instead of UNK. If a residue name was already assigned, the assigned one is kept.
```
hannahbaumann:
```suggestion
* Small molecules in RelativeHybridTopology topologies (including the output PDB) are now named LG1 amd LG2 (alchemical ligand) and COF (cofactors) instead of UNK. If a residue name was already assigned, the assigned one is kept.
```

hannahbaumann:
pre-commit.ci autofix

hannahbaumann:
The failing test is only failing because of hashing/caching with the smc, the previous test sets the resname of the cofactor to LG1, the next test assigns a new resname (MYC) to the offmol, but after `openfe.SmallMoleculeComponent.from_openff(off)` that resname is again LG1.

hannahbaumann:
I just tested this against gufe@store_residue_info (PR https://github.com/OpenFreeEnergy/gufe/pull/811) and the failing test passes with that.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -753,6 +759,47 @@ def run(
...
+        for smc, offmol in small_mols.items():
+            if _get_offmol_resname(offmol) is not None:
+                continue
+            if smc in alchemical:
+                base = "LG1" if smc in stateA_comps else "LG2"
+                name = base if base not in used else _unique(used)
+                used.add(name)
+            else:
+                name = "COF"
+            _set_offmol_resname(offmol, name)
```
jthorton, line 790:
```suggestion
        used: set[str] = set()

        for smc, offmol in small_mols.items():
            existing = _get_offmol_resname(offmol)
            if existing is not None:
                used.add(existing)
                continue

            if smc in alchemical:
                preferred = "LG1" if smc in stateA_comps else "LG2"
                name = preferred if preferred not in used else _unique(used)
            else:
                name = "COF"

            _set_offmol_resname(offmol, name)
            used.add(name)
```

jthorton: I think we can get this down to one loop and I think it makes it easier to read as well

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -753,6 +759,47 @@ def run(
...
+
+        names: set[str] = set()
+        for comp in (mapping.componentA, mapping.componentB):
```
jthorton: Why do we use the mapping and not the alchemical variable from above?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -1641,6 +1697,7 @@ def run(
             trj_file=trajectory,
             output_directory=self.shared_basepath,
             dry=dry,
+            ligand_resnames=ligand_resnames,
```
jthorton, line 1698: Its not clear to me what name is actually stored in the output pdb file for the hybridtopology ligand would we expect just the stateA name LG1 if so what is LG2 used for?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -1573,6 +1626,7 @@ def run(
         pdb_file: pathlib.Path,
         trajectory: pathlib.Path,
         checkpoint: pathlib.Path,
+        ligand_resnames: list[str] = ["LG1", "LG2"],
```
jthorton: It might be better to have the defaults declared in a single place as we currently have it here and in the execute function

jthorton (commented):
This looks great @hannahbaumann though the role of LG2 is not clear currently as the output seems to use LG1 as the name for the hybrid ligand in the output topology. Having the other name around if we write out seperate PDBs for the end states would make sense but maybe its confusing until then?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: i had separated these for the (probably unlikely case) that e.g. a cofactor was named LG1, if it came after the ligand, the LG1 would not be in used yet and both would carry the same residue name. Or is there a different way around that?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: I put the default value in the top of the file and am now reusing it for both.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: Ah, this was left over from before, change this!

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: Yes, the output PDB would only have the residue name for the stateA ligand, as that's applied o the hybrid topology ligand. I had changed this to the list of both ligand names based on a comment from Irfan above, I think the residue name for ligand B would be needed if e.g. we decided to equilibrate both end states separately first. Maybe @IAlibay can chime in on additional reasons why we'd want to store both residue names?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -80,6 +81,8 @@
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_LIGAND_RESNAMES = ("LG1", "LG2")
```
IAlibay: Sorry I'm going to go counter to what Josh said - I'm really not a big fan of this in such a big file. I'd prefer it being redefined than to have a hanging global variable that gets referenced effectively one time.

At worst let's just remove the default from `run`.

jthorton: Sorry @hannahbaumann I should have been more specific I did mean just declare in execute and remove the default from run if you agree to only have it defined in one place.

hannahbaumann: Ok, removed the default from run now!

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: Ah, but this now results in an API break...

news/fix_unk_resnames.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Small molecules in RelativeHybridTopology topologies (including the output PDB) are now named LG1 amd LG2 (alchemical ligand) and COF (cofactors) instead of UNK. If a residue name was already assigned, the assigned one is kept.
```
IAlibay:
```suggestion
* Small molecules in RelativeHybridTopologyProtocol topologies (including the output PDB) are now named LG1 amd LG2 (alchemical ligand) and COF (cofactors) instead of UNK. If a residue name was already assigned, the assigned one is kept.
```
[nit] just was a bit confusing to have topology twice in a row.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -692,9 +693,14 @@ def _subsample_topology(
         bfactors[np.isin(selection_indices, list(atom_classes["unique_new_atoms"]))] = 0.75
 
         if len(selection_indices) > 0:
+            sub_top = hybrid_topology.subset(selection_indices)
+            # Renumber sequentially so same-named residues (e.g. two COF cofactors,
```
IAlibay: Do we know if resids from the original PDB make it through all the way here?

I.e. if a user has a PDB file that starts at 4 instead of 1, until you do this renumbering - would the first residue also start at 4?

The main reason I ask this is that renumbering essentially loses any input information, so in some ways it makes the user experience worse because they can't then map things back to the input.

IAlibay: Do the small molecules have a different chain ID from the protein residues? If so, would assigning the residue number at the same time we assign residue names to the offmol be suitable?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -753,6 +759,47 @@ def run(
...
+        # All cofactors are assigned "COF" and differ by residue index.
+        for smc, offmol in small_mols.items():
+            if _get_offmol_resname(offmol) is not None:
+                continue
+            if smc in alchemical:
+                base = "LG1" if smc in stateA_comps else "LG2"
+                name = base if base not in used else _unique(used)
+                used.add(name)
+            else:
+                name = "COF"
```
IAlibay: [nit] this technically allows a case where the ligand(s) are named COF and the cofactors also get called COF.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
IAlibay: Ah yeah, I might have focused too much on the idea that the ligands could have different residue names.

Here's where I'm roughly at (sorry it's messy thoughts, maybe we should discuss in a call!):

- I think it's important that we track & pass through a list of ligand residues (as @hannahbaumann mentions, it may be useful longer term if we analyze different end states etc...).
- At the risk of being annoying (sorry!), it might be helpful to just name all the ligand residues the same thing by default (e.g. going back to LIG) -- sorry that's my bad, I had forgotten that only one end state gets propagated through to the PDB.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -1669,6 +1726,7 @@ def _execute(
             pdb_file=pdb_file,
             trajectory=trajectory,
             checkpoint=checkpoint,
+            ligand_resnames=setup_results.outputs.get("alchemical_resnames", ["LG1", "LG2"]),
```
IAlibay: Thinking this through a little bit more, is this necessary? I.e. yes it's an API break, but (maybe I'm forgetting something!) I _think_ that would only be an issue if you tried to resume an execution between versions of `openfe`, which would be disallowed by the resume checks.

IAlibay (changes requested):
Sorry, I kinda convinced myself to go back to LIG 😅  - happy to discuss in a call if needed.

src/openfe/protocols/openmm_rfe/hybridtop_units.py
IAlibay: I _think_ the API break is ok because we shouldn't have execution that spans multiple versions of openfe, but maybe I'm forgetting something!

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: I checked it and the small molecules are on a separate chain from the protein, so I'll switch to your suggestion and assign the residue number at the same time as the residue name!

hannahbaumann:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/hybridtop_units.py
hannahbaumann: Yes, this was here to avoid the API breaking change, but yes since iresuking an old simulation would anyways not be possible, I just removed this.

news/fix_unk_resnames.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Small molecules in RelativeHybridTopologyProtocol topologies (including the output PDB) are now named LG1 amd LG2 (alchemical ligand) and COF (cofactors) instead of UNK. If a residue name was already assigned, the assigned one is kept.
```
IAlibay: I think this needs updating to mention LIG or LG<some number>?

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -753,6 +758,46 @@ def run(
...
+            _set_offmol_resname(offmol, name)
+            _set_offmol_metadata(offmol, "residue_number", resnum)
+            resnum += 1
+
+        names: set[str] = set()
+        for comp in alchemical:
+            name = _get_offmol_resname(small_mols[comp])
+            assert name is not None
+            names.add(name)
+        alchem_resnames = sorted(names)
```
IAlibay, line 799: [nit] not something to fix now, but just an FYI that long term we might need to know what name belongs to what end state.

src/openfe/tests/protocols/openmm_rfe/test_hybrid_top_protocol.py
```diff
@@ -1091,7 +1096,7 @@ def test_dry_run_complex(
 ):
     # this will be very time consuming
     solv_settings.simulation_settings.sampler_method = method
-    solv_settings.output_settings.output_indices = "protein or resname  UNK"
+    solv_settings.output_settings.output_indices = "protein or resname  LG1"
```
IAlibay: Should this be LIG?


```suggestion
    solv_settings.output_settings.output_indices = "protein or resname  LIG"
```

src/openfe/protocols/openmm_rfe/hybridtop_units.py
```diff
@@ -1669,6 +1724,7 @@ def _execute(
             pdb_file=pdb_file,
             trajectory=trajectory,
             checkpoint=checkpoint,
+            ligand_resnames=setup_results.outputs.get("alchemical_resnames"),
```
IAlibay: This will silently get `None` if the entry isn't there. it will eventually just make structural analysis fail completely silentlly (because it's alll wrapped in a try/except).

IAlibay (changes requested):
Looking great! 
Couple of small things - mostly the test and the silent faillure with the `get`.

news/fix_unk_resnames.rst
hannahbaumann:
```suggestion
* Small molecules in RelativeHybridTopologyProtocol topologies (including the output PDB) are now named LIG (alchemical ligand) and COF (cofactors) instead of UNK. If a residue name was already assigned, the assigned one is kept.
```

hannahbaumann: Changed this!

IAlibay (approved):
lgtm thanks!

jthorton (approved):
LGTM!

PR #2121, opened 2026-08-04
Check for bond breaks in Hybrid Topology validation

jthorton:
This PR expands the atom mapping validation in the RelativeHybridTopology protocol to check for mappings which would break a chemical bond in either end state which is not currently supported by the protocol.

Checklist


jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
```diff
@@ -299,6 +299,40 @@ def _validate_mapping(
             if not m.componentA_to_componentB:
                 raise ValueError("No atoms are mapped between the two alchemical components.")
 
+            # check for broken bonds in the mapping
+            RelativeHybridTopologyProtocol._check_for_bond_breaks(mapping=m)
```
jthorton, line 304: Added as a function so that we can easily remove in future if we support this feature

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
```diff
@@ -299,6 +299,46 @@ def _validate_mapping(
...
+
+    @staticmethod
+    def _check_for_bond_breaks(
```
IAlibay, line 307: Am I correct in vaguely remembering that we have had some cases where bond breaks did lead to reasonable results (hif2a or eg5 maybe)?

If so, I'm wondering if we should add a `strict` keyword, to let users push through some validation in cases where they absolutely want to try to make it happen.

IAlibay (changes requested):
Just the one question.

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
jthorton: As the protocol does not formally handle bond breaks, it's safer not to have an option I would say. Then if users want to impliment bond breaks, they are free to subclass and overwrite the methods?

IAlibay: I would say, let's see if there's any regression in creating networks for the industry benchmark set - if there isn't then I agree it should be heavy handed, if there is, we should work out what that means for scope of application.

jthorton: Confirmed that all edges in our industry benchmarking set pass this validation.

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
```diff
@@ -299,6 +299,46 @@ def _validate_mapping(
...
+            frozenset((bond.GetBeginAtomIdx(), bond.GetEndAtomIdx()))
+            for bond in mapping.componentB.to_rdkit().GetBonds()
+        }
+        atom_map = mapping.componentA_to_componentB
+
+        # for each bond in moleculeA if the atoms are mapped then make sure the mapped atoms are also bonded
+        for atom_a, atom_b in mol_a_bonds:
+            if atom_a in atom_map and atom_b in atom_map:
+                mapped_bond = frozenset((atom_map[atom_a], atom_map[atom_b]))
+                if mapped_bond not in mol_b_bonds:
```
IAlibay: What about the reverse, i.e. what if you had a bond in mol_b that wasn't in mol_a?

IAlibay (changes requested):
I _think_ the test is one-directional, can you check it can be reversed, i.e. an appearing bond in stateB?

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
jthorton: Good catch I have added the reverse check as well and an example case which would get through without the reverse check!

jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_rfe/hybridtop_protocols.py
IAlibay, line 277: Could you update the raises section (I know it's a separate method but since it's being checked here it's useful to protocol authors).

IAlibay (approved):
Just the one thing, otherwise looks good to me!

PR #2151, opened 2026-08-11
Fix ligand order after multiprocessing charging

### LINKED ISSUE #1412: bulk_assign_partial_charges with processors>1 can change order of ligands
If you use multiple processors in bulk_assign_partial_charges, the order of the ligands in the input molecules list can change. Presumably the smaller ligands are faster to process and come out first.

jthorton:
Fixes #1412 by making sure the ligands are returned in the input order when multiprocessing is used. 

Checklist


jthorton:
pre-commit.ci autofix

src/openfecli/tests/commands/test_charge_generation.py
```diff
@@ -148,15 +157,21 @@ def test_charge_settings(methane, tmp_path, caplog, yaml_nagl_settings, ncores):
         caplog.set_level(logging.INFO)
         # make sure the charges are picked up
         result = runner.invoke(
-            charge_molecules, ["-M", mol_path, "-o", output_file, "-s", settings_path, "-n", ncores]
+            charge_molecules, ["-M", benzene_modifications_filepath, "-o", output_file, "-s", settings_path, "-n", ncores]
```
jthorton: Is this too expensive to charge this many molecules in the CI?

jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -505,6 +505,8 @@ def bulk_assign_partial_charges(
     if processors > 1:
         from concurrent.futures import ProcessPoolExecutor, as_completed
 
+        # track the input ordering as multiprocessing can shuffle the order of the ligands
+        input_order = [mol.smiles for mol in molecules]
```
IAlibay: I'm thinking it's possible for someone to have a list of the same molecule but different conformations as an input.

Maybe hash on inchi & positions?

IAlibay: We probably should also guard against the case where someone has the same SMC twice - I think it's ok to just crash out if that happens.

IAlibay (changes requested):
Probably need a bit more guards on how to handle the same molecule being passed multiple times in the list.

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Good point, I went with a simpler task order tracking approach which should avoid issues with positions and precision, with this we shouldn't need to be too restrictive on what the user supplies, let me know what you think.

jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -500,27 +500,36 @@ def bulk_assign_partial_charges(
...
-    charged_ligands = []
 
     if processors > 1:
         from concurrent.futures import ProcessPoolExecutor, as_completed
 
+        charged_ligands: list[SmallMoleculeComponent | None] = [None] * len(molecules)
         with ProcessPoolExecutor(max_workers=processors) as pool:
-            work_list = [
+            # track the input ordering as multiprocessing can shuffle the order of the ligands
+            future_to_index = {
```
IAlibay, line 510: This looks good, but could we not just use pool.map or would that be a problem for tqdm call (iirc tqdm progresses in the next ordered ligand completing, so if the first ligand is very slow and everything else is very fast, the progress bar is kinda useless).

IAlibay (approved):
Approving since it looks good, the only question is if Pool.map would be cleaner - but I understand that might be a problem with the progress bar.

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Yes map would work well but the progress bar might become useless, so maybe we can stick with this for now!

Issue #2117, opened 2026-07-31
Change default partial charge protocol settings to be "from force field"


IAlibay:

atravitz:
decided to keep this as an option but not default.

PR #2152, opened 2026-08-12
Let the force field resolve the partial charge assignment

### LINKED ISSUE #2116: Add a "forcefield" optional to `assign_partial_charges`
Add this as an extra method & also a kwarg that is list[str] | str | None and let's users define a series of offxml (either paths or serialized offxml).

jthorton:
Fixes #2116, #2117

Checklist


src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -339,6 +350,20 @@ def assign_offmol_partial_charges(
         if not overwrite:
             return offmol
 
+    if force_field is not None:
+        ff = ForceField(force_field)
+        if "ToolkitAM1BCC" in ff.registered_parameter_handlers:
+            ff.deregister_parameter_handler("ToolkitAM1BCC")
```
jthorton: If the user has 2.3.0 this will apply NAGL charges as the force field defines but might not be what the user wants if they expect am1bcc

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -20,6 +20,7 @@
     RDKitToolkitWrapper,
 )
 from openff.units import unit
+from sympy.codegen.ast import continue_
```
IAlibay: Is this import necessary?

src/openfe/protocols/openmm_utils/charge_generation.py
IAlibay: I liked your idea yesterday of having a "forcefield" entry to make it clear that this is the method that was being applied.

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -290,6 +291,7 @@ def assign_offmol_partial_charges(
     toolkit_backend: Literal["ambertools", "openeye", "rdkit"],
     generate_n_conformers: int | None,
     nagl_model: str | None,
+    force_field: list[str] | None = None,
```
IAlibay:
```suggestion
    force_fields: list[str] | None = None,
```
Should it be plural?

src/openfe/protocols/openmm_utils/charge_generation.py
IAlibay: Not sure I understand what you mean here sorry.

IAlibay (changes requested):
quick first look

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: good catch not sure why pycharm pulled that in!

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Added!

jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -441,11 +471,12 @@ def assign_offmol_partial_charges(
 def bulk_assign_partial_charges(
     molecules: list[SmallMoleculeComponent],
     overwrite: bool,
-    method: Literal["am1bcc", "am1bccelf10", "nagl", "espaloma"],
+    method: Literal["am1bcc", "am1bccelf10", "nagl", "espaloma", "forcefield"],
```
IAlibay: We'll need to update the OpenFFPartialChargeSettings to match this.

src/openfecli/parameters/plan_network_options.py
```diff
@@ -223,6 +223,7 @@ def load_yaml_planner_options(path: Optional[str], context) -> PlanNetworkOption
             off_toolkit_backend: ambertools
             number_of_conformers: None
             nagl_model: None
+            forcefields: None
```
IAlibay, line 226: Are there other changes necessary to enable this on the CLI? If so and you want them in a different PR, I'd move these changes to a separate CLI-focused PR.

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -286,10 +287,11 @@ def _generate_offmol_conformers(
 def assign_offmol_partial_charges(
     offmol: OFFMol,
     overwrite: bool,
-    method: Literal["am1bcc", "am1bccelf10", "nagl", "espaloma"],
+    method: Literal["am1bcc", "am1bccelf10", "nagl", "espaloma", "forcefield"],
```
IAlibay, line 297: We'll need to update the OpenFFPartialChargeSettings to match this.

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -339,6 +350,25 @@ def assign_offmol_partial_charges(
...
+                "The forcefield method requires a force field or list of force fields' to be provided "
+                "via `forcefields`."
+            )
+            raise ValueError(errmsg)
+
+        if isinstance(forcefields, str):
+            forcefields = [forcefields]
+
+        # this expects the full file name of the force field offxml file, e.g. "openff-2.0.0.offxml"
+        # which is different to how settings work which can leave off the .offxml extension
```
IAlibay: I think now we can use `.offxml` directly right? Should we change the defaults?

Also what happens if this encounters a non `.offxml` str? Should we try to add it?

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -319,6 +321,15 @@ def assign_offmol_partial_charges(
...
       If ``None``, the latest am1bcc NAGL charge model is used.
+    forcefields : list[str] | None, default None
+        An optional list of SMIRNOFF style force field offxml paths or strings which should be used to assign partial charges.
+
+    Notes
+    -----
+    Charges are applied based on the following source preferences:
+     - Charges already present on the ligand are retained if overwrite is ``False``
+     - Charges are applied using the input method and settings
+     - the forcefield option will apply the default charges as intended by the force field.
```
IAlibay: I think this is no longer true, is that correct?

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -339,6 +350,25 @@ def assign_offmol_partial_charges(
...
+            raise ValueError(errmsg)
+
+        if isinstance(forcefields, str):
+            forcefields = [forcefields]
+
+        # this expects the full file name of the force field offxml file, e.g. "openff-2.0.0.offxml"
+        # which is different to how settings work which can leave off the .offxml extension
+        ff = ForceField(*forcefields)
+        # let the force field resolve the partial charge assignment
+        charges = ff.get_partial_charges(offmol)
```
IAlibay: I believe this needs to get wrapped around `toolkit_registry_manager`, otherwise we'll go back to encountering the annoying rdkit & openeye toolkit aren't compatible problem.

src/openfe/tests/protocols/test_openmmutils.py
```diff
@@ -1291,6 +1292,64 @@ def test_openeye_import_error(self, monkeypatch, uncharged_mol):
...
+            toolkit_backend="rdkit",
+            generate_n_conformers=None,
+            # set the model to none this should use the model define in the force field.
+            nagl_model=None,
+            # use a force field that has a NAGL handler and another redundant force field file
+            forcefields=["openff-2.3.0.offxml", "opc-1.0.0.offxml"],
+        )
+
+        assert uncharged_mol.partial_charges is not None
+        assert np.any(uncharged_mol.partial_charges)
```
IAlibay: Do you want to check against charging with ashgc directly?

IAlibay (changes requested):
Couple of things - mainly it doesn't look like the CLI would work right now (maybe it just needs a test to check it).

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Updated!

jthorton: Updated!

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Updated to represent the correct ordering.

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Probably but which registry would we use as it could influence the charge method used? I think the default is openeye and am1bccelf10 and then fall back to AmberToolsam1bcc?

src/openfe/tests/protocols/test_openmmutils.py
jthorton: Yeah extended the test to cover this.

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Updated to accept both and added a test, and yes maybe the default should now have the extension as well.

src/openfecli/parameters/plan_network_options.py
jthorton: Finished up the CLI support and added tests, I did intend to do it all in this PR.

jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Just use what the user has in the settings for the toolkit_backend and make it clear in the docs that the backend is always followed when a charge is generated.

jthorton: Ah this doesn't work if the force field uses nagl charges, as our default is ambertools, we might need to use a different method to make the registry. Maybe something like:
If the user has openeye pass:
- openeye
- nagl 
- ambertools
If the user has rdkit and no openeye:
- rdkit
- nagl
- ambertools

IAlibay: Sorry I don't understand why it's not working.

The "AmberTools" backend is AmberTools + RDKit, that should be enough for NAGL to work no?

IAlibay: Ah ok - the issue is that the NAGL registry isn't in there?

IAlibay: I _think_ it might be ok to just add the NAGLToolkitWrapper to both AmberTools & OpenEye backend lists - please double check but I think it will still do the "protection" that we're trying to do (i.e. it will block you from doing am1bcc with openeye if you don't want it).

src/openfecli/parameters/plan_network_options.py
IAlibay, line 226: How about including an example of this under the settings help section?

src/openfecli/tests/commands/test_charge_generation.py
```diff
@@ -188,3 +189,71 @@ def test_charge_settings(
...
+    # make sure we can use forcefield charges if we specify a forcefield
+    runner = CliRunner()
+    mol_path = tmp_path / "methane.sdf"
+    methane.to_file(str(mol_path), "sdf")
+    out_path = str(tmp_path / "charged_methane.sdf")
+
+    settings = {
+        "partial_charge": {
+            "method": "forcefield",
+            "settings": {"forcefields": ["openff_unconstrained-2.3.0"]},
```
IAlibay: [nit] Any reason for using unconstrained here? Might be better to use the default we use day-to-day.

src/openfe/protocols/openmm_utils/charge_generation.py
IAlibay: Can you open a PR on gufe to update the default please?

IAlibay (commented):
Overall looks good, but yeah we might need to change the backends behaviour. Let me know if you want to have a chat tomorow morning.

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -339,6 +349,42 @@ def assign_offmol_partial_charges(
         if not overwrite:
             return offmol
 
+    if method.lower() == "forcefield":
```
IAlibay: Can we have a check for the other way around too? I'm thinking new users might not easily know you need to set both - especially via the CLI.

jthorton: Add the reverse check and test.

src/openfecli/tests/commands/test_charge_generation.py
jthorton: No was just trying different one, will change to the normall one.

src/openfecli/parameters/plan_network_options.py
jthorton: That section is already getting quite big, I think it would be better to point to the docs and have small examples there that users can copy for some different options, this would simplify the CLI help message as well!

jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Add the wrapper to all backends if nagl is available which I think is what we want?

src/openfecli/parameters/plan_network_options.py
jthorton: Added the new options to the CLI settings template [here](https://github.com/OpenFreeEnergy/ExampleNotebooks/pull/325)

src/openfe/protocols/openmm_utils/omm_settings.py
```diff
@@ -271,11 +273,17 @@ class OpenFFPartialChargeSettings(BasePartialChargeSettings):
       Only ``ambertools`` and ``rdkit`` `off_toolkit_backend`` options
       are supported. A maximum of one conformer is allowed.
 
+    ``forcefield``:
```
IAlibay, line 276: I've been thinking about the impact of doing this - i.e. what would this mean to users if we merged this PR with a `forcefield` entry here and how confusing it would be at the Protocol level.

Is there any way we could *not* update these settings and just have it as a separate arg that gets picked up from the YAML file (i.e. I think right now we only need this for the CLI?).

IAlibay (changes requested):

src/openfe/protocols/openmm_utils/omm_settings.py
IAlibay: I don't love it, but what if we a) created a subclass of settings specifically for the CLI, or b) just allowed extras?

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -333,12 +350,54 @@ def assign_offmol_partial_charges(
...
+
+    # validate the input combination first
+    if method_name == "forcefield":
+        if forcefields is None:
+            errmsg = (
+                "The forcefield method requires a force field or list of force fields' to be provided "
+                "via `forcefields`."
+            )
+            raise ValueError(errmsg)
+    elif forcefields is not None:
```
IAlibay: This can pass even if forcefields is an empty list,

src/openfe/protocols/openmm_utils/omm_settings.py
jthorton: I think this would depend on what we do with #2132, if we say that no charge assignment will be done by OpenFE and set the partial charge settings to None by default and have a warning that the option is deprecated and ignored at the protocol level then using this as the CLI charge settings class would be fine. 

If not would we not want a to get the pydantic validation on the settings?

src/openfe/protocols/openmm_utils/charge_generation.py
jthorton: Thanks this should be caught now!

jthorton:
pre-commit.ci autofix

src/openfe/protocols/openmm_utils/omm_settings.py
IAlibay: I think this PR has to try to exist standalone of any follow up. There's a good chance we won't be able to deal with the other PR properly ahead of the next release. It would be better if we could safely merge this PR first. 

I'm going to advocate for a specific CLI settings subclass for now.

jthorton:
pre-commit.ci autofix

IAlibay:
From today's iteration meeting - we should add a test for a system with virtual sites.

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -457,7 +517,7 @@ def bulk_assign_partial_charges(
     overwrite : bool
       Whether or not to overwrite any existing non-zero partial charges.
       Note that zeroed charges will always be overwritten.
-    method : Literal['am1bcc', 'am1bccelf10', 'nagl', 'espaloma']
+    method : Literal['am1bcc', 'am1bccelf10', 'nagl', 'espaloma', 'forcefield]
```
IAlibay:
```suggestion
    method : Literal['am1bcc', 'am1bccelf10', 'nagl', 'espaloma', 'forcefield']
```

src/openfe/protocols/openmm_utils/charge_generation.py
```diff
@@ -457,7 +517,7 @@ def bulk_assign_partial_charges(
...
       Partial charge assignment method.
       Supported methods include; am1bcc, am1bccelf10, nagl, and espaloma.
```
IAlibay:
```suggestion
      Supported methods include; am1bcc, am1bccelf10, nagl, espaloma and forcefield.
```

IAlibay (approved):
Just the one thing, otherwise it looks good to me - approving early.

jthorton:
pre-commit.ci autofix

IAlibay (approved):
lgtm, thanks!

PR #2155, opened 2026-08-13
filesystem warehouse usability improvements

atravitz:
I think that creating a new filesystem warehouse and loading an existing one should have different guard rails. 

Needs gufe changes in https://github.com/OpenFreeEnergy/gufe/pull/827

otherwise, you:
 - can think you’re loading an existing warehouse with data in it, but instead a new, empty warehouse is created in a slightly different location
 - think you're loading a new warehouse, but you've just added tasks to an existing warehouse when you didn't mean to
 
 

LLMs or other AI-powered tools (beyond simple IDE use cases) were used in this contribution:  no

Checklist
* [x] Added a ``news`` entry, or the changes are not user-facing. - news items will happen when the Epic PR gets merged.

### DIFF
--- a/src/openfe/orchestration/exorcist_utils.py
+++ b/src/openfe/orchestration/exorcist_utils.py
@@ -4,11 +4,11 @@ This module translates an :class:`gufe.AlchemicalNetwork` into Exorcist task
 structures and can initialize an Exorcist task database from that graph.
 """
 
-import sys
 from pathlib import Path
 
 import exorcist
 import networkx as nx
+import pandas as pd
 from gufe import AlchemicalNetwork, ProtocolDAG
 
 from openfe.orchestration import FileSystemWarehouse
@@ -102,3 +102,22 @@ def build_task_db_from_alchemical_network(
     db = exorcist.TaskStatusDB.from_filename(db_path)
     db.add_task_network(global_task_dag, max_tries)
     return db, warehouse
+
+
+def get_task_df(task_db: exorcist.TaskStatusDB) -> pd.DataFrame:
+    """Create a pandas Dataframe from task_db.
+
+    Parameters
+    ----------
+    task_db : exorcist.TaskStatusDB
+        A task database.
+
+    Returns
+    -------
+    pd.DataFrame
+        A dataframe of the tasks and their statuses
+    """
+    status_name_encoding = {e.value: e.name for e in exorcist.TaskStatus}
+    task_table = pd.read_sql_table("tasks", task_db.engine)
+    task_table.replace({"status": status_name_encoding}, inplace=True)
+    return task_table
--- a/src/openfe/orchestration/exorcist_utils.py
+++ b/src/openfe/orchestration/exorcist_utils.py
@@ -4,18 +4,19 @@ This module translates an :class:`gufe.AlchemicalNetwork` into Exorcist task
 structures and can initialize an Exorcist task database from that graph.
 """
 
+import sys
 from pathlib import Path
 
 import exorcist
 import networkx as nx
-import pandas as pd
-from gufe import AlchemicalNetwork
+from gufe import AlchemicalNetwork, ProtocolDAG
 
 from openfe.storage.warehouse import WarehouseBaseClass
 
 
-def alchemical_network_to_task_graph(
-    alchemical_network: AlchemicalNetwork, warehouse: WarehouseBaseClass
+def _alchemical_network_to_task_graph(
+    alchemical_network: AlchemicalNetwork,
+    warehouse: WarehouseBaseClass,
 ) -> nx.DiGraph:
     """Build a global task DAG from an alchemical network.
 
@@ -30,9 +31,8 @@ def alchemical_network_to_task_graph(
     Returns
     -------
     nx.DiGraph
-        A directed acyclic graph where each node is a task ID in the form
-        ``"<transformation_key>:<protocol_unit_key>"`` and edges encode
-        protocol-unit dependencies.
+        A directed acyclic graph where each node is a task ID with the ProtocolUnit key as a name
+        and edges encode protocol-unit dependencies.
 
     Raises
     ------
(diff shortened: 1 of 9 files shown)


src/openfe/storage/warehouse.py
```diff
@@ -408,14 +410,16 @@ class FileSystemWarehouse(WarehouseBaseClass):
...
     """
 
-    def __init__(self, name):
-        # TODO: should name and location be different?
-        self.root_dir = pathlib.Path(f"{name}")
+    def __init__(self, root_dir: pathlib.Path, exist_okay=False):
+        self.root_dir = pathlib.Path(root_dir)
+        if self.root_dir.is_dir() and not exist_okay:
+            raise ValueError(
+                "`root_dir` already exists. To load an existing Warehouse, use FileSystemWarehouse.load(`root_dir`)"
```
ethanholz: I like the error, but I am a little concerned about how it gets passed down to a user. If a user reruns a command using this under the hood will it make sense? Do we even care about that here?

src/openfe/storage/warehouse.py
```diff
@@ -1,5 +1,7 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/gufe
+from __future__ import annotations
```
ethanholz, line 3: I am sure you are aware but seems there are changes to this in Python 3.14 https://docs.python.org/3.14/whatsnew/3.14.html#from-future-import-annotations

ethanholz (approved):
I think this is great, a few minor notes but non-blocking

atravitz:
@ethanholz note that I have the same complaint about exorcist `TaskStatusDB.from_filename()` - it silently creates a new, empty, db if the path doesn't exist.

src/openfe/storage/warehouse.py
```diff
@@ -424,4 +428,14 @@ def __init__(self, name):
             tasks=tasks_store,
             protocol_dags=protocol_dag_store,
         )
-        super().__init__(stores, name)
+        name = self.root_dir.resolve().name
+        super().__init__(stores=stores, name=name)
+
+    @classmethod
+    def load(cls, root_dir: pathlib.Path) -> FileSystemWarehouse:
```
atravitz: "load" might be the wrong word, since it's not actually being loaded into memory. maybe "from_filepath" to parallel task db?

atravitz:
If upon initial creation we create all the store top directories, then when pointing to a an existing warehouse, we could do a quick validation to make sure all the dirs exist.
Otherwise, if you point it to a generic non-warehouse directory,  trying to access any of the stores would just create them in real time. 

I think the broader question here is "what makes a directory a FileSystemWarehouse?" and how do we validate that.

src/openfe/storage/warehouse.py
ethanholz: Maybe even `from_dir` might be good.

ethanholz:
> If upon initial creation we create all the store top directories, then when pointing to a an existing warehouse, we could do a quick validation to make sure all the dirs exist. Otherwise, if you point it to a generic non-warehouse directory, trying to access any of the stores would just create them in real time.
> 
> I think the broader question here is "what makes a directory a FileSystemWarehouse?" and how do we validate that.

I think this is the right path forward!

src/openfe/storage/warehouse.py
atravitz: i like it, done!

src/openfe/storage/warehouse.py
atravitz: thanks for the link! I think this is fine in this instance, since it's pulling in Python 3.14 functionality.

src/openfe/orchestration/exorcist_utils.py
```diff
@@ -30,69 +32,109 @@ def alchemical_network_to_task_graph(
...
+    return global_task_dag
 
 
-# TODO: do we test adding a multiple alchemical networks to the same task graph?
 def build_task_db_from_alchemical_network(
     alchemical_network: AlchemicalNetwork,
-    warehouse: WarehouseBaseClass,
-    db_path: Path | None = None,
+    warehouse_dir: Path,  # TODO: make optional?
+    db_path: Path,  # TODO: make optional?
```
ethanholz, line 74: I think since these are keyed to the `campaign_name`, I think its fine to make these `Optional`.

src/openfe/storage/warehouse.py
```diff
@@ -408,20 +408,49 @@ class FileSystemWarehouse(WarehouseBaseClass):
...
+        if self.root_dir.is_dir() and not exist_ok:
+            raise ValueError(
+                "`root_dir` already exists. To load an existing Warehouse, use FileSystemWarehouse.from_dir(`root_dir`)"
+            )
+        self.root_dir.mkdir(exist_ok=exist_ok)  # make parents?
+        setup_store = FileStorage(f"{self.root_dir}/setup", exist_ok=exist_ok)
+        result_store = FileStorage(f"{self.root_dir}/result", exist_ok=exist_ok)
+        shared_store = FileStorage(f"{self.root_dir}/shared", exist_ok=exist_ok)
+        tasks_store = FileStorage(f"{self.root_dir}/tasks", exist_ok=exist_ok)
+        protocol_dag_store = FileStorage(f"{self.root_dir}/protocol_dags", exist_ok=exist_ok)
```
ethanholz, line 422: Just food for thought, does not block... If a warehouse is partially corrupted, does this work? For example, if the default is `exist_ok` is `False`, what happens when someone accidentally deletes a single store, is the convention to just recreate the whole store? Is this worth caring about?

src/openfe/tests/orchestration/test_exorcist_utils.py
```diff
@@ -175,57 +180,93 @@ def test_build_task_db_checkout_order_is_dependency_safe(tmp_path, request, fixt
...
+@pytest.fixture()
+def benzene_star_map_task_db(benzene_variants_star_map, tmp_path):
+    warehouse = _RecordingWarehouse()
+    network = benzene_variants_star_map
+    global_task_dag = _alchemical_network_to_task_graph(
+        network, cast(WarehouseBaseClass, warehouse)
+    )
+    db = exorcist.TaskStatusDB.from_filename(
+        tmp_path / "tasks.db"
+    )  # TODO: how to make a barebones db that doesn't write to disk?
```
ethanholz:
```suggestion
    db = exorcist.TaskStatusDB.from_filename(
        ":memory:"
    )  # TODO: how to make a barebones db that doesn't write to disk?
```

This is possible to do, SQLite has an in-memory database option. We use different convention for this in `exorcist` testing here: https://github.com/OpenFreeEnergy/exorcist/blob/c591241c0be2be44840e92332a67196122b3df06/exorcist/tests/test_taskstatusdb.py#L98 but it comes from this https://docs.sqlalchemy.org/en/21/dialects/sqlite.html#connect-strings.

ethanholz (approved):
Overall this is great. I think that what you have done here is fantastic and looks to be maintainable. A few minor comments that could help with architecture/testing.

src/openfe/storage/warehouse.py
atravitz: I _think_ it makes sense - if you rerun this command, it should fail because you've tried to initialize a new warehouse in the same place twice.

src/openfe/orchestration/exorcist_utils.py
atravitz: agreed - will do as a follow-up since it's non-breaking.

src/openfe/storage/warehouse.py
atravitz: This is on my list of things to be addressed after the MVP is in to make things more robust.

src/openfe/tests/orchestration/test_exorcist_utils.py
atravitz: that's exactly what I was looking for, thank you!

PR #2174, opened 2026-08-25
CLI: support worker-based execution

atravitz:

LLMs or other AI-powered tools (beyond simple IDE use cases) were used in this contribution: no

Checklist


src/openfecli/tests/commands/test_plan_rhfe_network.py
```diff
@@ -255,6 +255,10 @@ def test_plan_rhfe_network_charge_overwrite(dummy_charge_dir_args, tmp_path, yam
         if overwrite:
             assert "Overwriting partial charges" in result.output
 
+        # make sure --networks-only works as expected
```
atravitz, line 258: for both this and plan-rbfe-network, I used the ``--networks-only`` option for tests where we don't check the transformation output anyway. I figure this speeds up our tests instead of piling on more cases.

src/openfecli/commands/setup_task_campaign.py
```diff
@@ -0,0 +1,44 @@
...
+
+import click
+from gufe import AlchemicalNetwork
+
+from openfecli import OFECommandPlugin
+from openfecli.parameters import ALCHEMICAL_NETWORK
+from openfecli.utils import print_duration, write
+
+
+# TODO: add n_repeats
```
atravitz, line 11: i'll do this in a follow-up PR since it touches the API as well. 
https://github.com/OpenFreeEnergy/openfe/issues/2190

docs/guide/execution/run_tasks.sh
IAlibay, line 1: Is there an advantage to having this as a separate file rather than in-line? The latter feels like it would be easier to update in the future.

docs/guide/execution/task_based_execution.rst
```diff
@@ -0,0 +1,132 @@
+.. _userguide_task_based_execution:
+
+Task-based Execution
+====================
+
+In contrast to :ref:`quickrun execution <userguide_quickrun>`, task-based execution does not require that you explicitly define the Transformation to be executed.
```
IAlibay, line 6:
```suggestion
In contrast to :ref:`quickrun execution <userguide_quickrun>`, task-based execution does not require that you explicitly define the ``Transformation`` to be executed.
```
Also outside of this PR, a future todo would be to try to convert all of these to :class:`Transformation`, etc.. or similar, so that we can reference in text.

docs/guide/execution/run_tasks.sh
atravitz: I have no preference, happy to move it to in-line!

docs/guide/execution/task_based_execution.rst
```diff
@@ -0,0 +1,132 @@
...
+In task-based execution, an OpenFE ``Worker`` is given a ``Warehouse`` and a ``TaskStatusDB`` which handle storage and orchestration, respectively.
+The ``Worker`` finds the next available ``task`` in the ``TaskStatusDB`` and retrieves the necessary data from the ``Warehouse`` to execute the task.
```
IAlibay: This is a lot for a user to read as the second sentence in a user guide entry. Is there any way this could be simplified or maybe the details moved to some other section?

docs/guide/execution/task_based_execution.rst
```diff
@@ -0,0 +1,132 @@
...
+Task-based Execution with the CLI
+---------------------------------
+
+Setting up a Campaign
+~~~~~~~~~~~~~~~~~~~~~
+
+An ``AlchemicalNetwork`` will be our input for executing a campaign.
+See the following examples for how to create an ``AlchemicalNetwork``:
+
+.. TODO
```
IAlibay: Unclear if this todo is still active and you're missing some links / examples here.

docs/guide/execution/task_based_execution.rst
```diff
@@ -0,0 +1,132 @@
...
+
+    > openfe plan-rbfe-network --networks-only
+
+
+Once you have an AlchemicalNetwork, use the following command to set up the task-based campaign.
+
+
+.. code:: bash
+
+    > openfe setup-task-campaign --alchemical-network tyk2_alchemical_network.json
```
IAlibay: The introduction of `tyk2` here is done abruptly without any mention. I would either mention that you're using tyk2 as a placeholder or just use a blank generic (i.e. either "my_alchemical_network.json" or "alchemical_network.json").

Same thing with follow-on uses of tyk2.

docs/guide/execution/task_based_execution.rst
```diff
@@ -0,0 +1,132 @@
...
+.. TODO: explain task dag? maybe save that for developer docs?
+
+To execute a single ``task`` (where here a ``task`` is one ``ProtocolUnit``), you can simply run:
+
+.. code:: bash
+
+    > openfe run-task --warehouse warehouse_tyk2/ --task-db tasks_tyk2.db
+
+
+You'll now see that one task has been completed, and a new task has been unblocked:
```
IAlibay, line 86: "once that command finishes"?

docs/guide/execution/run_tasks.sh
IAlibay: My vote would be to move it inline - purely to make it easier to edit.

docs/guide/execution/task_based_execution.rst
```diff
@@ -0,0 +1,132 @@
...
+In practice, you will likely be submitting many workers simultaneously using SLURM or similar.
+You can call this command in a loop, so that after a ``task`` is completed, the ``Worker`` automatically picks up a new ``task``, continuing to run tasks in serial until the walltime runs out.
+
+
+.. literalinclude:: run_tasks.sh
+    :caption: Example SLURM submission script for task-based execution
+    :linenos:
+    :language: bash
+
+To run multiple workers in parallel, submit ``run_tasks.sh`` multiple times as separate jobs.
```
IAlibay, line 126: It would be good if there was an easy way to get the number of tasks of a given status - that way folks would know the maximum number of workers they need to submit.

src/openfe/tests/conftest.py
```diff
@@ -885,3 +894,116 @@ def appearing_bond_mapping(chlorobenzene):
         componentB=chlorobenzene,
         componentA_to_componentB={5: 0, 7: 1, 8: 2, 9: 3, 10: 4, 11: 5, 12: 6},
     )
+
+
+@pytest.fixture
+def solv_comp():
```
IAlibay, line 900: These names are, in my opinion, too generic. Could you please make these more specific - e.g. "kcl_solvent_component" (solv_comp is a name used a lot in a tests).

src/openfe/tests/conftest.py
```diff
@@ -885,3 +894,116 @@ def appearing_bond_mapping(chlorobenzene):
...
+    yield SolventComponent(positive_ion="K", negative_ion="Cl", ion_concentration=0.0 * unit.molar)
+
+
+@pytest.fixture
+def solvated_complex(T4_protein_component, benzene_transforms, solv_comp):
```
IAlibay, line 905: We have a toluene_complex_system in the protocols conftest, would that be suitable here or does it need to be KCL?

src/openfe/tests/conftest.py
```diff
@@ -885,3 +894,116 @@ def appearing_bond_mapping(chlorobenzene):
...
+    return ChemicalSystem(
+        {
+            "ligand": benzene_transforms["toluene"],
+            "solvent": solv_comp,
+        }
+    )
+
+
+@pytest.fixture
+def absolute_transformation(solvated_ligand, solvated_complex):
```
IAlibay, line 926: absolute_dummy_transformation?

src/openfe/tests/conftest.py
```diff
@@ -885,3 +894,116 @@ def appearing_bond_mapping(chlorobenzene):
...
+    return gufe.Transformation(
+        solvated_ligand,
+        solvated_complex,
+        protocol=DummyProtocol(settings=DummyProtocol.default_settings()),
+        mapping=None,
+    )
+
+
+@pytest.fixture
+def complex_equilibrium(solvated_complex):
```
IAlibay, line 936: what does "equilibrium" mean here?

src/openfe/tests/conftest.py
```diff
@@ -885,3 +894,116 @@ def appearing_bond_mapping(chlorobenzene):
...
+    return gufe.NonTransformation(
+        solvated_complex,
+        protocol=DummyProtocol(settings=DummyProtocol.default_settings()),
+    )
+
+
+@pytest.fixture
+def benzene_variants_star_map(benzene_transforms, solv_comp, T4_protein_component):
```
IAlibay, line 944:
```suggestion
def benzene_variants_star_map_alchemicalnetwork(benzene_transforms, solv_comp, T4_protein_component):
```
Took me a while to realise that this was meant to be an alchemical network, not a ligand network.

src/openfecli/commands/setup_task_campaign.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@click.command(
+    "setup-task-campaign",
+    short_help="Build a Warehouse and corresponding TaskDB from an AlchemicalNetwork.",
+)
+# TODO: add --name?
+@ALCHEMICAL_NETWORK.parameter(multiple=False, required=True, help=ALCHEMICAL_NETWORK.kwargs["help"])
+@print_duration
+def setup_task_campaign(alchemical_network: str | Path):
```
IAlibay, line 33: Some docstring for this would be good.

src/openfecli/commands/to_legacy_json.py
```diff
@@ -0,0 +1,55 @@
+from pathlib import Path
+
+import click
+
+from openfecli import OFECommandPlugin
+from openfecli.parameters import OUTPUT_DIR  #  OVERWRITE, WAREHOUSE
```
IAlibay:
```suggestion
from openfecli.parameters import OUTPUT_DIR
```
?

IAlibay (changes requested):
Overall looks good, but I'm going to request changes mostly over the conftest fixture names.

src/openfecli/commands/to_legacy_json.py
atravitz: ah thank you!

docs/guide/execution/run_tasks.sh
atravitz: done!

docs/guide/execution/task_based_execution.rst
atravitz: agreed - this is something I intend to do when we do our docs audit soon.

docs/guide/execution/task_based_execution.rst
atravitz: I think that's less precise - the command can finish, but if the _task_ is not completed (e.g. errors), then the downstream task will not be unblocked.

docs/guide/execution/task_based_execution.rst
atravitz: My understanding is that User Guide is meant for this type of explanatory content. Maybe more importantly, this mirrors the structure of the [quickrun page](https://docs.openfree.energy/en/stable/guide/execution/quickrun_execution.html), and I think having parallel structure is preferable until we determine a better overall organization.

Bigger picture, I think it would be appropriate to move some of the content from both the quickrun and task-based execution guides _out_ of the User Guide and into cookbooks, but for the scope of this PR I vote to keep as-is.

Thoughts?

docs/guide/execution/task_based_execution.rst
atravitz: Ah thanks - I meant to link to the "creating an alchemical network" example notebook once it was done, and now it is! Done.

src/openfe/tests/conftest.py
atravitz: see comment on this PR

src/openfe/tests/conftest.py
atravitz: I agree, but see comment on this PR

src/openfe/tests/conftest.py
atravitz: see comment on this PR

docs/guide/execution/task_based_execution.rst
atravitz: I agree and consider this a high priority for v1.14 or a nice-to-have for v1.13. I'll open an issue to track - let me know if you consider this a must-have for v1.13.

atravitz: https://github.com/OpenFreeEnergy/openfe/issues/2196

src/openfe/tests/conftest.py
atravitz: I agree - see the PR comment

src/openfe/tests/conftest.py
atravitz: I agree we can clean this up, see the PR comment

docs/guide/execution/task_based_execution.rst
atravitz: Good point, I added an additional sentence here. I'm trying to use a more real-world example to hopefully make this clear that a typical case creates a warehouse and task db for a single campaign, similar to the scope of our other execution examples.

src/openfecli/commands/setup_task_campaign.py
atravitz: added! I might want to expand on this, but I think it'll do for now.

atravitz (commented):
You're right about the unclear fixture names. However, I think we should address these as a follow-up since the fixtures in that conftest.py are showing up in this diff only because of a file renaming, and are not actually being introduced in this PR.

I think the follow-up should be two simultaneous gufe and openfe PRs, since there's a bunch of duplicated code from fixtures being pull from gufe into openfe. We should rename them all at the same time, or even import directly from gufe if appropriate. Renaming here will just make it harder to find the duplicated code in the follow-up work.

tracking here: https://github.com/OpenFreeEnergy/openfe/issues/2195

docs/guide/execution/task_based_execution.rst
IAlibay: I'll open up a separate issue, there's a deeper usability problem here imho.

IAlibay: I think this, alongside a means of easily setting the status of one or more task(s), should be a must-have for v1.13.

IAlibay:
> You're right about the unclear fixture names. However, I think we should address these as a follow-up since the fixtures in that conftest.py are showing up in this diff only because of a file renaming, and are not actually being introduced in this PR.
> 
> I think the follow-up should be two simultaneous gufe and openfe PRs, since there's a bunch of duplicated code from fixtures being pull from gufe into openfe. We should rename them all at the same time, or even import directly from gufe if appropriate. Renaming here will just make it harder to find the duplicated code in the follow-up work.
> 
> tracking here: #2195

That's fine with me, although I'd ask that this be done before we merge the epic. I.e. the names (especially `solv_comp`) overlap enough with non-fixture names in protocol tests that this could become painful as we develop things.

docs/guide/execution/task_based_execution.rst
IAlibay: It's ok to have details - in my opinion, this is something that should be in the userguide. What I'm trying to convey is these two sentences are really hard to parse, even for someone like me that already knows what it means. So having this as the second sentence in a userguide is maybe too much for someone new to this tooling.

docs/guide/execution/task_based_execution.rst
IAlibay: The wording isn't exactly what I'm trying to convey.

What I'm trying to point to here is that, to me, there's a narrative jump - in my opinion it's missing some kind linker. I.e. We're jumping straight from "run this task", "now you see a new task has been unblocked".

IAlibay (approved):
There's a couple of things I think need improving in the userguide, but I'll approve early.

docs/guide/execution/task_based_execution.rst
atravitz: got it - I split this information to fit into the flow of the guide a bit better so it's not all dumped up front.

docs/guide/execution/task_based_execution.rst
atravitz: I think this is more clear now!

